PR #47, opened 2017-01-13
Add Apache 2.0 License

CloudNiner:
Also updates subproject commit to include subproject LICENSE.

Trivial, merging.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,13 @@
+Copyright 2017 Azavea, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


spencerrecneps:
Just curious what the thinking was behind Apache 2.0. I'm not opposed, but it's a conversation that I should probably have with PFB at some point to make sure they're on board. Would be helpful to have your thoughts on it first.

My only concern is the potential for commercial use. PFB may prefer to restrict it to non-commercial use. Otherwise, my understanding of the license is in line with the goals of the project.

CloudNiner:
Apache 2.0 is the default Azavea license of choice for open source projects whenever one isn't specified. If you'd like to switch to a different license, just let us know which one and we can update.

spencerrecneps:
Makes sense. Let's leave it as is for now. I'll bring it up sometime with PFB and we can change if needed.

PR #48, opened 2017-01-17
Add NB_BOUNDARY_BUFFER to README

CloudNiner:
Running the analysis without explicitly setting NB_BOUNDARY_BUFFER can lead to a confusing scenario where you end up with no linked census blocks, and all of the statistics in the `neighborhood_overall_scores` table are zero.


flibbertigibbet:
Tested and it ran successfully.

Issue #42, opened 2016-12-22
Autodetect UTM zone before analysis


jbranigan:
Figure out the SRID based on lat/lon

UTM zone 10N is 32610
UTM zone 10S is 32710

Reference for ideas:
https://github.com/jbranigan/geo-scripts-python/blob/master/latlng2utm/detect-utm-zone.py

spencerrecneps:
I gather this means we will continue to use meters as the unit for measuring distances? I think that's a good thing, but we will need to incorporate some of the changes from #39 that swapped calculations from feet to meters.

spencerrecneps:
The commit on my repo can be seen here: https://github.com/spencerrecneps/pfb/commit/e1c7d31100f73e68d29bcd72c8f31e7e40d19c8d

CloudNiner:
Yes, that seems correct. We'll have to make those ft -> m changes as part of this task, otherwise measurements would be wrong if brought in as part of azavea/pfb#10

PR #59, opened 2017-02-06
Feature/autodetect utm zone srid

flibbertigibbet:
Depends on azavea/pfb#17

Connects #42 

Install `python-gdal` in environment so UTM zone may be detected from input shapefile.


CloudNiner:
Subproject commit update is also in #65

flibbertigibbet:
Closing in favor of #17 update to run

PR #60, opened 2017-02-06
Add AnalysisJob model

CloudNiner:
## Overview

Adds the Django model we'll use to track analysis jobs run via the web application. Moves the Neighborhood model to `pfb_analysis` since its only used to track various boundaries used by the different analysis jobs.

This model should encapsulate all of the information the two scripts `import.sh` and `run_connectivity.sh` need to run.

An AnalysisJob requires:
- [x] Boundary to run against
- [x] The state the boundary is in
- [x] State fips code

### Notes

- Removed all pre-existing references to Neighborhood in the `pfb_network_connectivity` and `users` apps, since that use case doesn't apply here.
- Added `django-storages` and the configuration necessary to both Django and Terraform to ensure that the requisite buckets exist, and that we have the proper credentials set in all environments.
- Added the `us` package, which gives us the oh so helpful `USStateField` and helper methods which automatically print the fips code we'll need to pass to the analysis jobs (via `neighborhood.state.fips`)
- Added no tests, as there's no non-trivial application specific logic to test here
- If you delete an instance of Neighborhood from the database, the django-storages plugin does not cleanup your file on S3. I don't think that's a huge issue for us in this case.

## Testing

Ensure `./scripts/server` is running and that you've set up your PFB AWS credentials with `aws configure --profile pfb`.

Open up a Django console with `./scripts/django-manage shell_plus`. Copy the cambridge neighborhood boundary file (neighborhood_boundary_02138.zip) from FILESHARE to `./data/`.

Then, run the following commands:
```
boundary_file = File(open('/data/neighborhood_boundary_02138.zip'))
cambridge = Neighborhood.objects.create(name='cambridge-ma', label='Cambridge, MA', boundary_file=boundary_file, organization=org, state_abbrev='MA')
```
This should take a second, then go to S3 (via AWS console or CLI) and inspect the `<username>-pfb-storage-us-east-1` bucket. Your file should exist there.

Then, you can also do:
```
cambridge.boundary_file.url
```
to print the signed s3 download url, and copy/paste that to a browser to download the file you just uploaded.

Connects #55, #99


src/django/pfb_analysis/models.py
```diff
@@ -0,0 +1,79 @@
...
+
+class Neighborhood(models.Model):
+    """Neighborhood boundary used for an AnalysisJob """
+
+    def __repr__(self):
+        return "<Neighborhood: {}, ({})>".format(self.name, self.label)
+
+    name = models.SlugField(max_length=256, primary_key=True)
+    label = models.CharField(max_length=256, help_text='Human-readable label for boundary')
+    state_abbrev = USStateField(help_text='The US state that the uploaded neighborhood is in')
```
KlaasH: Is it possible to require this at the database level?  It gets set to empty string if omitted.  I tried adding 'null=False, blank=False` but `makemigrations` didn't detect a change.

src/django/users/models.py
```diff
@@ -23,41 +23,35 @@ class OrganizationTypes(object):
...
-    users within the system. Membership in this organization will be tightly controlled and will be
-    limited to PFB personnel who are responsible for activating/deactivating neighborhood sites and
-    creating initial admin users within a neighborhood organization.
-
-    **Neighborhood.** Membership in an organization that represents a neighborhood will only have
-    access to resources for that particular neighborhood.
+    and the ability to create/delete users within the system. Membership in this organization
+    will be tightly controlled and will be limited to PFB personnel who are responsible
+    for activating/deactivating neighborhood sites and creating initial admin users
+    within a neighborhood organization.
```
CloudNiner: As far as I can tell, we'll eventually need a structure where PFB is the only superuser org, and they are able to create other non-superuser organization accounts that they can hand out to the various cities. Either way, I don't think the third level is necessary. I think what we have here pretty much meets the requirements above, we may need to tweak the permissions classes farther down the line. For the April 1 deadline, that's not necessary, since we can run all jobs from the root org account.

src/django/pfb_analysis/models.py
CloudNiner: Not directly with model field options at the database level, because of the way Django handles CharFields. I updated the model's save method to call `full_clean()` which forces validation before save. This doesn't happen by default in django for reasons.

CloudNiner:
Updated, could use another look. I also added a FK to Organization to Neighborhood so that we can scope unique neighborhoods to the org that uploads them.

src/django/pfb_analysis/models.py
```diff
@@ -19,15 +21,24 @@ class Neighborhood(models.Model):
     """Neighborhood boundary used for an AnalysisJob """
 
     def __repr__(self):
-        return "<Neighborhood: {}, ({})>".format(self.name, self.label)
+        return "<Neighborhood: {} ({})>".format(self.name, self.organization.name)
 
-    name = models.SlugField(max_length=256, primary_key=True)
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    name = models.SlugField(max_length=256, unique=True)
```
CloudNiner: Derp. Yup, good catch.

CloudNiner: Looking at this again before merging, I'm pretty sure that the SlugField is now superfluous. UUID for uniqueness at db level + unique_together on name (as human-readable CharField) + org is sufficient for our purposes. Thoughts?

Issue #62, opened 2017-02-09
Switch to use of Docker container for running analysis in development


CloudNiner:
- [x] Remove all ansible configuration around setting up the VM to run analysis jobs directly
- [x] Update project README 'Running the Analysis' section with docker commands
- [x] Update commands in #10

Issue #10, opened 2016-11-01
Running other neighborhoods


CloudNiner:
## Germantown, Philadelphia, PA

_Note:_ For this job, ensure you first put the gtown_westside.osm file (on FILESHARE) into your `./data` directory. Alternatively, remove the `PFB_OSM_FILE` var from the command below.
```bash
docker run \
    -e PFB_SHPFILE=/data/gtown_westside.shp \
    -e PFB_STATE=pa -e PFB_STATE_FIPS=42 \
    -e NB_INPUT_SRID=4326 \
    -e NB_BOUNDARY_BUFFER=50 \
    -e PFB_OSM_FILE=/data/gtown_westside.osm \
    -e NB_OUTPUT_DIR=/data/output \
    -v /vagrant/data/:/data/ \
    pfb-analysis > pfb-analysis/germantown.log 2>&1
```

## Boulder CO

```bash
docker run \
    -e PFB_SHPFILE=/data/boulder.shp \
    -e PFB_STATE=co \
    -e PFB_STATE_FIPS=08 \
    -e NB_INPUT_SRID=4326 \
    -e NB_BOUNDARY_BUFFER=3600 \
    -e NB_OUTPUT_DIR=/data/output \
    -v /vagrant/data/:/data/ \
    pfb-analysis > pfb-analysis/boulder.log 2>&1
```

CloudNiner:
These commands should be run inside the VM in the `/vagrant` dir

CloudNiner:
If running the docker version, ensure your analysis image is up to date by first running `./scripts/update` or `docker build -t pfb .` within the VM.

flibbertigibbet:
Due to the mount point, I think the first variable should be set like `-e PFB_SHPFILE=/data/boulder.shp` (root directory is `/data`, not `/vagrant`).

CloudNiner:
Indeed. Updated.

CloudNiner:
Updated:
- Use meters as units for NB_BOUNDARY_BUFFER
- Update name of docker image tag (`pfb` -> `pfb-analysis`) to match changes in #106

CloudNiner:
This issue is superseded by README.RUNNING.md added in #136

PR #65, opened 2017-02-10
Feature/awf/dockerize cleanup

CloudNiner:
## Overview

Contains the remaining cleanup around switching to a dockerized analysis run. 

Once this is merged, I will delete the original commands in #10. I left both in for now, since folks could potentially have both strategies provisioned.

## Testing

- Destroy your VM
- Follow the updated README instructions for "Getting Started", then "Running the Analysis". 

You should be able to cleanly run the analysis via docker.

Connects #62


CloudNiner:
It appears that when I run an analysis job via docker, the container cannot be killed via ctrl+c once the analysis completes, nor does it exit on its own. I had to do `docker stop <containerid>` to kill it.

PR #66, opened 2017-02-10
Switch to ansible-local provisioning

CloudNiner:
## Overview

Switch to ansible-local provisioning so that the VM can be used easily on all three major OSs. 

## Testing

`vagrant destroy` then `vagrant up --provision`. The VM should still provision successfully, using ansible-local instead of ansible.

This is untested on Windows ATM. Will attempt to test on Windows over the weekend at home and update as necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@ PFB Bicycle Network Connectivity
 Requirements:
 - Vagrant 1.8+
 - VirtualBox 4.3+
-- Ansible 2.0+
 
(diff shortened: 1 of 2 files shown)


CloudNiner:
@sharph you mentioned at scrum that you're working on the same thing. If this is useful as is, I'll merge it, since its used in #69

sharph:
This is useful as is. You should merge it, and I'll rebase my PR on top of it.

PR #67, opened 2017-02-13
Update DB package versions

flibbertigibbet:
Fix provisioning by updating package version globs for PostgreSQL and PostGIS.


CloudNiner:
#65 should make this PR obsolete, I'm going to merge that one, hopefully there won't be issues with the Dockerized version of the analysis.

flibbertigibbet:
Closing in favor of #69. FYI @sharph in case we need this later this PR was fixing an apparent typo where the minor version number got the wildcard, instead of the patch (changing to do what you suggest.)

PR #70, opened 2017-02-14
Remove output SRID from README

flibbertigibbet:
Now set as part of scripts in azavea/pfb#17.


CloudNiner:
Nice. I also updated the example commands in #10 to only give the docker command and to remove the output SRID.

PR #71, opened 2017-02-15
Merge pfb-analysis into this repo (and reorganize it)

KlaasH:
**The main event**
Replaces the submodule link to the pfb repo with a copy of that repo (including history) imported into this one.

I followed the procedure from: http://stackoverflow.com/a/8396318/6338892, i.e. I ran:
```             
git remote add pfb_analysis git@github.com:azavea/pfb.git
git fetch pfb_analysis  
git merge -s ours --no-commit pfb_analysis/master
git read-tree --prefix=pfb-analysis/ -u pfb_analysis/master
```                                 

Also                                        
- removes the `.gitmodules` file                
- removes the submodule update command from `scripts/setup`

**The bonus bit**
Moves all the files that were sitting in the root directory of `pfb` into directories representing rough categories (there was already `connectivity`, now there's also `import`, `features`, `stress`, and `scripts`).

**To test**
Git won't want to check this out over top of the submodule, so you'll have to do `rm -rf pfb-analysis` before checking out the branch.  After that everything should behave exactly like before.


KlaasH:
I added two changes to `entrypoint.sh`.  They're not specific to this change, but since it's this PR that brings `entrypoint.sh` into this repo, easier to do them here than separately.

One exports the variables, because without that my census blocks were getting imported in 4326 (and a buffer of 11000 in lat/lon is too big).

The other adds `bash` at the bottom of `entrypoint.sh` to keep containers that are run interactively (with `docker run -t`) from exiting, as discussed yesterday on Slack.

CloudNiner:
Nice changes. Moving into sub-directories for organization is good and the breakup of files makes sense. 

Thanks for the documentation, glad to see this was relatively straightforward. A few additional tasks to complete:
- [ ] duplicate any remaining issues in the pfb repo to this one
- [ ] either delete old forked pfb repo or add a note to the README there indicating it is deprecated in favor of continuing development in this repo

CloudNiner:
@spencerrecneps note that once this is merged you'll have to follow these instructions to update your fork as well after pulling develop:

"Git won't want to check this out over top of the submodule, so you'll have to do rm -rf pfb-analysis before checking out the branch. After that everything should behave exactly like before."

PR #69, opened 2017-02-14
Provision successfully on Windows

CloudNiner:
## Overview

@spencerrecneps will begin working on our forked repository rather than his original upstream source. This requires that he be able to run the analysis from a Windows host. This PR updates the repo with instructions and changes to support this. I tested the setup instructions through building the Docker pfb container in the VM.

## Demo

screenshot 1

## Notes

Includes commit to switch to Ansible local-provisioning. Supersedes #66 

Changes to make so far:
- [x] Add instruction to create the `~/.aws` dir and generate credentials if they don't exist
- [x] Add instruction to `cd ./pfb-analysis` before manually running `docker build`

## Testing

Gotta have a Windows host. If you do, install everything and follow the README instructions.

Connects #63

PR #72, opened 2017-02-15
Fix reference to ./scripts/setup in README.md

spencerrecneps:
Just fixing a typo I noticed in the docs as I was setting up the VM.


CloudNiner:
Whoops, good catch. I saw this when working on #69 but forgot to include. Thanks for the PR!

Issue #61, opened 2017-02-09
Django container should receive an 'environment' ENV variable


CloudNiner:
That can be used to configure settings based on environment. In development, this variable should be the value of the host's user account (e.g. `ENV.USER`) so that dev envs get isolated AWS resources.

PR #73, opened 2017-02-15
Add DEV_USER environment variable to separate AWS dev environments

sharph:
## Overview

In order to separate development resources on AWS (such as S3 buckets), we need to pass the developer's username from their host machine through the vagrant VM, into the docker container, and finally into the django instance.

This pull request accomplishes this by passing the username from the host environment to Ansible through the `Vagrantfile`. Ansible writes this to `/etc/environment` as the `DEV_USER` variable. When `docker-compose up` is run, this environment variable is passed into the container as specified in `docker-compose.yml`.

## Testing

* Reprovision
* `docker-compose up`
* While `docker-compose` is running, `docker exec -it vagrant_django_1 /bin/bash`
* `./manage.py shell`
* `from django.conf import settings; print settings.DEV_USER`
* Ensure your username is printed

Connects #61

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -51,7 +51,9 @@ Vagrant.configure("2") do |config|
     ansible.playbook = "deployment/ansible/pfb.yml"
     ansible.galaxy_role_file = "deployment/ansible/roles.yml"
     ansible.verbose = true
-    ansible.raw_arguments = ["--timeout=60"]
+    ansible.raw_arguments = ["--timeout=60",
+                             "--extra-vars",
+                             "dev_user=#{ENV.fetch("USER", "vagrant")}"]
 
     # local arguments
     # Ubuntu trusty base box already has system python + pip installed, no need to reinstall here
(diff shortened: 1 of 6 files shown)


CloudNiner:
It might make sense to namespace this var with a `PFB` prefix, as a good practice.

sharph:
Can you provide more discussion about why this is an appropriate place to namespace with a `PFB` prefix?

Other ENV namespaces already in the code include `POSTGRES_` and `DJANGO_`. The use and intention of this var doesn't seem limited to any of those.

CloudNiner:
It would provide an additional application level namespace, for general env settings that aren't appropriate for service level configurations. Not necessary, just a suggestion as it would easily allow someone to see that its an application setting without making the assumption that unprefixed vars are application level. Not necessary, just a suggestion that could potentially improve clarity.

sharph:
I'm going to merge as is. I can't quite rationalize namespacing this var, as it exists to bring in the not-namespaced `USER` variable from the host. We can revisit if it introduces confusion.

Issue #26, opened 2016-11-17
Add organization models


CloudNiner:
Copy from PEW ENRR

PR #84, opened 2017-02-17
Add users and organizations

flibbertigibbet:
Connects #25 and connects #26.

Adds user and organization models to the django app, largely copied from [PEW ENRR](https://github.com/azavea/ENRR/tree/develop/repository/django). Renamed 'state' field to more generic 'area' for geographic area being analyzed by an organization. Adds custom migration that adds a default user and organization (which differs significantly from ENRR's multiple migrations).

To test, bring up a fresh Django container, then set up with:
```
./scripts/django-manage migrate
./scripts/django-manage collectstatic
```
Then should be able to log in to log in to http://localhost:9200/api/ as: systems+pfb@azavea.com / root


CloudNiner:
Added @KlaasH as a reviewer since he also has familiarity with the User/Org model setup.

src/django/base/filters.py
```diff
@@ -0,0 +1,62 @@
...
+    """Filter used on users endpoint to limit queryset to only user if user is not admin"""
+
+    def filter_queryset(self, request, queryset, view):
+        if is_admin(request.user):
+            return queryset
+        else:
+            return queryset.filter(uuid=request.user.uuid)
+
+
+class AreaAutoFilterBackend(filters.BaseFilterBackend):
```
CloudNiner: These Area filters are probably not necessary since a Neighborhood is scoped to an AnalysisJob, not the org itself.

src/django/users/models.py
```diff
@@ -0,0 +1,215 @@
...
+    USERNAME_FIELD = 'email'
+    REQUIRED_FIELDS = ['organization', 'role']
+
+    email = models.EmailField(_('email address'),
+                              unique=True,
+                              help_text=_('Required. A valid email address'),
+                              error_messages={'unique': _("A user with that email already exists")})
+    first_name = models.CharField(_('first name'), max_length=30, blank=True)
+    last_name = models.CharField(_('last name'), max_length=30, blank=True)
+    organization = models.ForeignKey(Organization)
```
CloudNiner: This should have an explicit `on_delete` set.

flibbertigibbet:
Addition of `base` app and change to bootstrap app was to deal with circular dependencies, as well as being how ENRR is structured. I'll give it another pass to see if it's workable to switch back.

Issue is with the auditing; `PFBModel` (the only thing in `base`) depends on `users`, while users/orgs are themselves audited `PFBModel`s.

KlaasH:
A few comments so far that aren't very line-specific:
- Do we need `Area` and `OrganizationTypes`?  Seems like the user and org structure for this project is a bit simpler and we might be better off removing them and adding back what we need when we figure out what that is than adapting concepts that might not fit that well.
- There's some port mis-match that's not new with these changes but is related: the django Dockerfile has `EXPOSE 9210`, which I think is supposed to be `EXPOSE 9202`, and the Vagrantfile is forwarding port 9210 to both 9202 (or `PFB_GUNICORN_PORT`) and 9210 (or `PFB_WEBPACK_PORT`).  So the first one should be changed to 9202 as well.

src/django/pfb_network_connectivity/models.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+from __future__ import unicode_literals
+
+from django.db import models
+
+from base.models import PFBModel
+
+
+class Area(models.Model):
```
KlaasH: Further to my brief comment below: Area, adapted from State, isn't really the same thing as Neighborhood, as far as I understand.  State was an attribute of `Organization` that served to link related organization entities together and had permissions implications.  Neighborhood is, I think, linked to a specific analysis or series of analysis but isn't something that's independent of an organization or has a useful meaning across organizations.

So I think if we want/need organization hierarchy we should go with something named to reflect what it's doing (like `Group`, or ideally something  a little more specific), and if we don't we can get rid of it.

src/django/base/urls.py
```diff
@@ -20,13 +20,28 @@
 
 from rest_framework import routers
 
+from users import views as user_views
+
 router = routers.DefaultRouter()
 
+router.register(r'users', user_views.PFBUserViewSet, base_name='users')
+
```
KlaasH: users.views has `OrganizationViewSet` as well, so could add that here:
```
router.register(r'organizations', user_views.OrganizationViewSet, base_name='organizations')
```

flibbertigibbet:
Okay, updated to remove `base` app, rename `Area` to `Neighborhood`, protect `Organization` deletion if users still present, and updated docker port.

Issue #74, opened 2017-02-15
Investigate switching network algorithms to pgr_djikstra


spencerrecneps:
Refer to https://github.com/azavea/pfb/pull/18

Test performance improvements from connecting census blocks via pgr_dijkstra instead of pgr_drivingdistance.

spencerrecneps:
@flibbertigibbet Any thoughts on switching to [pgr_johnson()](http://docs.pgrouting.org/2.2/en/src/allpairs/doc/pgr_johnson.html#pgr-johnson)? I'm finding that running pgr_dijkstra on each block pair is no more efficient (and possibly much less efficient) than the previous approach.

I believe pgr_johnson() could be used to populate the connected_roads* tables much more efficiently than either of our other approaches. The drawback is that it appears to be a massive resource hog (which makes sense since we're effectively storing the whole graph in memory for the operation). Might also present problems for parallelization.

spencerrecneps:
For reference I ran pgr_johnson() on roads within the test neighborhood boundary in Cambridge and returned 19 million results in 14 minutes. Not bad. Run time and number of rows would be higher if I expanded to our 11,000 ft max distance but still seems manageable. Trying to run it on the entire region crashed our PostGIS server.

flibbertigibbet:
The intent with trying `pgr_dijkstra` I thought was to filter the origin nodes to only those in bounds, which the driving distance function would not allow. That should reduce run time by about 3/4, as that's how many nodes (for Boulder at least) are in the graph but outside the neighborhood bounds.

spencerrecneps:
I believe the current implementation already filters origin nodes to those within the neighborhood boundary. This is being accomplished by:
```
EXISTS (
            SELECT  1
            FROM    neighborhood_boundary AS b
            WHERE   ST_Intersects(b.geom,r1.geom)
)
```
I think the main source of the performance issues is that we're recreating the entire road graph every time we make a call to either pgr_drivingdistance or pgr_dijkstra. If we go with pgr_johnson we create the graph only once and then store the results in the neighborhood_reachable_roads table.

Come to think of it, that's probably a good way to describe the whole setup: we're effectively saving the graph to a table for quick retrieval later. So pgr_johnson basically lets us dump the graph all at once whereas with pgr_drivingdistance and pgr_dijkstra we're dumping the graph piece by piece.

I've created a branch on my fork to implement pgr_johnson. I'll submit it to you for testing when I'm done and we'll see how it stacks up.

[...]

PR #85, opened 2017-02-17
Feature/connectivity performance improvements

spencerrecneps:
I don't think it's possible to submit a PR that proposes a new branch, but this should remain in a separate branch for now. I can resubmit if it's easier to create a new branch and have me create a new PR.

These changes relate to https://github.com/azavea/pfb-network-connectivity/issues/74. Since I can't run the full thing on my Windows machine I'd like to have someone test this for performance improvements. 

For large neighborhood areas memory usage might be extremely large. In addition, I'm not sure there's a way to parallelize this. If performance improvements are significant enough, we'd need to test on a large area in the target environment to make sure it's still feasible.

### DIFF
--- a/pfb-analysis/connectivity/connected_census_blocks.sql
+++ b/pfb-analysis/connectivity/connected_census_blocks.sql
@@ -45,14 +45,14 @@ SET         high_stress_cost = (
                                 FROM    neighborhood_census_block_roads     roads,
                                         neighborhood_ways_net_vert          verts
                                 WHERE   source_blockid10 = roads.blockid10
-                                        roads.road_id = verts.road_id
+                                AND     roads.road_id = verts.road_id
                             ),
                             (
                                 SELECT  array_agg(verts.vert_id)
                                 FROM    neighborhood_census_block_roads     roads,
                                         neighborhood_ways_net_vert          verts
                                 WHERE   target_blockid10 = roads.blockid10
-                                        roads.road_id = verts.road_id
+                                AND     roads.road_id = verts.road_id
                             ),
                             directed := true
                         )
@@ -71,14 +71,14 @@ SET         high_stress_cost = (
                                 FROM    neighborhood_census_block_roads     roads,
                                         neighborhood_ways_net_vert          verts
                                 WHERE   source_blockid10 = roads.blockid10
-                                        roads.road_id = verts.road_id
+                                AND     roads.road_id = verts.road_id
                             ),
                             (
                                 SELECT  array_agg(verts.vert_id)
                                 FROM    neighborhood_census_block_roads     roads,
                                         neighborhood_ways_net_vert          verts
                                 WHERE   target_blockid10 = roads.blockid10
-                                        roads.road_id = verts.road_id
+                                AND     roads.road_id = verts.road_id
                             ),
                             directed := true
                         )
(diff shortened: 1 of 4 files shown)


CloudNiner:
Thanks for taking a look at this. We're working out some issues with the analysis jobs in development, and building out the environment to run these jobs on AWS. Will likely take a look at this PR once we've got those tasks ironed out. We'll revisit this PR once we've got those other tasks ironed out.

spencerrecneps:
I ran a quick test using the same base data from Cambridge and came up with the following results for elapsed time on reachable_roads_high_stress_calc.sql with a neighborhood buffer of 5000 ft:

With pgr_johnson (modified code in this PR):
Run 1 - 3:32.34
Run 2 - 3:12.95
Run 3 - 3:17.35
Run 4 - 2:57.21
Run 5 - 2:50.93

Existing code:
Run 1 - 4:45.24
Run 2 - 3:54.56
Run 3 - 3:43.94
Run 4 - 3:51.18
Run 5 - 3:44.36

Please note that the existing code did not make use of multiple cores in my testing, so it's possible the performance improvement with pgr_johnson is offset by the ability to distribute the load among cores.

CloudNiner:
Thanks for running these tests. Can you clarify whether the "existing code" measurements are from the current `develop` branch (has the multithreading) or from some older branch that uses the single-threaded implementation, before we made those performance improvements?

spencerrecneps:
I removed the multithreaded parts of the code to run the analysis. I wasn't totally sure how to use the threads in a standalone process and I'm not actually sure how many threads are available on our server anyway.

As I mentioned, it's possible that multithreading can end up closing the performance gap. It's also possible that the performance gap can widen on larger datasets. The pgr_johnson algorithm works by essentially dumping a snapshot of the entire graph into a table. Our current approach mimics this by doing the driving distance analysis node-by-node and dumping the results to a table.

My suspicion is that the added processing time on a large dataset for pgr_johnson is a linear function, whereas the added processing time in the current setup is exponential. Unfortunately, pgr_johnson imposes higher requirements for RAM since you're storing the whole graph in memory and I haven't been able to test larger areas--it crashes our server, which runs out of available memory.

CloudNiner:
Ok. So in that case, for smaller areas at least, the existing solution is better, as we see a 2-3.5x speedup with the multithreaded version. So for initial testing, we'll continue forward with the existing strategy.

With that said, we could use AWS instance types with more RAM, so it might be worth testing down the road. How much RAM does your server have?

CloudNiner:
Also note that the high/low stress SQL files are not the ones that take the longest. The SQL script with the most running time is the `connected_census_blocks`, so further optimization should be focused there.

For example, for Boulder, the high/low scripts run in about ~5min, but `connected_census_blocks.sql` takes ~1hr. So small improvements to high/low connectivity mean much less than a similar improvement to connected_census_blocks.

spencerrecneps:
Our server has 10 GB of RAM.

Good point about optimizing connected_census_blocks.sql. I'll focus future efforts there. We can let this PR sit and perhaps do some testing later.

spencerrecneps:
See https://github.com/azavea/pfb-network-connectivity/pull/153

CloudNiner:
This PR is very out of date, and the optimizations made here were unnecessary for most analyses. If desired, this change can be investigated in a later phase of the project. Closing.

Issue #83, opened 2017-02-17
Add angularjs container with a bootstrapped app


CloudNiner:
This will be the development container for the admin views used to manage and trigger new analysis jobs. 

The UI will be the same as the one developed in enrr since the workflow is identical.

PR #86, opened 2017-02-21
Add angularjs container and parts of an app

KlaasH:
Connects #83.

Removes the Angular 2 stub app and replaces it with an Angular 1.4.9 app with somewhat-working user-management bits and not-currently-functional boundary upload bits.

Note that this contains Area logic that, as noted on PR #84, acts as a cross-organization grouping that might or might not be useful in some form but probably shouldn't be called Area.  I could rip it out altogether from the PR, but there are pieces in the Django app as well, and it might be simpler and more useful to have a separate issue for figuring out what we need and implementing it.

I did some work toward upgrading to Angular 1.5.11 ~~but I figured I'd put this 1.4.9 version up for now since it's hard to predict how successful that will be and how long it will take.~~ and it seemed to go fine as far as the front-end code and dependencies.  It's a pretty simple app at this point, so maybe not surprising.  Upgrading the node/build dependencies was not as clean, so I left them.

image

**To Test**
`./scripts/update` should build the container and `./scripts/server` should run it, at which http://localhost:9301/ should get you a login page that will let you login with systems+pfb@azavea.com/root and do some user and organization management things.


Vagrantfile
```diff
@@ -30,14 +30,11 @@ Vagrant.configure("2") do |config|
...
-  config.vm.network :forwarded_port, guest: 9210, host: ENV.fetch("PFB_WEBPACK_PORT", 9210)
-  config.vm.network :forwarded_port, guest: 9211, host: ENV.fetch("PFB_WEBPACK_LR_PORT", 9211)
-  config.vm.network :forwarded_port, guest: 9212, host: ENV.fetch("PFB_WEBPACK_PROD_PORT", 9212)
-  config.vm.network :forwarded_port, guest: 9213, host: ENV.fetch("PFB_WEBPACK_PROD_LR_PORT", 9213)
+  config.vm.network :forwarded_port, guest: 9202, host: ENV.fetch("PFB_GUNICORN_PORT", 9202)
+  config.vm.network :forwarded_port, guest: 9203, host: ENV.fetch("PFB_RUNSERVER_PORT", 9203)
   config.vm.network :forwarded_port, guest: 5432, host: ENV.fetch("PFB_ANALYSIS_DB_PORT", 9214)
-
+  config.vm.network :forwarded_port, guest: 9301, host: ENV.fetch("PFB_REPOSITORY_PORT", 9301)
+  config.vm.network :forwarded_port, guest: 3001, host: ENV.fetch("PFB_REPOSITORY_BROWSERSYNC", 3001)
```
CloudNiner: Shouldn't this be `9302` as well by default?

CloudNiner:
It looks like the upload expects XML files? Maybe we should change the filter to take only ZIP files?

Vagrantfile
KlaasH: Yes indeedy. Also needed to be fixed in the gulp server config.  I think gulp, docker-compose, Vagrant, and the README are now all on the same page...

CloudNiner:
Check CI, looks like it failed because 'no such service: angular'. 

I think building the angular container in CI is fine for now, I opened a separate issue #95 to ensure that these static assets get built and pushed to staging as part of CI.

KlaasH:
Ok, I believe I'm done fiddling with this.  It now also resolves #95.

PR #88, opened 2017-02-22
Staging Terraform deployment

CloudNiner:
## Overview

Adds the terraform tooling necessary to deploy the staging environment to AWS.

Not much there, just the healthcheck and an API you can't log into yet because I haven't run migrations, which I will do in a bit.

## Demo

https://staging.pfb.azavea.com

## Notes

Peripheral stuff I had to add:
- django-watchman for the healthcheck endpoint
- Fixup settings.py to properly set things based on the actual ENV variables
- Remaining scripts to rule them all
- Update Jenkinsfile (currently untested)

## Testing

After following instructions to set env properly in deployment/README.md, a `./scripts/infra plan` should show no assets to update

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,6 +20,42 @@ node {
         sh 'scripts/cibuild'
       }
     }
+
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/')) {
+      env.AWS_PROFILE = 'pfb'
+      env.AWS_DEFAULT_REGION = 'us-east-1'
+      env.PFB_SETTINGS_BUCKET = 'staging-pfb-config-us-east-1'
+      env.PFB_S3STORAGE_BUCKET = 'staging-pfb-static-us-east-1'
+
+      // Publish container images built and tested during `cibuild`
+      // to the private Amazon Container Registry tagged with the
+      // first seven characters of the revision SHA.
+      stage('cipublish') {
+        // Decode the `AWS_ECR_ENDPOINT` credential stored within
+        // Jenkins. In includes the Amazon ECR registry endpoint.
+        withCredentials([[$class: 'StringBinding',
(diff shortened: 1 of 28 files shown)


CloudNiner:
I just finished running migrations and updated the Staging root account password, which is saved to lastpass in the Shared Urban Apps/PFB folder.

deployment/terraform/dns.tf
```diff
@@ -0,0 +1,48 @@
...
+  type    = "CNAME"
+  ttl     = "10"
+  records = ["${module.database.hostname}"]
+}
+
+#
+# Public DNS resources
+#
+resource "aws_route53_zone" "external" {
+  name = "${var.r53_public_hosted_zone}"
```
tnation14: Are we completely in charge of the DNS for this project? If we aren't, it may give us some more flexibility to manage hosted zones outside of terraform, and use a `r53_hosted_zone_id`, like we do in azavea/fund-equity-map#104

CloudNiner: I'm not sure what the project DNS status is. I think it's likely they'll want to manage that at some point, but I'm inclined to leave it as is for now and adjust later if necessary.

Managing outside terraform makes sense in that case, thanks for noting that.

Jenkinsfile
```diff
@@ -20,6 +20,41 @@ node {
         sh 'scripts/cibuild'
       }
     }
+
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/')) {
+      env.AWS_DEFAULT_REGION = 'us-east-1'
+      env.PFB_SETTINGS_BUCKET = 'staging-pfb-us-east-1-config'
+      env.PFB_S3STORAGE_BUCKET = 'pfb-static-staging'
```
CloudNiner: Gotcha. I'll update these.

CloudNiner:
Updated. Removed unnecessary static site bucket var from variables.tf

PR #106, opened 2017-02-24
Push analysis container to ECR repository on CI build

CloudNiner:
## Overview

Builds and pushes the pfb-analysis container, formerly tagged as `pfb` to a new `pfb-analysis` ECR endpoint.

I renamed the `pfb` container in the README to keep naming conventions consistent.

## Testing

Log into AWS and check to ensure that a tagged container exists in the pfb-analysis registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ In this example, we configure the analysis to be run for Cambridge MA.
 Run:
 ```
 pushd pfb-analysis
-docker build -t pfb .
+docker build -t pfb-analysis .
 popd
 
 docker run \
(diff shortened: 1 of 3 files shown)


CloudNiner:
After merging, update commands in #10

README.md
```diff
@@ -70,7 +70,7 @@ docker run \
     -e NB_INPUT_SRID=2249 \
     -e NB_BOUNDARY_BUFFER=11000 \
     -v /vagrant/data/:/data/ \
-    pfb
+    pfb-analysis
```
KlaasH: Not from a change in this PR, but since we're editing the README it would be good to change the example commands to reflect that the output SRID is now calculated automatically into the appropriate meter-based UTM projection, so this boundary buffer that was intended as 2 miles in feet is now 6.8 miles in meters.

CloudNiner: True. I will update that as well.

Issue #90, opened 2017-02-23
PFB container postgresql.conf work_mem setting


CloudNiner:
Set to 4gb to start

Issue #76, opened 2017-02-15
Large analysis jobs trigger postgresql checkpoints too frequently


KlaasH:
_Moved from https://github.com/azavea/pfb/issues/12, created by @CloudNiner on 2/3/17_

Checkpoint triggered about every ~5 seconds during graph building in TDG functions when postgresql is using its default configuration

PR #107, opened 2017-02-27
`work_mem` and checkpoint-related database settings

KlaasH:
A few tiny tweaks, plus...

Adds some configuration commands to setup_database.sh that:
- per issue #90, increases work_mem (by a lot--4mb to 4gb)
- per issue #76, makes checkpoints happen less frequently
    Note that this used to be controlled by `checkpoint_segments` but is
    now (as of Postgres 9.6) mainly a function of `max_wal_size`.
- Related to checkpoints, I also increased `checkpoint_completion_target`
  which should have the effect of reducing write bottlenecks at the cost
  off less robust crash recovery (which I don't think we care much
  about in this context)

### DIFF
--- a/pfb-analysis/scripts/entrypoint.sh
+++ b/pfb-analysis/scripts/entrypoint.sh
@@ -57,5 +57,5 @@ EOF
 bash
 
 # shutdown postgres
-kill $POSTGRES_PROC
+su postgres -c "/usr/lib/postgresql/9.6/bin/pg_ctl stop"
 wait
--- a/pfb-analysis/scripts/setup_database.sh
+++ b/pfb-analysis/scripts/setup_database.sh
@@ -6,11 +6,16 @@ NB_POSTGRESQL_DB=pfb
 NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
+# Set defaults for overridable configuration params
+PFB_WORK_MEM="${PFB_WORK_MEM:-2048MB}"
+PFB_CHECKPOINT_COMPLETION="${PFB_CHECKPOINT_COMPLETION:-0.8}"
+PFB_MAX_WAL_SIZE="${PFB_MAX_WAL_SIZE:-2GB}"
+
 # Set configuration parameters
 su postgres bash -c psql <<EOF
-ALTER SYSTEM SET work_mem TO '4096MB';
-ALTER SYSTEM SET checkpoint_completion_target TO 0.8;
-ALTER SYSTEM SET max_wal_size TO '5GB';
+ALTER SYSTEM SET work_mem TO '${PFB_WORK_MEM}';
+ALTER SYSTEM SET checkpoint_completion_target TO ${PFB_CHECKPOINT_COMPLETION};
+ALTER SYSTEM SET max_wal_size TO '${PFB_MAX_WAL_SIZE}';
 EOF
 
 # set up database
--- a/pfb-analysis/scripts/setup_database.sh
+++ b/pfb-analysis/scripts/setup_database.sh
@@ -6,6 +6,13 @@ NB_POSTGRESQL_DB=pfb
 NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
+# Set configuration parameters
+su postgres bash -c psql <<EOF
+ALTER SYSTEM SET work_mem TO '4096MB';
+ALTER SYSTEM SET checkpoint_completion_target TO 0.8;
+ALTER SYSTEM SET max_wal_size TO '5GB';
+EOF
+
 # set up database
 su postgres bash -c psql <<EOF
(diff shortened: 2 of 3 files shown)


pfb-analysis/scripts/setup_database.sh
```diff
@@ -6,11 +6,18 @@ NB_POSTGRESQL_DB=pfb
 NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
+# Set configuration parameters
+su postgres bash -c psql <<EOF
+ALTER SYSTEM SET work_mem TO '4096MB';
```
CloudNiner: Do these actually take effect? The docs states: "Values set with ALTER SYSTEM will be effective after the next server configuration reload, or after the next server restart in the case of parameters that can only be changed at server start. A server configuration reload can be commanded by calling the SQL function pg_reload_conf(), running pg_ctl reload, or sending a SIGHUP signal to the main server process."

I don't think we reload the pg conf, and I don't see that it happens anywhere in the base container's [documentation notes](https://hub.docker.com/_/postgres/).

pfb-analysis/scripts/setup_database.sh
```diff
@@ -6,11 +6,18 @@ NB_POSTGRESQL_DB=pfb
...
+ALTER SYSTEM SET checkpoint_completion_target TO 0.8;
+ALTER SYSTEM SET max_wal_size TO '5GB';
+EOF
+
 # set up database
 su postgres bash -c psql <<EOF
-CREATE USER gis WITH PASSWORD 'gis';
-ALTER USER gis WITH SUPERUSER;
-CREATE DATABASE pfb WITH OWNER gis ENCODING 'UTF-8';
+CREATE USER ${NB_POSTGRESQL_USER} WITH PASSWORD '${NB_POSTGRESQL_PASSWORD}';
```
CloudNiner: Good catch.

pfb-analysis/scripts/entrypoint.sh
```diff
@@ -57,5 +57,5 @@ EOF
 bash
 
 # shutdown postgres
-kill $POSTGRES_PROC
+su postgres -c "/usr/lib/postgresql/9.6/bin/pg_ctl stop"
```
CloudNiner: Assuming this is just so that the server process is shut down cleanly? Were you running into issues with the previous implementation?

pfb-analysis/scripts/setup_database.sh
CloudNiner: Just ran the new container, it does look like the server process restarts after running `setup_database.sh`.

pfb-analysis/scripts/setup_database.sh
```diff
@@ -6,11 +6,18 @@ NB_POSTGRESQL_DB=pfb
 NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
+# Set configuration parameters
+su postgres bash -c psql <<EOF
+ALTER SYSTEM SET work_mem TO '4096MB';
+ALTER SYSTEM SET checkpoint_completion_target TO 0.8;
+ALTER SYSTEM SET max_wal_size TO '5GB';
```
CloudNiner: Does it make sense for each of these values to be configurable via ENV?

pfb-analysis/scripts/setup_database.sh
CloudNiner: Setting this to always be 4gb doesn't currently work in dev, as the VM only has 4gb of mem assigned to it. I'm inclined to resolve this by defaulting this setting to 2gb and then allowing for it to be bumped via ENV, rather than further increasing the resources allocated to the dev VM (since additional resources would only be necessary when running large analyses, which we likely won't often do in dev).

pfb-analysis/scripts/entrypoint.sh
KlaasH: Yeah, I was getting analysis containers not exiting, and when I opened a shell and ran `ps` there were postgres processes that seemed to just be hanging out.  Maybe that's not happening anymore for other reasons, but this seems no worse than the `kill` in any case.

pfb-analysis/scripts/setup_database.sh
CloudNiner: This setting also causes adverse effects in dev and probably will also do so in production because we'll be a bit limited on storage space there too, assuming we can reconfigure as described below.

This setting allows the WAL files to grow to 5Gb. Docker, when using the `devicemapper` storage driver, defaults to volumes capped at 10G. So we use up half of our default available storage space holding WAL files instead of the analysis data. I haven't tested if theres a performance hit to leaving the `max_wal_size` low, but that's almost certainly preferable to using up 5G of storage in dev for transaction logs, which we don't care about. Doubly so when the VM volume is limited to 40G total without significant, not automatically provisionable workarounds.

Either way, I'm not quite sure yet of two things:
- Whether we want this setting at all, since it balloons disk usage at the expense of (maybe) a slight performance increase
- Whether this is the root cause of the analysis run failures I'm working on in #109

PR #110, opened 2017-02-28
Add -I flag to shp2pgsql calls to build spatial index on tables

spencerrecneps:
I noticed we're not building a spatial index on the neighborhood or the census tables when we import them. It's possible these are being built elsewhere, but it seems easiest to build them on import with the -I flag.

### DIFF
--- a/pfb-analysis/import/import_neighborhood.sh
+++ b/pfb-analysis/import/import_neighborhood.sh
@@ -51,7 +51,7 @@ then
         NB_STATE_FIPS="${2}"
 
         # Import neighborhood boundary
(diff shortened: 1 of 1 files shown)


CloudNiner:
If `-I` builds the index, is it still necessary to build the index elsewhere? Can that code be removed?

spencerrecneps:
As far as I can tell the indexes are simply not being built. I didn't have anything in my original code to handle it because I was manually uploading these datasets.

flibbertigibbet:
This would be a great addition. Yes, I noticed there were no GIST indices on `neighborhood_ways` or `neighborhood_boundaries`, and added them in https://github.com/azavea/pfb/pull/18/files (unmerged). So I don't think there would be anything to remove @CloudNiner.

PR #112, opened 2017-03-01
Enable local copies of downloads

KlaasH:
These are changes I made in the course of working on PR #107 because I got tired of waiting for the downloads over and over (the PA tabblock file is 247MB...), and they seem useful enough and harmless enough to commit.

For the jobs and blocks files, this just makes it so that if the exact filename the script is looking to download already exists in the `data` directory, it will use that.  If it doesn't, it acts normally.  It doesn't copy the script once downloaded into the data dir to be used for next time, just uses it if it's there, so to take advantage of this you have to download it manually (or copy it out of a container that downloaded it, but that's tricky because it gets deleted as soon as it's been used...).

For the OSM file, since it's based on the boundary rather than being standard for a state, this adds an environment variable `PFB_OSM_FILE` that will be used if provided in lieu of downloading from Overpass.  This has the added benefits of reducing requests to Overpass and allowing us to do multiple runs for comparison without the risk that changes to live OSM data would affect the results.

### DIFF
--- a/pfb-analysis/import/import_jobs.sh
+++ b/pfb-analysis/import/import_jobs.sh
@@ -38,8 +38,12 @@ function import_job_data() {
     NB_DATA_TYPE="${2-main}"    # Either 'main' or 'aux'
     NB_JOB_FILENAME="${NB_STATE_ABBREV}_od_${NB_DATA_TYPE}_JT00_2014.csv"
 
-    wget -P "${NB_TEMPDIR}" "http://lehd.ces.census.gov/data/lodes/LODES7/${NB_STATE_ABBREV}/od/${NB_JOB_FILENAME}.gz"
-    gunzip -c "${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz" > "${NB_TEMPDIR}/${NB_JOB_FILENAME}"
+    if [[ -f "/data/${NB_JOB_FILENAME}.gz" ]]; then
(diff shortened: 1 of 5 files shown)


pfb-analysis/import/import_jobs.sh
```diff
@@ -38,8 +38,12 @@ function import_job_data() {
     NB_DATA_TYPE="${2-main}"    # Either 'main' or 'aux'
     NB_JOB_FILENAME="${NB_STATE_ABBREV}_od_${NB_DATA_TYPE}_JT00_2014.csv"
 
-    wget -P "${NB_TEMPDIR}" "http://lehd.ces.census.gov/data/lodes/LODES7/${NB_STATE_ABBREV}/od/${NB_JOB_FILENAME}.gz"
-    gunzip -c "${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz" > "${NB_TEMPDIR}/${NB_JOB_FILENAME}"
+    if [[ -f "../../data/${NB_JOB_FILENAME}.gz" ]]; then
+        gunzip -c "../../data/${NB_JOB_FILENAME}.gz" > "${NB_TEMPDIR}/${NB_JOB_FILENAME}"
```
CloudNiner: Since this is running in the container, which is mounted to `/data`, maybe the relative dir prefixes go away and we just check `/data/${NB_JOB_FILENAME}`?

Issue #80, opened 2017-02-15
Add export_connectivity.sh script to Docker analysis


KlaasH:
_Moved from https://github.com/azavea/pfb/issues/20, created by @CloudNiner on 2/9/17_

Should be configured to save export data somewhere

PR #116, opened 2017-03-01
Run export connectivity in docker analysis

KlaasH:
Connects #80.

Adds export_connectivity.sh to entrypoint.sh, using `NB_OUTPUT_DIR`.

Notes:
- It doesn't define a default nor does it raise an error at the beginning of the run if the variable isn't set. This could be a pain, since we could run analyses and then end up not getting the results, but on the other hand once we've built further tooling around running the analysis it might not be a problem in practice.  In any case there are ways to get into an image and re-run the export if necessary.
- It does, however, run `mkdir -p NB_OUTPUT_DIR`, so if you give it a path that doesn't exist but does make sense, it create and use that directory (regular mkdir throws an error if a directory already exists, but with `-p` it's fine)
- This branch also includes a change to `import_neighborhood.sh` to speed up how the boundary and census blocks shapefiles get loaded.
- With that and the "pre-downloaded files" trick in PR #112, it's possible to run the littlest analysis run in, on my machine, 2 minutes.

***The littlest analysis run***
Not intended to produce useful or correct results, but it provides a way to test the basic functionality of options and code changes (i.e. whether things run and produce results) without having to wait forever.  (Except it doesn't produce school results. There are schools in the OSM data but I don't think any of them are actually inside the boundary+tiny buffer.)

Steps:
1. Get `gtown_westside.osm` and `gtown_westside.zip` from the fileshare, copying the former and extracting the latter into your `data` directory.
2. In a shell in your `data` directory, run:
```
wget http://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/tabblock2010_42_pophu.zip
wget http://lehd.ces.census.gov/data/lodes/LODES7/pa/od/pa_od_aux_JT00_2014.csv.gz
wget http://lehd.ces.census.gov/data/lodes/LODES7/pa/od/pa_od_main_JT00_2014.csv.gz
```
3. In `/vagrant/` in the VM, run:
```
docker run -e PFB_SHPFILE=/data/gtown_westside.shp -e PFB_STATE=pa -e PFB_STATE_FIPS=42 \
    -e NB_INPUT_SRID=4326 -e NB_BOUNDARY_BUFFER=50 \
    -e PFB_OSM_FILE=/data/gtown_westside.osm -e NB_OUTPUT_DIR=/data/output \
    -v /vagrant/data/:/data/ pfb-analysis
```

### DIFF
--- a/pfb-analysis/import/import_neighborhood.sh
+++ b/pfb-analysis/import/import_neighborhood.sh
@@ -41,6 +41,20 @@ NB_POSTGRESQL_PASSWORD - Default: gis
 "
 }
 
+# Function to import a shapefile (using 'dump' mode for quickness) and convert it to the target SRID
+function import_and_transform_shapefile() {
+    IMPORT_FILE="${1}"
+    IMPORT_TABLENAME="${2}"
+
+    echo "START: Importing ${IMPORT_TABLENAME}"
+    shp2pgsql -I -d -D -s 4326 "${IMPORT_FILE}" "${IMPORT_TABLENAME}" \
+        | psql -h "${NB_POSTGRESQL_HOST}" -U "${NB_POSTGRESQL_USER}" -d "${NB_POSTGRESQL_DB}" > /dev/null
+    psql -h "${NB_POSTGRESQL_HOST}" -U "${NB_POSTGRESQL_USER}" -d "${NB_POSTGRESQL_DB}" \
+        -c "ALTER TABLE ${IMPORT_TABLENAME} ALTER COLUMN geom \
+            TYPE geometry(MultiPolygon,${NB_OUTPUT_SRID}) USING ST_Transform(geom,${NB_OUTPUT_SRID});"
+    echo "DONE: Importing ${IMPORT_TABLENAME}"
+}
+
 if [ "${BASH_SOURCE[0]}" = "${0}" ]
 then
     if [ "${1:-}" = "--help" ] || [ -z "${1:-}" ]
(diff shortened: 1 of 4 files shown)


CloudNiner:
In order to get OSM data to load from the local file, I had to change `-e PFB_OSM_FILE=gtown_westside.osm` to `-e PFB_OSM_FILE=/data/gtown_westside.osm`

KlaasH:
Was the NB_DATA_TYPE initialization not working for you?  Just curious--the way you changed it to is right, it's just confusing that it was working, at least for me, when written wrong.

And yeah, evidently I copied in the wrong iteration of my command.  Thanks for noting.  I edited the initial comment to reflect that.

CloudNiner:
Yeah, import_jobs.sh was working fine as far as I could tell, both with and without the fix. Which is surprising, I would expect it to not.

CloudNiner:
Added germantown instructions to https://github.com/azavea/pfb-network-connectivity/issues/10

pfb-analysis/import/import_neighborhood.sh
```diff
@@ -41,6 +41,20 @@ NB_POSTGRESQL_PASSWORD - Default: gis
 "
 }
 
+# Function to import a shapefile (using 'dump' mode for quickness) and convert it to the target SRID
+function import_and_transform_shapefile() {
+    IMPORT_FILE="${1}"
+    IMPORT_TABLENAME="${2}"
+
+    echo "START: Importing ${IMPORT_TABLENAME}"
+    shp2pgsql -I -d -D -s 4326 "${IMPORT_FILE}" "${IMPORT_TABLENAME}" \
```
CloudNiner: Any specific reason to revert to static SRID 4326 here over using `-s "${NB_INPUT_SRID}"`? The cambridge neighborhood boundary is in a different projection.

I made the change back and it doesn't seem to have affected anything so far.

CloudNiner: Oh. I see. Potentially different projections for the block import and the neighborhood import. I'll fix this up.

KlaasH: Doh, I think I was focused on the blocks one initially because it was the slowest then didn't note the difference when converting the boundary import.  Sorry about that.

Issue #118, opened 2017-03-02
Increase docker volume size in AWS


CloudNiner:
In development, we provision docker with `--storage-opt dm.basesize=20G` to give us a bit more storage space to work with on the docker volume. Production jobs will require up to ~120GB of disk space, so we need to ensure we set the same parameter in production as appropriate, or investigate an alternate storage driver, since the default in ECS is also devicemapper due to the RHEL base image.

Potentially helpful documentation:
- https://docs.docker.com/engine/userguide/storagedriver/selectadriver/
- https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/
- http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html#ecs-ami-storage-config

hectcastro:
Has this become a non-issue now?

Issue #115, opened 2017-03-01
Improve performance of queries in neighborhood_connected_census_blocks


CloudNiner:
Particularly investigate https://github.com/azavea/pfb-network-connectivity/blob/develop/pfb-analysis/connectivity/connected_census_blocks.sql#L46

spencerrecneps:
I made major changes to this in https://github.com/azavea/pfb-network-connectivity/pull/85. I'm hoping performance is improved but it comes at the expense of high memory usage.

CloudNiner:
Those changes affect the high/low stress connectivity sql scripts. Do the changes made there mean that the connected_census_blocks sql runs on a much smaller area?

spencerrecneps:
Ah you're right. I had initially tried some changes on this file and then reverted in favor of improving the high/low stress scripts. Carry on...

CloudNiner:
Improvements were made to this in #158, consider additional changes after testing.

Issue #81, opened 2017-02-16
Rename nb_boundary_buffer environment variable


spencerrecneps:
The name nb_boundary_buffer is a little misleading. I suggest renaming to something like max_travel_distance. In some cases we use the value as a buffer on the neighborhood but it's really just an assumption about the maximum travel distance we want to allow.

CloudNiner:
rename to `NB_MAX_TRAVEL_DISTANCE` and adjust the parameter value accordingly.

KlaasH:
See https://github.com/azavea/pfb-network-connectivity/pull/119#discussion_r104037827.  We should truncate the census blocks table to half the max travel distance.

PR #119, opened 2017-03-02
Feature/awf/fix docker analysis

CloudNiner:
## Overview

Here we have a grab bag of fixes that hopefully resolve some of the inconsistencies of running large analyses in the virtualized docker container. This issue was mostly a result of bugfixes that fixed up the parameterization of the buffer distance in connected_census_blocks, along with further tweaks to postgresql settings that constrain database resources as are appropriate for the VM by default.

I was able to successfully run Germantown. Also Boulder, with a reduced boundary bufffer of 1000m and the normal 2mi buffer (3600m). The 3600m Boulder analysis ran in 1hr30m, which may be a bit slower than before (we have docs that state 2mi Boulder ran in 1h10m), but it no longer crashes by running out of mem/disk when attempting to calculate connected_census_blocks.sql. Before I parameterized buffer distance in https://github.com/azavea/pfb-network-connectivity/commit/1c19e635d6c03d63ccca075dc34981c9231e5605 a Boulder run with a buffer of 11000m successfully ran in 8hr (!). But it ran without crashing.

_EDIT_: 3600m Cambridge just ran, but hit the artificially imposed `temp_file_limit=10G` limit while calculating `connected_census_blocks`, which prevented the analysis from crashing, but key metrics are missing in `overall_scores`. I'm not sure there's a way around this issue currently, given the resource constraints of the VM. It's likely we'll need to come up with separate performance improvements there, which is blocked out as a separate task in #115  

### Notes

The change to docker devicemapper basesize will also need to happen somehow in AWS, opened a separate issue to investigate that: #118

The destruction of the VM as described in 'Testing' is not strictly necessary. However, without it you'll need to tweak the postgresql settings added in this PR appropriately, and run smaller jobs.

Fixes issue identified here: https://github.com/azavea/pfb-network-connectivity/pull/116#discussion_r103973428

## Demo

Boulder 3600m results:
```
id |  category  |                  score_name                   |   score    |                                              notes
----+------------+-----------------------------------------------+------------+--------------------------------------------------------------------------------------------------
  1 | Population | Median population accessible by low stress    |  5342.0000 | Total population accessible by low stress expressed as the median of all census blocks in the   +
    |            |                                               |            |             neighborhood
  2 | Population | Median population accessible by high stress   | 48921.0000 | Total population accessible by high stress expressed as the median of all census blocks in the  +
    |            |                                               |            |             neighborhood
  3 | Population | Median ratio of access to population          |     0.1030 | Ratio of population accessible by low stress to population accessible overall, expressed as     +
    |            |                                               |            |             the median of all census blocks in the                                              +
    |            |                                               |            |             neighborhood
  4 | Population | 70th percentile ratio of access to population |     0.1961 | Ratio of population accessible by low stress to population accessible overall, expressed as     +
    |            |                                               |            |             the 70th percentile of all census blocks in the                                     +
    |            |                                               |            |             neighborhood
  5 | Population | Average ratio of access to population         |     0.1372 | Ratio of population accessible by low stress to population accessible overall, expressed as     +
    |            |                                               |            |             the average of all census blocks in the                                             +
    |            |                                               |            |             neighborhood
  6 | Employment | Median employment accessible by low stress    |  3603.0000 | Total jobs accessible by low stress expressed as the median of all census blocks in the         +
    |            |                                               |            |             neighborhood
  7 | Employment | Median employment accessible by high stress   | 49604.0000 | Total jobs accessible by high stress expressed as the median of all census blocks in the        +
    |            |                                               |            |             neighborhood
  8 | Employment | Median ratio of access to employment          |     0.0730 | Ratio of employment accessible by low stress to employment accessible overall, expressed as     +
    |            |                                               |            |             the median of all census blocks in the                                              +
    |            |                                               |            |             neighborhood
  9 | Employment | 70th percentile ratio of access to employment |     0.1371 | Ratio of employment accessible by low stress to employment accessible overall, expressed as     +
    |            |                                               |            |             the 70th percentile of all census blocks in the                                     +
    |            |                                               |            |             neighborhood
 10 | Employment | Average ratio of access to employment         |     0.1144 | Ratio of employment accessible by low stress to employment accessible overall, expressed as     +
    |            |                                               |            |             the average of all census blocks in the                                             +
    |            |                                               |            |             neighborhood
 11 | Schools    | Average low stress school access              |     2.2588 | Number of schools accessible by low stress expressed as an average of all census blocks in the  +
    |            |                                               |            |             neighborhood
 12 | Schools    | Average high stress school access             |    17.9393 | Number of schools accessible by high stress expressed as an average of all census blocks in the +
    |            |                                               |            |             neighborhood
 13 | Schools    | Average school low stress population shed     |  5332.8276 | Population with low stress access to schools in the neighborhood expressed as an average of all +
    |            |                                               |            |             schools in the neighborhood
 14 | Schools    | Average school high stress population shed    | 39677.5517 | Population with high stress access to schools in the neighborhood expressed as an average of all+
    |            |                                               |            |             schools in the neighborhood
 15 | Schools    | Average school population shed ratio          |     0.1301 | Ratio of population with low stress access to schools to population with high stress access     +
    |            |                                               |            |             in the neighborhood expressed as an average of all                                  +
    |            |                                               |            |             schools in the neighborhood
```

## Testing

The change to the docker storage options requires that the VM be destroyed and reprovisioned, so do that first with `vagrant destroy && ./scripts/setup`.

Then rebuild the pfb-analysis container following current README instructions.

Once the container is built, you should be able to run any of the successful analysis jobs noted above.

If you want to run a long running job, but not block farther development, build the analysis container with a different tag, and then run the analysis using a different tag (substituting in `docker run` as appropriate). Once the container is built, you can safely switch off this branch.

### DIFF
--- a/pfb-analysis/connectivity/reachable_roads_high_stress_calc.sql
+++ b/pfb-analysis/connectivity/reachable_roads_high_stress_calc.sql
@@ -1,7 +1,8 @@
 ----------------------------------------
 -- INPUTS
 -- location: neighborhood
--- maximum network distsance: 10560 ft
+-- :nb_boundary_buffer psql var must be set before running this script,
+--      e.g. psql -v nb_boundary_buffer=11000 -f connected_census_blocks.sql
 ----------------------------------------
 INSERT INTO generated.neighborhood_reachable_roads_high_stress (
     base_road,
@@ -21,7 +22,7 @@ FROM    neighborhood_ways r1,
                     link_cost AS cost
             FROM    neighborhood_ways_net_link',
             v1.vert_id,
-            10560,
+            :nb_boundary_buffer,
             directed := true
         ) sheds
 WHERE r1.road_id % :thread_num = :thread_no
--- a/pfb-analysis/scripts/setup_database.sh
+++ b/pfb-analysis/scripts/setup_database.sh
@@ -7,15 +7,32 @@ NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
 # Set defaults for overridable configuration params
-PFB_WORK_MEM="${PFB_WORK_MEM:-2048MB}"
 PFB_CHECKPOINT_COMPLETION="${PFB_CHECKPOINT_COMPLETION:-0.8}"
+# Disable to improve performance, if disabled, data loss occurs on server crash (ok for analysis)
+PFB_FSYNC="${PFB_FSYNC:-off}"
+# Only one process at a time, can be higher, set to 1/4 of available system memory
+PFB_MAINTENANCE_WORK_MEM="${PFB_MAINTENANCE_WORK_MEM:-1024MB}"
+# WAL is not useful for us, so keep small to limit disk usage
 PFB_MAX_WAL_SIZE="${PFB_MAX_WAL_SIZE:-256MB}"
+# Set to ~1/4 of available system memory
+PFB_SHARED_BUFFERS="${PFB_SHARED_BUFFERS:-1024MB}"
+# Disable to improve performance, _slightly_ safer than fsync=off
+PFB_SYNCHRONOUS_COMMIT="${PFB_SYNCHRONOUS_COMMIT:-off}"
+# Limits disk usage of query temp tables, set to ~1/2 of available disk space in KB, default 10GB
+PFB_TEMP_FILE_LIMIT="${PFB_TEMP_FILE_LIMIT:-10485760}"
+# This is per operation, so can't be massive. For analysis, 1/8 of available system memory
+PFB_WORK_MEM="${PFB_WORK_MEM:-512MB}"
 
 # Set configuration parameters
 su postgres bash -c psql <<EOF
-ALTER SYSTEM SET work_mem TO '${PFB_WORK_MEM}';
 ALTER SYSTEM SET checkpoint_completion_target TO ${PFB_CHECKPOINT_COMPLETION};
+ALTER SYSTEM SET fsync TO ${PFB_FSYNC};
+ALTER SYSTEM SET maintenance_work_mem TO '${PFB_MAINTENANCE_WORK_MEM}';
 ALTER SYSTEM SET max_wal_size TO '${PFB_MAX_WAL_SIZE}';
+ALTER SYSTEM SET shared_buffers TO '${PFB_SHARED_BUFFERS}';
+ALTER SYSTEM SET synchronous_commit TO ${PFB_SYNCHRONOUS_COMMIT};
+ALTER SYSTEM SET temp_file_limit TO ${PFB_TEMP_FILE_LIMIT};
+ALTER SYSTEM SET work_mem TO '${PFB_WORK_MEM}';
 EOF
 
 # set up database
--- a/pfb-analysis/scripts/setup_database.sh
+++ b/pfb-analysis/scripts/setup_database.sh
@@ -6,10 +6,11 @@ NB_POSTGRESQL_DB=pfb
 NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
+PFB_AUTOVACUUM="${PFB_AUTOVACUUM:-off}"
 # Set defaults for overridable configuration params
 PFB_CHECKPOINT_COMPLETION="${PFB_CHECKPOINT_COMPLETION:-0.8}"
 # Disable to improve performance, if disabled, data loss occurs on server crash (ok for analysis)
-PFB_FSYNC="${PFB_FSYNC:-off}"
+PFB_FSYNC="${PFB_FSYNC:-on}"
 # Only one process at a time, can be higher, set to 1/4 of available system memory
 PFB_MAINTENANCE_WORK_MEM="${PFB_MAINTENANCE_WORK_MEM:-1024MB}"
 # WAL is not useful for us, so keep small to limit disk usage
@@ -20,11 +21,14 @@ PFB_SHARED_BUFFERS="${PFB_SHARED_BUFFERS:-1024MB}"
 PFB_SYNCHRONOUS_COMMIT="${PFB_SYNCHRONOUS_COMMIT:-off}"
 # Limits disk usage of query temp tables, set to ~1/2 of available disk space in KB, default 10GB
 PFB_TEMP_FILE_LIMIT="${PFB_TEMP_FILE_LIMIT:-10485760}"
+# Setting to at least a few MB can improve write performance on a server with many commits at once
+PFB_WAL_BUFFERS="${PFB_WAL_BUFFERS:-8192}"
 # This is per operation, so can't be massive. For analysis, 1/8 of available system memory
 PFB_WORK_MEM="${PFB_WORK_MEM:-512MB}"
 
 # Set configuration parameters
 su postgres bash -c psql <<EOF
+ALTER SYSTEM SET autovacuum TO ${PFB_AUTOVACUUM};
 ALTER SYSTEM SET checkpoint_completion_target TO ${PFB_CHECKPOINT_COMPLETION};
 ALTER SYSTEM SET fsync TO ${PFB_FSYNC};
 ALTER SYSTEM SET maintenance_work_mem TO '${PFB_MAINTENANCE_WORK_MEM}';
@@ -32,6 +36,7 @@ ALTER SYSTEM SET max_wal_size TO '${PFB_MAX_WAL_SIZE}';
 ALTER SYSTEM SET shared_buffers TO '${PFB_SHARED_BUFFERS}';
 ALTER SYSTEM SET synchronous_commit TO ${PFB_SYNCHRONOUS_COMMIT};
 ALTER SYSTEM SET temp_file_limit TO ${PFB_TEMP_FILE_LIMIT};
+ALTER SYSTEM SET wal_buffers TO ${PFB_WAL_BUFFERS};
 ALTER SYSTEM SET work_mem TO '${PFB_WORK_MEM}';
 EOF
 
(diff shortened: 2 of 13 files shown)


pfb-analysis/connectivity/reachable_roads_high_stress_calc.sql
```diff
@@ -1,7 +1,8 @@
 ----------------------------------------
 -- INPUTS
 -- location: neighborhood
--- maximum network distsance: 10560 ft
+-- :nb_boundary_buffer psql var must be set before running this script,
+--      e.g. psql -v nb_boundary_buffer=11000 -f connected_census_blocks.sql
```
KlaasH: Worth mentioning here, I think, that the units are those of the the coordinate system of the `neighborhood_ways` tables.

pfb-analysis/scripts/setup_database.sh
```diff
@@ -7,15 +7,32 @@ NB_POSTGRESQL_USER=gis
 NB_POSTGRESQL_PASSWORD=gis
 
 # Set defaults for overridable configuration params
-PFB_WORK_MEM="${PFB_WORK_MEM:-2048MB}"
 PFB_CHECKPOINT_COMPLETION="${PFB_CHECKPOINT_COMPLETION:-0.8}"
-PFB_MAX_WAL_SIZE="${PFB_MAX_WAL_SIZE:-2GB}"
+# Disable to improve performance, if disabled, data loss occurs on server crash (ok for analysis)
+PFB_FSYNC="${PFB_FSYNC:-off}"
```
CloudNiner: 👍

pfb-analysis/scripts/setup_database.sh
```diff
@@ -7,15 +7,32 @@ NB_POSTGRESQL_USER=gis
...
+# Only one process at a time, can be higher, set to 1/4 of available system memory
+PFB_MAINTENANCE_WORK_MEM="${PFB_MAINTENANCE_WORK_MEM:-1024MB}"
+# WAL is not useful for us, so keep small to limit disk usage
+PFB_MAX_WAL_SIZE="${PFB_MAX_WAL_SIZE:-256MB}"
```
CloudNiner: Yeah, all of the moving pieces here were a bit confusing to me. It does not seem possible to completely turn off WAL, even if fsync=off. The [continuous archiving](https://www.postgresql.org/docs/9.6/static/continuous-archiving.html) docs state that "_At all times_, PostgreSQL maintains a write ahead log (WAL)...".

@hectcastro is this something you're more familiar with? We don't need crash recovery since this is a temp database for each analysis run so it just uses up unnecessary system resources.

pfb-analysis/connectivity/reachable_roads_high_stress_calc.sql
CloudNiner: 👍

pfb-analysis/connectivity/reachable_roads_high_stress_calc.sql
```diff
@@ -21,7 +22,7 @@ FROM    neighborhood_ways r1,
                     link_cost AS cost
             FROM    neighborhood_ways_net_link',
             v1.vert_id,
-            10560,
+            :nb_boundary_buffer,
```
CloudNiner: This suggestion is essentially the goal of #81. 

@spencerrecneps if we dropped the boundary buffer to just be 1/2 of max trip distance, would that be a sensible optimization from your side? Currently the boundary buffer == max trip distance, which is more than necessary.

deployment/ansible/group_vars/all
```diff
@@ -30,6 +30,7 @@ postgis_package_version: "2.3.1+dfsg-1.pgdg14.04+1"
 # azavea.docker
 # 1.11.* to match AWS ECS
 docker_version: 1.11.*
+docker_options: "--storage-opt dm.basesize=20G"
```
hectcastro: I see. I don't think it'll be too bad to get ECS to recognize customizations to the storage driver (unsure about Batch though).

pfb-analysis/scripts/setup_database.sh
hectcastro: What is the access pattern we're trying to optimize for here? Bulk import?

I think knowing that will better inform how I comment on the other changes being made in this PR.

CloudNiner: The analysis job essentially does the following:
- Import a bunch of data from various sources (OSM, CSV, SHP)
- Run a series of I/O and memory bound INSERT/UPDATE queries
- Export data via pgsql2shp

hectcastro: I recently played a bunch of games to try and get the `pg_restore` times for Cicero down and settled on the following options, which made a significant difference compared to the defaults:

- `autovacuum off`
- `checkpoint_segments 256`
- `checkpoint_timeout 1800`
- `maintenance_work_mem 1048576` (1GB; might need to tweak given VM resources)
- `synchronous_commit off`
- `wal_buffers 8192`

I'd probably leave `fsync` on and try these settings to see how much better than is from the baseline.

CloudNiner: Currently testing a run with these options merged with our existing ones.

I left our existing `max_wal_size` setting at 256MB. The formula suggested by the 9.5 changelog is `max_wal_size = (3 * checkpoint_segments) * 16MB` which would be ~12GB if we used the `checkpoint_segments` value above, which we don't have the disk space for (nor do I think such a large value would be particularly helpful). Hopefully its not the case that leaving `max_wal_size` unchanged adversely impacts the other changes.

CloudNiner: Most of our queries are initial insert/update queries to tables where no further changes are made once the tables are constructed. So VACUUM should have limited use either way. I'm currently testing without adding manual vacuums to the tables, once this run is done I'll test with manual VACUUMs sprinkled throughout the code.

CloudNiner: The merged settings without doing manual vacuuming were slightly faster and saved about ~5mins in `connected_census_blocks.sql`, dropping run time from 1hr6min to 1hr 1min.

CloudNiner: Adding manual VACUUMs made no difference in run time. I'm going to leave them in since they don't hurt to have if the analysis algorithms change in the future.

CloudNiner:
Updated with additional changes to postgresql settings. Notes in existing comment thread.

CloudNiner:
If no further comments on this, I'm going to merge around noon, since it does provide an improvement for larger runs.

deployment/ansible/group_vars/all
```diff
@@ -30,6 +30,7 @@ postgis_package_version: "2.3.1+dfsg-1.pgdg14.04+1"
 # azavea.docker
 # 1.11.* to match AWS ECS
 docker_version: 1.11.*
```
hectcastro: Default in ECS is now 1.12.*.

deployment/ansible/group_vars/all
hectcastro: After some more investigation here, I think it is possible that we are declaring a bare volume without an explicit mapping to a location on the host file system. Mappings with an explicit location on the host file system should bypass any base volume size constraints with the storage driver.

Where do we define the volume that is running into this constraint?

CloudNiner: Hmmm. We don't. Docker sets up the default volume pool mapped to `/var/lib/docker/volumes`. Can you point me to the docs that would indicate what you describe? Definitely willing to play around with this more, although in the case of Vagrant, we're still limited to the default 40Gb VM volume unless we want to mount docker volumes to a shared folder, which would incur the performance hit and probably isn't worth it for this project.

hectcastro: There really isn't any explicit documentation, but that's my takeaway from the references to bind-mounts in https://docs.docker.com/engine/tutorials/dockervolumes/#mount-a-host-directory-as-a-data-volume.

```bash
ubuntu@ip-172-31-52-80:~$ mkdir joker
ubuntu@ip-172-31-52-80:~$ docker run -ti --rm -v ${PWD}/joker:/joker -w /joker debian:jessie /bin/bash
Unable to find image 'debian:jessie' locally
jessie: Pulling from library/debian
693502eb7dfb: Already exists
Digest: sha256:52af198afd8c264f1035206ca66a5c48e602afb32dc912ebf9e9478134601ec4
Status: Downloaded newer image for debian:jessie
root@2b2bbba847ad:/joker# ls
root@2b2bbba847ad:/joker# fallocate -l 20G joker.img
root@2b2bbba847ad:/joker# ls -lh
total 21G
-rw-r--r-- 1 root root 20G Mar  8 14:58 joker.img
root@2b2bbba847ad:/joker# exit
exit
ubuntu@ip-172-31-52-80:~$ ls -lh joker/
total 21G
-rw-r--r-- 1 root root 20G Mar  8 09:58 joker.img
ubuntu@ip-172-31-52-80:~$ docker info | grep Storage
 WARNING: Usage of loopback devices is strongly discouraged for production use. Use `--storage-opt dm.thinpooldev` to specify a custom block storage device.
WARNING: No swap limit support
Storage Driver: devicemapper
```

CloudNiner: Ah! Gotcha. I see where were missing each other. None of the host mounted volumes are the issue. The problem is the postgreql data dir, which is _not_ a host mounted volume. I _think_ we could convert it to a host mounted volume by setting `PGDATA` and mounting the dir we specify with `docker -v` as described here: https://hub.docker.com/_/postgres/.

In AWS, as long as we mount to the EBS volume, any transient data will disappear when the EBS volume is destroyed on instance termination.

hectcastro: Yeah, I think we need to do something equivalent to:

```
$ docker run -v /mnt/${JOB_ID}:/var/lib/postgresql/data ...
```

Where `/mnt` is some larger amount of storage (ephemeral or EBS). Scoping by job ID, or something else, helps support having multiple jobs like these running alongside each other.

CloudNiner: Exactly. In a separate branch this AM I was able to setup an umanaged compute environment and attach ECS autoscaling/instance launch configs that automatically map an EBS volume on boot via cloud init. So I'll play with this more today over there on that branch rather than here.

I can remove the storageopts setting as that's probably not necessary with the other coming changes.

deployment/ansible/group_vars/all
CloudNiner: I'll update. Will need to update the ECS AMI separately.

Issue #96, opened 2017-02-23
Deploy new version of Analysis batch queue when pfb container changes


CloudNiner:
May need to think about how to handle queue draining of old versions.

CloudNiner:
Depends on #91

Issue #91, opened 2017-02-23
Setup staging AWS Batch environment


CloudNiner:
We'll need to launch and destroy a few of these, a boto script to do it, run from `./scripts/infra` is probably better than manual creation/deletion.

Resources needed:
- [x] AWS Batch managed compute environment
- [x] Analysis job batch queue

End goal of this task is to have the staging environment able to trigger and start an analysis job task via boto, but not necessarily have it complete successfully.

CloudNiner:
EC2 instance choice must be one which has additional storage space attached at instance boot. Good starting choices are `m3.xlarge` or `m3.2xlarge`.

CloudNiner:
TODO for this task:
- [ ] Configure Django with compute env ARN, job queue ARN, job definition name via ENV
- [ ] Add README notes about setting above ENV in development
- [ ] Use settings to implement `AnalysisJob.run()` method in Django
- [ ] Write python script (as Django manage command so deps exist) that updates a Batch job definition with a new revision
- [ ] Run new script from `./scripts/infra` after successful terraform deploy

PR #121, opened 2017-03-03
Trigger pfb-analysis jobs via Django in AWS Batch

CloudNiner:
## Overview

This updates the Django application with the configuration necessary to trigger and control jobs in AWS Batch. Also adds instructions to the deployment README that detail how to setup a Batch compute environment along with the necessary job queues.

Remaining TODOs:
- [x] Fix jenkins build (Django tests fail to start because it's missing the new required ENV variables

## Demo

screen shot 2017-03-06 at 11 47 47
screen shot 2017-03-06 at 11 48 56

## Notes

- Added a custom IAM policy for the container instances so they can view and manage Batch jobs
- TODOs marked in source with additional future improvements to consider
- For now, development VMs are configured to target the staging Batch compute environment, rather than having multiple environments up. This would be pretty trivial to change.
- A particular Batch job definition + revision is linked to each deployment. This revision is bumped with each deploy.
- Fixed issue where job_name being longer than 50 chars keeps jobs from starting due to CloudWatch logs stream prefix limitation

## Testing

First, follow the new steps in the 'Provisioning the VM' section of the README to copy the example group vars file and add the appropriate Batch configuration. After making the changes you'll need to `vagrant provision`.

Then, SSH into the ECS container instance, and follow the commands in the screenshot to duplicate. This will work as long as another merge and deploy doesn't happen to develop. Otherwise, the same commands can be run in your development instance after doing `docker-compose build && ./scripts/server` to start the dev server. 

Connects #91, Connects #96


CloudNiner:
Additional Jenkins jobs will likely fail, since it requires manual modification of the Jenkinsfile via the 'replay' functionality to manually add the changes made to the Jenkinsfile in this PR. But the build is happy, see: http://urbanappsci.internal.azavea.com/job/azavea/job/pfb-network-connectivity/job/PR-121/11/

src/django/pfb_analysis/aws_batch.py
```diff
@@ -0,0 +1,36 @@
...
+    """ Get the latest revision of an AWS Batch job definition
+
+    Raises NoActiveJobDefinitionRevision if no current active revision for the
+        requested job definition
+
+    """
+    client = boto3.client('batch')
+    response = client.describe_job_definitions(jobDefinitionName=job_definition_name,
+                                               status='ACTIVE')
+    # TODO: Continue building the array if more than 100 results
```
KlaasH: A little worried about this falling through the cracks until it catches us by surprise.  What about either
```
    while(response.get('nextToken') is not None):
        response = client.describe_job_definitions(jobDefinitionName=job_definition_name,
                                                   status='ACTIVE', nextToken=response['nextToken'])
        job_definitions.extend(response.get('jobDefinitions', []))
```
or raising an exception if `nextToken` has a value, so we'll be alerted to the fact that we've hit the 100-definition mark?

src/django/pfb_analysis/models.py
```diff
@@ -75,19 +82,62 @@ class Status(object):
...
+        def create_environment(**kwargs):
+            return [{'name': k, 'value': v} for k, v in kwargs.iteritems()]
+
+        if self.batch_job_id is not None:
+            logger.warn('Attempt to re-run job: {}. Skipping.'.format(self.uuid))
+            return
+
+        client = boto3.client('batch')
+        job_definition = settings.PFB_AWS_BATCH_JOB_DEFINITION_NAME_REVISION
+        job_name = '{}:{}'.format(job_definition, str(self.uuid)[:8]).replace(':', '--')
```
KlaasH: ~~What about `str(self.uuid)[:8]` first, since that adds more information and would potentially cause more bother if it got totally cut off?~~

Just kidding, that would make them annoying to look for in CloudWatch.  What about `job_name = '{}--{}'.format(job_definition.replace(':', '--')[:40], str(self.uuid)[:8])` ?

src/django/pfb_analysis/aws_batch.py
CloudNiner: Ah. I forgot I added this TODO forever ago and then never circled back to it. Yes, this should definitely be done, some things changed while working on this PR that will make us hit this way faster.

src/django/pfb_analysis/models.py
CloudNiner: Hmm, that's probably better. We'd lose rivision info first, but that's saved separately on the job as a part of the definition name. So that's probably fine. Better than losing some or all of the job django uuid.

deployment/ansible/roles/pfb.env/tasks/main.yml
```diff
@@ -2,3 +2,12 @@
 
 - name: Set DEV_USER environment variable
   lineinfile: dest=/etc/environment regexp=^DEV_USER line="DEV_USER={{ dev_user }}"
+
+- name: Set PFB_AWS_BATCH_COMPUTE_ENVIRONMENT_ARN environment variable
```
CloudNiner: 👍

CloudNiner:
I'll make all of the suggested changes, I did add a commit for the read-only filesystem error, haven't pushed a new job definition that uses it yet though.

KlaasH:
Changes look good.  :+1: all around.

CloudNiner:
@hectcastro I'm going to merge this as is for now as its blocking more work. Happy to address any concerns in a follow up PR.

deployment/terraform/iam.tf
```diff
@@ -36,6 +36,31 @@ data "aws_iam_policy_document" "container_instance_ses_send_email" {
...
+      "batch:TerminateJob"
+    ]
+  }
+}
+
+#
+# Custom policies
+#
+resource "aws_iam_policy" "batch_manage_jobs" {
+  name = "BatchManageJobs"
```
hectcastro: We'll ultimately want this policy name to be scoped by environment, otherwise Terraform can run into issues when the resource for two managed infrastructures exists within the same AWS account.

src/django/pfb_analysis/models.py
```diff
@@ -75,19 +82,62 @@ class Status(object):
...
+            self.status = self.Status.IMPORTING
+            self.save()
+        except (botocore.exceptions.BotoCoreError, KeyError):
+            logger.exception('Error starting AnalysisJob {}'.format(self.uuid))
+
+    def batch_job_status(self):
+        """ Return current AWS Batch job status for this job
+
+        List of available statuses: http://docs.aws.amazon.com/batch/latest/userguide/jobs.html
+        TODO: Refactor to cache in db?
```
hectcastro: Meh, I think querying the API directly is fine until we see problems.

Issue #89, opened 2017-02-23
Analysis job should push/pull files to S3


CloudNiner:
- [ ] Input: neighborhood boundary
- [ ] Output: shapefiles
- [ ] Output: CSVs

All files should be within a single bucket, but keyed by job id or some other unique bucket key path

Issue #120, opened 2017-03-03
Put "Running the analysis" and "Importing other neighborhoods" in the pfb-analysis README


KlaasH:
It appears I neglected to remove the README from the pfb-analysis directory when I moved it into this repo, so I was going to say we should add one there but we can just transform the (outdated) one that's already there.  Right now the description, usage information, and examples for the analysis are spread across the main README, issue #10, and various PRs that introduce new options but might not affect either of those so don't add any documentation outside the PR discussion.

I think it would make sense to move all of that into a README inside `pfb-analysis` (and linked from the main README) that includes the existing introduction (or one like it), documentation of all the arguments and variables that can affect the run, and an examples section that includes the ones in issue #10.

Aside from being useful in itself, this would make it easier to see the evolution of behavior and options, since they'd be documented in a tracked file when the changes happen.

CloudNiner:
How difficult do you think this is? We have some additional PFB points left in the placeholder.

KlaasH:
I think 1 point.  A bunch of moving things around followed by a bit of looking for undocumented parameters and writing a sentence or two about each.

CloudNiner:
Agree. I'll update the placeholder and add to the sprint. This task is still a lower priority than the remaining pieces necessary to get an analysis running in AWS though.

PR #128, opened 2017-03-07
Use S3 for analysis files

KlaasH:
Adds S3 uploading and downloading.  Requires that `AWS_STORAGE_BUCKET_NAME` be set.  In development, `AWS_PROFILE` must also be set and the `.aws` directory must be mounted in the container.
The new bits:
- Given `PFB_SHPFILE_KEY`, downloads the boundary file from the given S3 key, unzips it, and sets the value of `PFB_SHPFILE`.  (If `PFB_SHPFILE` is also set, this will override it.)
- When finished, zips up the results and uploads to `{AWS_STORAGE_BUCKET_NAME}/results/{PFB_NEIGHBORHOOD_NAME}.zip`.  (`PFB_NEIGHBORHOOD_NAME` is now required by the export.)
- Adds an S3 clause to the "reuse big downloads" logic in `import_neighborhood.sh` and `import_jobs.sh`. They'll look first in the local `/data/` directory, then in `{AWS_STORAGE_BUCKET_NAME}/data/`, then do the download.

This doesn't do anything with OSM files so far.  Those aren't general like the state-level block and jobs information, so we'll need to add that to the `Neighborhood` or `AnalysisJob` model to make them uploadable/reusable.

Notes:
- The aws cli can be installed with pip or from a standalone bundle.  I started with the former, but after several issues with unstated dependencies and build errors, I went for the bundle instead.
- The first commit fixes the NB_INPUT_SRID bug in the same way that PR #119 does. Since I was changing the file, I wanted to be starting from the good version.  I didn't cherry-pick, but I tried to copy the diff so that it would merge/rebase cleanly.

**To Test**
Make a neighborhood.  In the django shell:
```
boundary_file = File(open('/data/gtown_westside.zip'))
gtown = Neighborhood.objects.create(name='gtown-westside', label='Germantown, PA', boundary_file=boundary_file, organization=org, state_abbrev='PA')
```
gtown.boundary_file.file.obj.key will now return the key of the file just uploaded to S3, which should be `neighborhood_boundaries/gtown-westside/gtown_westside.zip`

Then run the analysis.  In `/vagrant/` inside the VM:
```
pushd pfb-analysis && docker build -t pfb-analysis . && popd && \
docker run -it -e PFB_SHPFILE_KEY='neighborhood_boundaries/gtown-westside/gtown_westside.zip' \
    -e PFB_NEIGHBORHOOD_NAME='gtown-westside' -e PFB_STATE=pa -e PFB_STATE_FIPS=42 \
    -e NB_BOUNDARY_BUFFER=50 \
    -v /vagrant/data/:/data/ -v /home/vagrant/.aws:/root/.aws \
    -e AWS_PROFILE=pfb -e AWS_STORAGE_BUCKET_NAME="${DEV_USER}-pfb-storage-us-east-1" \
    pfb-analysis
```
That will run the analysis and upload the results.  (Note the `-it`, which will make it drop you to a shell rather than exiting when finished.  Remove that if not wanted.)

Connects #89.

### DIFF
--- a/pfb-analysis/import/import_neighborhood.sh
+++ b/pfb-analysis/import/import_neighborhood.sh
@@ -69,12 +69,18 @@ then
 
         # Get blocks for the state requested
         NB_BLOCK_FILENAME="tabblock2010_${NB_STATE_FIPS}_pophu"
-        if [[ -f "/data/${NB_BLOCK_FILENAME}.zip" ]]; then
-            unzip "/data/${NB_BLOCK_FILENAME}.zip" -d "${NB_TEMPDIR}"
+        if [ -f "/data/${NB_BLOCK_FILENAME}.zip" ]; then
+            echo "Using local census blocks file"
+            BLOCK_DOWNLOAD="/data/${NB_BLOCK_FILENAME}.zip"
+        elif [ "${AWS_STORAGE_BUCKET_NAME}" ] && aws s3 ls "s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_BLOCK_FILENAME}.zip"; then
+            echo "Using census blocks file from S3"
+            BLOCK_DOWNLOAD="${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.zip"
+            aws s3 cp "s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_BLOCK_FILENAME}.zip" "${BLOCK_DOWNLOAD}"
         else
-            wget -P "${NB_TEMPDIR}" "http://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/${NB_BLOCK_FILENAME}.zip"
-            unzip "${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.zip" -d "${NB_TEMPDIR}"
+            BLOCK_DOWNLOAD="${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.zip"
+            wget -O "${BLOCK_DOWNLOAD}" "http://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/${NB_BLOCK_FILENAME}.zip"
         fi
+        unzip "${BLOCK_DOWNLOAD}" -d "${NB_TEMPDIR}"
 
         # Import block shapefile
         import_and_transform_shapefile "${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.shp" neighborhood_census_blocks
--- a/pfb-analysis/scripts/entrypoint.sh
+++ b/pfb-analysis/scripts/entrypoint.sh
@@ -46,12 +46,13 @@ export NB_OUTPUT_SRID="$(./scripts/detect_utm_zone.py $PFB_SHPFILE)"
 
 ./scripts/import.sh $PFB_SHPFILE $PFB_STATE $PFB_STATE_FIPS $PFB_OSM_FILE
 ./scripts/run_connectivity.sh
-./scripts/export_connectivity.sh $NB_OUTPUT_DIR
 
 # print scores (TODO: replace with export script)
 psql -U "${NB_POSTGRESQL_USER}" -d "${NB_POSTGRESQL_DB}" <<EOF
 SELECT * FROM neighborhood_overall_scores;
 EOF
+
+./scripts/export_connectivity.sh $NB_OUTPUT_DIR $PFB_NEIGHBORHOOD_NAME
 
 # This will exit immediately when there's no pseudo-TTY but provide a shell if there is,
 # so it enables keeping a docker container alive after processing by running it with `-t`
--- a/pfb-analysis/scripts/entrypoint.sh
+++ b/pfb-analysis/scripts/entrypoint.sh
@@ -38,6 +38,20 @@ do
     ((counter++))
 done
 
+PFB_TEMPDIR=`mktemp -d`
+
+# If given a URL for the shapefile, dowload and unzip it. Overrides PFB_SHPFILE.
+if [ "${PFB_SHPFILE_URL}" ]
+then
+    echo "Downloading shapefile"
+    pushd "${PFB_TEMPDIR}"
+    wget "${PFB_SHPFILE_URL}" -O boundary.zip
+    unzip boundary.zip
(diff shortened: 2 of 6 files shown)


pfb-analysis/scripts/entrypoint.sh
```diff
@@ -38,6 +38,20 @@ do
     ((counter++))
 done
 
+PFB_TEMPDIR=`mktemp -d`
+
+# If given an S3 key for the shapefile, dowload and unzip it. Overrides PFB_SHPFILE.
+if [ "${PFB_SHPFILE_KEY}" ]
```
CloudNiner: It probably makes sense to make this more generic, such that we `wget` or `curl` a file provided to something like `PFB_SHPFILE_URL`. This way we could send in non-s3 download links. This also works with the current workflow because `job.boundary_file.url` returns a signed s3 download url valid for 24hrs, so we wouldn't need the aws CLI to retrieve this file.

I do think it makes sense that the download overwrites a locally provided file.

pfb-analysis/import/import_neighborhood.sh
```diff
@@ -45,9 +44,10 @@ NB_POSTGRESQL_PASSWORD - Default: gis
 function import_and_transform_shapefile() {
     IMPORT_FILE="${1}"
     IMPORT_TABLENAME="${2}"
+    IMPORT_SRID="${3:-4326}"
```
CloudNiner: I'm surprised these `IMPORT_SRID` changes are in here. Wasn't that change made awhile ago?

pfb-analysis/scripts/export_connectivity.sh
```diff
@@ -58,36 +59,56 @@ function ec_export_table_shp() {
...
         # Export neighborhood_ways as SHP
-        ec_export_table_shp "${NB_OUTPUT_DIR}" "neighborhood_ways"
+        ec_export_table_shp "${NB_OUTPUT_DIR}" "${NB_OUTPUT_FILE_PREFIX}" "neighborhood_ways"
 
         # Export neighborhood ways as GeoJSON
         # TODO: Add ogr2ogr and try again
 
         # Export neighborhood_connected_census_blocks as SHP
-        ec_export_table_shp "${NB_OUTPUT_DIR}" "neighborhood_connected_census_blocks"
+        ec_export_table_shp "${NB_OUTPUT_DIR}" "${NB_OUTPUT_FILE_PREFIX}" "neighborhood_connected_census_blocks"
```
CloudNiner: Let's disable the connected_census_blocks export for now, its a massive file and doesn't hold much value.

pfb-analysis/scripts/export_connectivity.sh
```diff
@@ -58,36 +59,56 @@ function ec_export_table_shp() {
...
+        if [ -v AWS_STORAGE_BUCKET_NAME ]
+        then
+          pushd "${NB_OUTPUT_DIR}"
+          sync  # Probably superfluous, but the s3 command said "file changed while reading" once
+
+          OUTPUT_FILENAME="${NB_OUTPUT_FILE_PREFIX}.zip"
+          zip "${OUTPUT_FILENAME}" "${NB_OUTPUT_FILE_PREFIX}_"*
+          sync
+
+          aws s3 cp "${OUTPUT_FILENAME}" "s3://${AWS_STORAGE_BUCKET_NAME}/results/"
```
CloudNiner: This is likely not going to work in the staging environment. There, we define a single storage bucket to hold all the analysis job results. We need to upload all files to a bucket scoped by the django job id (not currently provided to the analysis), otherwise files will end up clobbered. What if we uploaded our output files to:
- `s3://${AWS_STORAGE_BUCKET_NAME/results/${DJANGO_JOB_ID}/neighborhood_ways.zip`
- `s3://${AWS_STORAGE_BUCKET_NAME/results/${DJANGO_JOB_ID}/neighborhood_overall_scores.csv` 

That way, its trivial for the Django app to generate download links to the export files when a job is done.

pfb-analysis/import/import_neighborhood.sh
KlaasH: They're in PR #119.  I was thinking it would be better to base changes off that version than have to merge later, but actually the changes I made don't overlap, so I'll remove them from this branch.

pfb-analysis/scripts/entrypoint.sh
KlaasH: Yeah, wasn't sure which would be preferable, so I did the one then thought the expiring URLs might be a bother and switched to the other.  I'll switch back.

pfb-analysis/import/import_neighborhood.sh
CloudNiner: 👍

KlaasH:
Updated.  Command is now:
```
docker run -it \
    -e PFB_SHPFILE_URL='{url goes here}' \
    -e PFB_JOB_ID='{job id goes here}' -e PFB_STATE=pa -e PFB_STATE_FIPS=42 \
    -e NB_BOUNDARY_BUFFER=50 \
    -v /vagrant/data/:/data/ -v /home/vagrant/.aws:/root/.aws \
    -e AWS_PROFILE=pfb -e AWS_STORAGE_BUCKET_NAME="${DEV_USER}-pfb-storage-us-east-1" \
    pfb-analysis
```
Note the places to substitute the URL from the neighborhood boundary file and the job ID from the AnalysisJob (or make one up...).

Will comment on issue #120 when this is fully settled.

KlaasH:
Updated AnalysisJob.run().  Also caught an issue with the jobs downloads (they're using the state abbreviation in URLs and need it to be lower case, so I lowercased it in that script).

Making and running jobs like on PR #121 got me (eventually) to a successful one.  I got tripped up once by the fact that `settings.PFB_AWS_BATCH_JOB_DEFINITION_NAME_REVISION` gets set on load (so using the same `shell_plus` session got me a new job with the old definition) and then was surprised to see that it wrote to my bucket rather than the staging bucket, but of course that's correct because I made the job in my instance.

The commands, for convenient reference:
```
settings.PFB_AWS_BATCH_JOB_DEFINITION_NAME_REVISION
gtown = Neighborhood.objects.get(name='gtown-westside')
job.run()
job.batch_job_status()
```

Issue #127, opened 2017-03-06
Add Neighborhood CRUD API endpoints


flibbertigibbet:
Current analysis run process via the docker command as in #10 does not create any Django objects for the `Neighborhood` or the `AnalysisJob`.

Pre-requisite for #93. Or at least, need a way to create related `Neighborhood` in API before `AnalysisJob` can be created and run.

CloudNiner:
Right. I think when I wrote the issues a few weeks ago, I assumed that Neighborhood would be essentially just the boundary and would be tied one-to-one to an AnalysisJob. That is clearly no longer the case, so we'll need separate API endpoints for Neighborhood CRUD operations. I'm going to update the issue title to reflect that creating those endpoints is the issue here.

CloudNiner:
Creating these views in DRF is pretty trivial, any reason for this to be more than one point? I'd like to get this task into this sprint since its a prereq

Issue #93, opened 2017-02-23
Add API endpoint in Django to trigger an analysis job


CloudNiner:

CloudNiner:
Depends on #99 and #55

CloudNiner:
Since it should be relatively trivial, include endpoint for GET /api/analysis-job/

flibbertigibbet:
Also depends on #119

CloudNiner:
Actually, this should be fine, since all of the modification in #121 is contained within the `AnalysisJob.run()` method which is all this task should be concerned about calling. But, it may not hurt to wait since there may be other non-obvious concerns.

Since we'll be triggering jobs in AWS, #119 shouldn't apply.

PR #130, opened 2017-03-07
Feature/analysis job endpoint

flibbertigibbet:
### Add API endpoints for Neighborhood an AnalysisJob

Run `AnalysisJob` on creation, or if updated to reset its status to `CREATED`.

### Testing

 - Clear existing database with `./scripts/django-manage reset_db`
 - Run migrations: `./scripts/django-manage migrate`
 - Ensure AWS `pfb` configuration exists, as per README
 - Create a neighborhood using the form at http://localhost:9200/api/neighborhoods/
 - Create an analysis job using the form at http://localhost:9200/api/analysis_jobs/

Neighborhood and analysis job should be created successfully. Creation of analysis job should kick of job run, set batch job ID, and update status.

### Notes

After this is merged, will need to clear database, or at least delete existing neighborhoods and/or the default user, then re-run migrations. This is to support addition of auditing fields to neighborhoods.

Updated max upload size nginx allows to 6MB from the default 1MB, somewhat arbitrarily; not sure what the largest shapefile zip upload would be that we should support.

Connects #127 and connects #93.


src/django/pfb_analysis/models.py
```diff
@@ -43,8 +43,11 @@ def __repr__(self):
 
     def save(self, *args, **kwargs):
         """ Override to do validation checks before saving, which disallows blank state_abbrev """
+        if not self.label:
+            self.label = slugify(self.name)
         self.full_clean()
-        super(Neighborhood, self).save(*args, **kwargs)
+        neighborhood = super(Neighborhood, self).save(*args, **kwargs)
+        return neighborhood
```
CloudNiner: What's the reason for this. The Django model API spec indicates that `save()` shouldn't return anything.

src/django/pfb_analysis/views.py
```diff
@@ -0,0 +1,48 @@
...
+    filter_backends = (DjangoFilterBackend, OrderingFilter,
+                       OrgAutoFilterBackend, SelfUserAutoFilterBackend)
+    ordering_fields = ('created',)
+
+    def perform_create(self, serializer):
+        """ Start analysis jobs as soon as created """
+        instance = serializer.save()
+        instance.run()
+
+    def perform_update(self, serializer):
```
CloudNiner: It seems to me that we should make AnalysisJob immutable (to the user) once its been created. If a particular job needs to be re-run it can be submitted as a new job with the same parameters. Otherwise we potentially destroy information if the older run was successful.

src/django/users/migrations/0001_initial.py
```diff
@@ -18,7 +18,8 @@ def add_root_user(apps, schema_editor):
     PFBUser = apps.get_model('users', 'PFBUser')
     root_user = PFBUser(role='ADMIN',
                         password='root',
-                        email='systems+pfb@azavea.com')
+                        email='systems+pfb@azavea.com',
+                        uuid='bba2038e-863e-4c1e-a64c-c06b6000a61f')
```
CloudNiner: Shouldn't the UUID be set here automatically on save? Or does that not happen in migrations?

src/django/pfb_analysis/models.py
```diff
@@ -43,8 +43,11 @@ def __repr__(self):
 
     def save(self, *args, **kwargs):
         """ Override to do validation checks before saving, which disallows blank state_abbrev """
+        if not self.label:
+            self.label = slugify(self.name)
```
CloudNiner: `slugify` isn't imported anywhere. 

Since `self.name` is a slugfield, its probably best to make it read-only and just set it on every save by doing `self.name = slugify(self.label)`.

src/nginx/etc/nginx/nginx.conf
```diff
@@ -23,6 +23,8 @@ http {
 
     keepalive_timeout 65;
 
+    client_max_body_size 6m;
```
CloudNiner: I don't have further insight into how big this should be. A few megabytes as modified here is definitely sufficient for now.

src/django/pfb_analysis/models.py
flibbertigibbet: It resulted in error otherwise. Noticed the existing `Organization` `save` override returned a value, did so, and it fixed the error.

src/django/pfb_analysis/models.py
flibbertigibbet: Borrowed behavior from the existing `Organization` model, which has reversed `name` and `label` text/slug types. These should be consistent; looks like `Organization` is the one that's backwards.

src/django/pfb_analysis/views.py
flibbertigibbet: Wasn't sure what behavior we would want with that. Will disable update POST then.

src/django/users/migrations/0001_initial.py
flibbertigibbet: It is. This is to support forward migration for the `Neighborhood` model, which was missing auditing data, and needs to know the ID for the default user to reference for the forward migration.

src/django/pfb_analysis/views.py
flibbertigibbet: Something, perhaps in the serializer tweaks, has fixed this. Will update `Organziation` save as well.

src/django/pfb_analysis/models.py
flibbertigibbet: Something, perhaps in the serializer tweaks, has fixed this. Will update Organziation save as well.

flibbertigibbet:
Updated. Everything should be addressed.

Issue #34, opened 2016-11-22
Add test to validate bike network analysis correctness


CloudNiner:

CloudNiner:
Maybe an extra step in the analysis, toggled with an env var, that validates the results output from the `neighborhood_overall_scores` table against a saved CSV of the same table for a particular neighborhood (cambridge?). The test would only run when the flag is set, throwing an error if the results don't sufficiently match.

@spencerrecneps any other ideas for ways to validate algorithm correctness after changes? How do you test this?

CloudNiner:
This should be a bit simpler with the changes in #112 and #116 which allow local files to be used in the analysis, which helps to keep results from changing due to modifications of the input files that are automatically downloaded.

CloudNiner:
Checkout #119 for now to improve chances of a successful job running.

Probably worth testing against germantown or boulder data, since those run successfully as described in the above PR.

PR #133, opened 2017-03-07
Feature/verify outputs

flibbertigibbet:
Adds a docker container for running the `csvdiff` tool against a known analysis output and the latest run in the `data` directory (using the existing non-AWS analysis run process).

See updated README for instructions to test. Try editing output to change a value (other than the row ID) to test diff output.

Connects #34.


CloudNiner:
We should document the analysis command/arguments that generated the boulder file. Otherwise, we don't explicitly know if any given boulder analysis should match the verified output. Maybe an additional README in the new `verified_output/` dir?

flibbertigibbet:
It was generated with the existing documented analysis run process from the README, just copied the one exported CSV from the outputs to a new directory. Currently we have no means of verification and certainly none for validation other than to compare to existing output.

flibbertigibbet:
Do you mean to add something like #10? Hesitant to add a README referencing FILESHARE. Do we have some other docs on how those area inputs were generated?

CloudNiner:
Sort of like #10. It's probably worth copying the zipped neighborhood files for germantown, cambridge and boulder from FILESHARE to a public S3 bucket in the PFB AWS account, that way we can provide a public download link instead.

And yeah, I think even though the command used in this case matches the one in the existing documentation, it wouldn't necessarily have to. When a job can take a significant amount of time, you wouldn't want to run it only to have the output not match because you used the wrong arguments. So specifying the command/files/arguments that each output CSV was generated with seems beneficial to me.

src/verifier/compare_outputs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+if [ $# -eq 0 ]; then
+    echo "Missing argument for name of existing CSV file from verified_output to check against."
+    exit $1
```
KlaasH: A fuller usage message would be nice.  i.e. with `usage: verifier BASE_FILE [COMPARISON_FILE]` and a note that the arguments should be relative to `./verified_output` and `./data/output` respectively.

Also this `exit $1` should be `exit 1`, right?

src/verifier/compare_outputs.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "Comparing ${VERIFIED} to ${UNVERIFIED}"
+
+csvdiff -q id ${VERIFIED} ${UNVERIFIED}
+
+if [ $? -eq 0 ]; then
+    echo "Output matches!"
+else
+    echo "Output mismatch:"
+    csvdiff id --style pretty ${VERIFIED} ${UNVERIFIED}
```
KlaasH: A "quiet" mode would be nice, since the diff can be long.  Alternately it could exit with a nonzero status when the files differ so that the output could be sent to `/dev/null` and the result deduced from the exit status.  (actually exiting abnormally on mismatch seems useful regardless of whether we add a quiet flag)

flibbertigibbet: The execution of `csvdiff` above with the `-q` mode is quiet mode. If it fails (has nonzero exit), this execution of `csvdiff` outputs the actual difference.

If someone wants to do fancier things with `csvdiff`, they might as well just invoke it directly rather than using this convenience tool.

src/verifier/compare_outputs.sh
flibbertigibbet: Good catch on the `$1`. Full documentation of usage is in the `README` docs. Sure, could duplicate that here too.

src/verifier/compare_outputs.sh
KlaasH: So indeed.  What about just capturing the exit status of the `-q` run and exiting with that status from the `else` block?

flibbertigibbet: Why not provide the diff? For that behavior there's not really any reason to have a wrapper script for `csvdiff` at all.

flibbertigibbet: Another option is the summary style, with add/delete change row counts:
image

KlaasH:
Relevant to documenting the command and sharing the input files, I just made a WIP PR toward the broader documentation issue (PR #136), which currently mentions the fileshare but I think should be changed to reference this hypothetical S3 bucket.  The task of making said bucket could be done for this PR or deferred and attached to that one.

flibbertigibbet:
Okay, I was about to set up the bucket. I'll go ahead with that.

src/verifier/compare_outputs.sh
KlaasH: I meant still provide the diff but also pass through the exit status.  Maybe in the actual use cases for this thing that wouldn't matter, but if it gets integrated into any kind of scripted or automated process having it exit abnormally on mismatch would be useful, and it seems like the sort of thing one would expect a "verifier" to do.

Summary table might be a better default, actually, if we end up using this on big CSVs (e.g. database table dumps rather than summary statistics), but there are probably pros and cons either way, so the diff is fine.

flibbertigibbet:
Just pushed a commit that should address everything apart from the S3 input bucket and its extra documentation.

flibbertigibbet:
The input shapefiles and OSM extracts are now publicly accessible on S3, [here](https://s3.amazonaws.com/test-pfb-inputs/). Included are Boulder, Lower Manhattan, West Germantown, Philadelphia, and Cambridge.

Currently the only verified output file present is for Boulder, and was generated using the parameters specified in #10.

@KlaasH would you like me to add documentation reflecting the above to this PR, or shall that go in #136?

KlaasH:
Cool.  Would you mind adding a line or a little table in the README section providing for each file in `verified_output` (i.e. just boulder.csv for now but maybe more in future...) what `NB_BOUNDARY_BUFFER` it was run with and links to the shapefile and the OSM file?

PR #134, opened 2017-03-08
Feature/add destination types

spencerrecneps:
Creates new tables to facilitate scoring for all the access categories (with the exception of transit, which is still in development).
I added psql calls to the run_connectivity.sh script so these should be running automatically with everything else.


pfb-analysis/scripts/run_connectivity.sh
```diff
@@ -12,6 +12,10 @@ NB_POSTGRESQL_USER="${NB_POSTGRESQL_USER:-gis}"
 NB_POSTGRESQL_PASSWORD="${NB_POSTGRESQL_PASSWORD:-gis}"
 NB_OUTPUT_SRID="${NB_OUTPUT_SRID:-4326}"
 NB_BOUNDARY_BUFFER="${NB_BOUNDARY_BUFFER:-0}"
+TOLERANCE_COMM_CTR="${TOLERANCE_COMM_CTR:-150}"
```
CloudNiner: Can you add a comment here that states what units these values should be in?

spencerrecneps:
Just committed. Is this what you're after? I assume there's no need to create a new PR but let me know if that's what I should do.

PR #136, opened 2017-03-08
Move and modify 'Running the analysis' documentation

KlaasH:
Connects #120.

The details of running the analysis change frequently and the documentation is scattered.  This adds a script, `scripts/run-local-analysis` to encapsulate the defaults so we don't have to include so much boilerplate in all our examples that quickly goes out of date.

It also moves the "Running the analysis" section into its own README and changes it to use Boulder (mainly because Cambridge requires overriding the input SRID).

This should replace Issue #10 and make the command examples in various other issues (e.g. #128) no longer necessary.  It should also make it easier to keep the behavior and parameters of the analysis well documented.

~~There are multiple TODOs here, but I wanted to put something up to facilitate the conversation in connection with the discussion on PR #133 ([here](https://github.com/azavea/pfb-network-connectivity/pull/133#issuecomment-285088349)) about putting the example source files in a publicly-accessible place.~~

### DIFF
--- a/pfb-analysis/scripts/entrypoint.sh
+++ b/pfb-analysis/scripts/entrypoint.sh
@@ -52,6 +52,18 @@ then
     popd
 fi
 
+# If given a URL for the OSM file, dowload and unzip it. Overrides PFB_OSM_FILE.
+if [ "${PFB_OSM_FILE_URL}" ]
+then
+    echo "Downloading OSM file"
+    pushd "${PFB_TEMPDIR}"
+    wget "${PFB_OSM_FILE_URL}" -O neighborhood_osm.zip
+    unzip neighborhood_osm.zip
+    PFB_OSM_FILE="${PFB_TEMPDIR}"/$(ls *.osm)  # Assumes there's exactly one .osm file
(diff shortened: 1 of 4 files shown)


pfb-analysis/scripts/entrypoint.sh
```diff
@@ -52,6 +52,18 @@ then
     popd
 fi
 
+# If given a URL for the OSM file, dowload and unzip it. Overrides PFB_OSM_FILE.
+if [ "${PFB_SHPFILE_URL}" ]
```
flibbertigibbet: Comment says this is to download OSM file, but parameter checked is for shapefile? Below it references an OSM URL parameter, but downloads it as a zip. Only the shapefile is zipped. There are two things to download, an OSM file and a zipped shapefile. The shapefile is the one needing extraction.

KlaasH: Oops, copy/paste/modify error.  Just pushed a fix.

I put zipped copies of the OSM files in the test files bucket, actually, because they're so big (220MB for lower Manhattan) and they deflate by about 90%.

Issue #129, opened 2017-03-07
Replace AWS Batch Managed compute environment with unmanaged environment


CloudNiner:
Goal is to have a drop in replacement, configured via Terraform as ECS stack. Cloud init and docker customizations can happen separately.

PR #137, opened 2017-03-08
Switch to Batch Unmanaged Compute Environment

CloudNiner:
## Overview

Our analysis jobs require large amounts of scratch space for the postgresql data dir. This required switching to unmanaged CE's so that we can mount the ephemeral storage space provided by the `i3` EC2 instance class. 

Remaining TODOs:
- [x] Update deployment/README#aws-batch section with info on setting up unmanaged compute environment
- [x] Figure out how to autoscale this ECS cluster. We should be able to scale on CPU...this may need to be left to a separate task.

## Demo

screen shot 2017-03-08 at 15 25 08

## Notes

Left some comments in the diff for things I'm unsure of how to handle.

The Jenkins build is currently disabled so that merges to develop do not delete the extra ECS stack that was deployed via this branch. Re-enable Jenkins before merging.

#118 is no longer necessary as we're mounting an external docker volume

## Testing

Pull down this branch, start your local Django instance, and then run the following commands:
```

# Option 1, if gtown doesn't exist in your local db already
boundary_file = File(open('/data/gtown_westside.zip'))
gtown = Neighborhood.objects.create(name='gtown-westside', label='Germantown, PA', boundary_file=boundary_file, organization=org, state_abbrev='PA', created_by=user, modified_by=user)
# OR option 2 if it does
gtown = Neighborhood.objects.first()

job.run()
```
Watch the AWS Batch console to ensure the job you started succeeds. The job should run definition `staging-pfb-analysis-run-job:42`.

Connects #118, Connects #129

### DIFF
--- a/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
+++ b/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
@@ -5,7 +5,7 @@
     "containerProperties": {
         "image": "{image}",
         "vcpus": 2,
-        "memory": 4096,
+        "memory": 12288,
         "command": [
             "/pfb/scripts/entrypoint.sh"
         ],
@@ -16,7 +16,19 @@
                 "sourcePath": "/media/nvme0n1/pgdata"
             }
         }],
-        "environment": [],
+        "environment": [{
+            "name": "PFB_MAINTENANCE_WORK_MEM",
+            "value": "2048MB"
+        }, {
+            "name": "PFB_SHARED_BUFFERS",
+            "value": "3072MB"
+        }, {
+            "name": "PFB_TEMP_FILE_LIMIT",
+            "value": "104857600"
+        }, {
+            "name": "PFB_WORK_MEM",
+            "value": "1536MB"
+        }],
         "mountPoints": [{
             "sourceVolume": "pgdata",
             "containerPath": "/pgdata",
--- a/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
+++ b/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
@@ -4,15 +4,24 @@
     "parameters": {},
     "containerProperties": {
         "image": "{image}",
-        "vcpus": 4,
+        "vcpus": 2,
         "memory": 4096,
         "command": [
             "/pfb/scripts/entrypoint.sh"
         ],
         "jobRoleArn": "",
-        "volumes": [],
+        "volumes": [{
+            "name": "pgdata",
+            "host": {
+                "sourcePath": "/media/nvme0n1/pgdata"
+            }
+        }],
         "environment": [],
-        "mountPoints": [],
+        "mountPoints": [{
+            "sourceVolume": "pgdata",
+            "containerPath": "/pgdata",
+            "readOnly": false
+        }],
         "readonlyRootFilesystem": false,
         "privileged": true,
         "ulimits": [],
--- a/deployment/terraform/batch-container-service.tf
+++ b/deployment/terraform/batch-container-service.tf
@@ -1,0 +1,80 @@
+#
+# Security group resources
+#
+resource "aws_security_group" "batch_container_instance" {
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgBatchContainerInstance"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+#
+# Autoscaling Resources
+#
+data "template_file" "batch_container_instance_cloud_config" {
+  template = "${file("cloud-config/ecs-batch-user-data.yml")}"
+
+  vars {
+    ecs_cluster_name = "${var.batch_ecs_cluster_name}"
+    environment      = "${var.environment}"
+  }
+}
+
+resource "aws_launch_configuration" "batch_container_instance" {
+  lifecycle {
+    create_before_destroy = true
+  }
+
+  ebs_optimized = true
+  iam_instance_profile = "${aws_iam_instance_profile.container_instance.name}"
+  image_id             = "${var.ecs_instance_ami_id}"
+  instance_type        = "${var.batch_container_instance_type}"
+  key_name             = "${var.aws_key_name}"
+  security_groups      = ["${aws_security_group.container_instance.id}"]
+
+  user_data = "${data.template_file.batch_container_instance_cloud_config.rendered}"
+}
+
+resource "aws_autoscaling_group" "batch_container_instance" {
+  name = "asg${var.environment}BatchContainerInstance"
+
+  launch_configuration      = "${aws_launch_configuration.batch_container_instance.name}"
+  health_check_type         = "EC2"
+  desired_capacity          = "${var.batch_container_instance_asg_desired_capacity}"
+  min_size                  = "${var.batch_container_instance_asg_min_size}"
+  max_size                  = "${var.batch_container_instance_asg_max_size}"
+
+  enabled_metrics = [
+    "GroupMinSize",
+    "GroupMaxSize",
+    "GroupDesiredCapacity",
+    "GroupInServiceInstances",
+    "GroupPendingInstances",
+    "GroupStandbyInstances",
+    "GroupTerminatingInstances",
+    "GroupTotalInstances",
+  ]
+
+  vpc_zone_identifier = ["${module.vpc.private_subnet_ids}"]
+
+  tag {
+    key                 = "Name"
+    value               = "BatchContainerInstance"
+    propagate_at_launch = true
+  }
+
+  tag {
+    key                 = "Project"
+    value               = "${var.project}"
+    propagate_at_launch = true
+  }
+
+  tag {
+    key                 = "Environment"
+    value               = "${var.environment}"
+    propagate_at_launch = true
+  }
+}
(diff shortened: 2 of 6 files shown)


deployment/terraform/batch-container-service.tf
```diff
@@ -0,0 +1,95 @@
...
+    key                 = "Environment"
+    value               = "${var.environment}"
+    propagate_at_launch = true
+  }
+}
+
+#
+# ECS Resources
+#
+resource "aws_ecs_cluster" "batch_container_instance" {
```
CloudNiner: I _think_ we need to define this resource here so that Terraform associates the other resources with this ECS cluster. However, we don't want Terraform to ever do anything to this resource, e.g. it should ignore changes, not delete if theres a diff, not create if it doesn't exist.

Is there another way to handle this? 

The `lifecycle` [meta-parameter](https://www.terraform.io/docs/configuration/resources.html) doesn't seem to quite have what I need.

deployment/terraform/batch-container-service.tf
```diff
@@ -0,0 +1,95 @@
...
+  iam_instance_profile = "${aws_iam_instance_profile.container_instance.name}"
+  image_id             = "${var.ecs_instance_ami_id}"
+  instance_type        = "${var.batch_container_instance_type}"
+  key_name             = "${var.aws_key_name}"
+  # TODO: Use a separate security group for Batch cluster instances?
+  security_groups      = ["${aws_security_group.container_instance.id}"]
+
+  user_data = "${data.template_file.batch_container_instance_cloud_config.rendered}"
+
+  # Not necessary for NVMe SSD instance stores
```
CloudNiner: I can delete this before merging, I just wanted to make it explicit here that this isn't necessary, since AWS automatically maps NVMe SSD ephemeral devices.

deployment/terraform/batch-container-service.tf
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+resource "aws_launch_configuration" "batch_container_instance" {
+  lifecycle {
+    create_before_destroy = true
+  }
+
+  ebs_optimized = true
+  # TODO: Use a separate container instance profile for Batch cluster?
```
CloudNiner: This TODO is probably not worth doing ATM.

deployment/terraform/batch-container-service.tf
```diff
@@ -0,0 +1,95 @@
...
+  iam_instance_profile = "${aws_iam_instance_profile.container_instance.name}"
+  image_id             = "${var.ecs_instance_ami_id}"
+  instance_type        = "${var.batch_container_instance_type}"
+  key_name             = "${var.aws_key_name}"
+  # TODO: Use a separate security group for Batch cluster instances?
```
CloudNiner: This TODO probably is worth doing, but I wanted to get everything working first. If we do decide to just reuse the existing SG, then the SG resource above can be deleted.

CloudNiner:
@KlaasH you might want to take a look at this too

deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
```diff
@@ -4,15 +4,36 @@
...
             "/pfb/scripts/entrypoint.sh"
         ],
         "jobRoleArn": "",
-        "volumes": [],
-        "environment": [],
-        "mountPoints": [],
+        "volumes": [{
+            "name": "pgdata",
+            "host": {
+                "sourcePath": "/media/nvme0n1/pgdata"
```
hectcastro: Did this path get created automatically when you booted the `i3` (RE: `nvme0n1`)?

deployment/terraform/batch-container-service.tf
```diff
@@ -0,0 +1,95 @@
...
+    environment      = "${var.environment}"
+  }
+}
+
+resource "aws_launch_configuration" "batch_container_instance" {
+  lifecycle {
+    create_before_destroy = true
+  }
+
+  ebs_optimized = true
```
hectcastro: I think that this will cause a failure if we ever try to use an instance type that doesn't support EBS optimizations (`t2.*`, etc.).

deployment/terraform/batch-container-service.tf
hectcastro: I created https://github.com/azavea/pfb-network-connectivity/issues/147 so you can drop the TODO.

deployment/terraform/batch-container-service.tf
hectcastro: I do think we'll want different ones, especially if one is gonna be for batch and the other for web (this one will likely contain less rules).

deployment/terraform/batch-container-service.tf
hectcastro: Interesting. This sometimes happens when there is only one ephemeral storage device available. Makes sense to drop if the default path doesn't cause any problems.

deployment/terraform/batch-container-service.tf
hectcastro: Do you know why we wouldn't want Terraform to just create it and then associate it with Batch later?

If there is a reason why we can't, I think we can try to use the `aws_ecs_cluster` data source to get relevant attributes (provided that the ones we need are supported): https://www.terraform.io/docs/providers/aws/d/ecs_cluster.html

deployment/terraform/cloud-config/ecs-batch-user-data.yml
```diff
@@ -0,0 +1,78 @@
+#cloud-config
+
+packages:
+    - awslogs
+
+fs_setup:
```
hectcastro: Above when you mention that the block device mapping isn't needed, is also true that we can omit this stuff too?

deployment/terraform/batch-container-service.tf
CloudNiner: When you create an 'unmanaged' batch compute environment, the ECS cluster associated with that gets created automatically, with the name you specify, plus some unique id. For example if I create an unmanaged compute environment in the AWS console with the name `staging-pfb-analysis-unmanaged-compute-environment`, AWS automatically creates the ECS cluster with the name `staging-pfb-analysis-unmanaged-compute-environment_Batch_575a710c-fa1c-3e0c-8182-13d0509059cc`. 

I don't see a way to associate an existing ECS cluster with an unmanaged compute environment or Batch job queue.

CloudNiner: For the data source. That would definitely work better, assuming the aws_ecs_cluster data source can be linked properly to the aws_launch_configuration. I don't see how that happens in the current implementation...maybe it happens in some auto-magic way, or I'm just missing the reference that links them.

deployment/terraform/batch-container-service.tf
CloudNiner: Ok. I just added #148 to address this separately as well.

deployment/terraform/cloud-config/ecs-batch-user-data.yml
CloudNiner: From http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html:

[NVMe instance store volumes] These volumes are mapped automatically as /dev/nvme[0-7]n1. You do not need to specify the NVMe volumes supported by an instance type in a block device mapping.

All this does is make the block device visible to the OS, e.g. via `fdisk`. You still need to mount it yourself, which is the purpose of the additions in the cloud-config file.

deployment/terraform/batch-container-service.tf
hectcastro: Ah, the connection is made via the `cloud-config` YAML: https://github.com/azavea/pfb-network-connectivity/blob/feature/awf/aws-batch-unmanaged/deployment/terraform/cloud-config/ecs-batch-user-data.yml#L21-L25

You just need the name passed into that template as `ecs_cluster_name`.

deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
CloudNiner: See other comment about how nvme SSDs get mounted in AWS.

`/dev/nvme0n1` is automatically exposed to the OS, then `/media/nvme0n1` is the mount point for `/dev/nvme0n1`, set via cloud-config. Docker automatically creates any subdirectories that are part of the volume mount point that don't exist.

deployment/terraform/cloud-config/ecs-batch-user-data.yml
hectcastro: That makes sense. There are some instances that automatically mount and format the devices. That's partially why `- [ /dev/nvme0n1, null ]` exists in the YAML (undo the default setups and redo them with custom options).

deployment/terraform/batch-container-service.tf
CloudNiner: Huzzah! In that case I'll remove the `aws_ccs_cluster` resource and replace with data source.

hectcastro: After expanding the view of the rest of your branch, you probably don't even need the data source. I think you just need to pass the name of the cluster in and that's it.

CloudNiner:
Updated, rebased and ready for one more look. Was able to drop the batch_ecs_cluster reference to just a var, no data_source necessary.

Was able to cleanup the Terraform plan by executing `terraform state rm aws_ecs_cluster.batch_container_instance` before running another round of `./scripts/infra plan && ./scripts/infra apply`. I'm going to re-enable Jenkins since the batch ECS cluster itself now lives outside Terraform.

Added #151 to address autoscaling, which I'm not worried about ATM since we're just focusing on the static 300 city run, which we can manually autoscale for.

Issue #99, opened 2017-02-23
Disassociate Neighborhood model from Users/Orgs in Django


CloudNiner:

CloudNiner:
Move to one-to-many in pfb_analysis app

PR #146, opened 2017-03-09
Remove 'area' org attribute

KlaasH:
Connects #102 (the back-end corollary is #99).

Removes the "area" association with organizations and users, which acted as a cross-org permissions category.

There was also some mention of "area" in what's currently called the "boundary-uploads" code.  In that context it does actually mean Neighborhood, so I changed it to that.  The structure of that section is somewhat different from what we'll need it to be, but I felt like there's enough overlap that it makes sense to adapt it into the AnalysisJob views rather than rip it out and build something entirely new.

**To test**
Start and log into the management site.  Under Org Management, the "Area" filter on the list view and "Area" field in the add view should be gone.  Under Data, they should now be called "Neighborhood".


CloudNiner:
Note that if you merge this before RRP it will not end up on staging, since builds are still disabled due to #137

PR #150, opened 2017-03-09
Feature/connectivity calculations

spencerrecneps:
This PR creates access scores for nearly all of the categories we're using for the evaluation. I have one or two other categories to work on still but in the interest of letting you integrate what I have as quickly as possible I'll take care of the last few items in a separate PR.


spencerrecneps:
Good. I'll finish up the changes and let you know when it's time to merge.

CloudNiner:
Is this ready for review and merge? Or can we just close and supersede with #153?

spencerrecneps:
We can close and superseded with #153 if you're OK with that. I've made SRID changes to these files there so this should now be considered obsolete.

PR #152, opened 2017-03-10
Replace NB_BOUNDARY_BUFFER with NB_MAX_TRIP_DISTANCE

KlaasH:
Connects #81.

Replaces the NB_BOUNDARY_BUFFER variable used in the analysis with
NB_MAX_TRIP_DISTANCE, since that's what it's actually doing.

In import_neighborhood.sh, where the buffer was used to truncate the
census blocks table to only the ones we care about, it now uses half
of NB_MAX_TRIP_DISTANCE, since it's not possible for a trip between
two points inside the boundary to go more than NB_MAX_TRIP_DISTANCE/2
outside of it.

NB_BOUNDARY_BUFFER is still accepted as an override to the census block
filter, in case that should prove desirable.

Also includes a few fixes and tweaks to `run-local-analysis`.


pfb-analysis/import/import_neighborhood.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 NB_INPUT_SRID="${NB_INPUT_SRID:-4326}"
-NB_OUTPUT_SRID="${NB_OUTPUT_SRID:-4326}"
-NB_BOUNDARY_BUFFER="${NB_BOUNDARY_BUFFER:-0}"
+NB_OUTPUT_SRID="${NB_OUTPUT_SRID:-2163}"
```
CloudNiner: I'm not sure we want this to default to MA state plane, as that's an odd choice. Should probably still default to 4326 as its a generic projection that would at least (sort of) work for most locations.

KlaasH: MA state plane is 2249.  2163 is "US National Atlas Equal Area".  The problem with 4326 as an output SRID is it's lat/lon, so the default max_trip_distance of 3300, intended to be about 2 miles, becomes the whole world.  So I went looking for a projection in meters that could make sense as a baseline default.

Alternatives include:
- Different default projection
- Leave the default as 4326 and change the default max trip distance to a value in degrees (not sure whether the pgrouting algorithms would handle that properly, but a very quick google made me think they might)
- Set the output SRID to 4326 and the max trip distance to something in meters* and don't worry about it

I meant to ask: is there a reason you went with 3600 as a default when switching the default commands from feet to meters?  I had been using 3300, since it's just over 2 miles.

CloudNiner: Whoops, I misspoke. 2163 seems to make the most sense here since we can then leave the default max trip distance in meters and the resulting analysis from the defaults would make sense.

 I think I probably chose 3600 since its an even number very close to 11000 when multiplied by 3, which is an approximation but not exact for ft -> m. So I'm fine with doing 3300. 

So summary of changes:
- Update all defaults/comments in the analysis to use NB_OUTPUT_SRID=2163
- Do the same for NB_MAX_TRIP_DISTANCE=3300

CloudNiner:
Consider waiting to merge until #153 is merged, there will likely be additional changes and conflicts necessary based on the added scripts in that PR.

CloudNiner:
This is good to rebase and merge.

KlaasH:
Closing, replaced by #167.

CloudNiner:
Looks like maybe you missed when you clicked the "close" button? 

Actually closing 😄

Issue #154, opened 2017-03-11
Remove TDG Tools


spencerrecneps:
Need to remove TDG Tools from the code.

spencerrecneps:
TDG Tools calls have been removed from the SQL scripts. Now we just need to eliminate from the VM setup.

PR #155, opened 2017-03-11
Feature/build network

spencerrecneps:
This removes the need for the TDG Tools extension by pulling the network building into a script. I didn't want to touch the VM setup, but there's no longer a need to pull TDG Tools from my repo.

This partially resolves https://github.com/azavea/pfb-network-connectivity/issues/154

### DIFF
--- a/pfb-analysis/Dockerfile
+++ b/pfb-analysis/Dockerfile
@@ -44,8 +44,6 @@ RUN set -xe && \
         (cd osm2pgsql && mkdir build && cd build && cmake ../ && make install) && \
       git clone --branch $GIT_BRANCH_QUANTILE https://github.com/tvondra/quantile.git && \
         (cd quantile && make install) && \
-      git clone --branch $GIT_BRANCH_TDGTOOLS https://github.com/spencerrecneps/TDG-Tools.git && \
-        (cd TDG-Tools/TDG\ SQL\ Tools && make clean && make install) && \
     cd /tmp/ && rm -rf /tmp/build/ /var/lib/apt/lists/* && \
     apt-get purge -y --auto-remove ${BUILD_DEPS}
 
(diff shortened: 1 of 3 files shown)


CloudNiner:
What is the reasoning for this change? After browsing the source for those TDG functions, it looks mostly the same.

CloudNiner:
The TDG Tools can be removed from the VM setup by deleting lines 47-48 in the `pfb-analysis/Dockerfile`:

https://github.com/azavea/pfb-network-connectivity/blob/develop/pfb-analysis/Dockerfile#L47

spencerrecneps:
You're correct that it's mostly the same. I basically took the TDG Tools code and copied it wholesale into the scripts here.

The primary purpose was to remove dependency on TDG Tools, which was never really intended to be part of the PFB source. It will be better to keep all functionality contained within the PFB repo.

CloudNiner:
Gotcha. The major disadvantage of this is that you're now copying code if you want the same functionality in TDGTools as the PFB repo. If its an issue of open sourcing, then adding a LICENSE to TDGTools (if one doesn't exist) should be sufficient. 

We could modify the Dockerfile build that pulls the TDGTools SQL code to always specify a particular commit. That way you could leave all the code in one place and continue developing on whatever branch you want, without the danger of accidentally deploying changes before they're ready.

Either way, if you're willing to handle the maintenance burden of two duplicate codebases, and/or TDGTools won't be used for anything going forward (i.e. it's only purpose was to eventually be merged into PFB) then we can merge this. I would just urge caution before taking on that burden, depending on your future plans, as it could very quickly become unsustainable.

spencerrecneps:
The functionality here is simply building a routable road network. It's a small piece of the TDG Tools suite, which I'm not sure we as a firm want to publicize. (It's obviously discoverable right now but we'd prefer not to trumpet its existence to the world.)

Apart from any bugs in the way it builds the network, I don't foresee making any changes. It's a very specific piece of functionality that I wouldn't be planning to add features to even if I made changes in the original codebase.

PR #153, opened 2017-03-11
Feature/census block performance improvements

spencerrecneps:
These changes result in orders-of-magnitude better performance on connected_census_blocks.sql. My testing went from over 8 minutes with the old code down to 30 seconds with the new code. This change also resolves the issue identified in https://github.com/azavea/pfb-network-connectivity/issues/149.

PR #158, opened 2017-03-14
Feature/census block performance improvements fixed

CloudNiner:
Due to some rebasing issues, likely caused by doing a `git pull` and `git push` after running `git rebase develop` rather than just doing `git push -f`. Some stackoverflow hints suggest that this is difficult to fix properly on the branch, so I opted to just quickly cherry-pick all the relevant commits to a new branch which will work just fine.

@spencerrecneps please review this diff and make sure this looks correct as compared to the original pull request (before the rebase attempt). If it looks good to you I'll clean up the commit log a bit and then merge.

Supersedes #153


spencerrecneps:
Odd. git diff is only showing differences related to https://github.com/azavea/pfb-network-connectivity/pull/155, as if the rebase went off without a hitch.

In any case, it looks good to me.

CloudNiner:
Great, that makes sense, since #155 was merged and was in develop, which the previous version of this PR was not. Thanks for taking a look.

PR #159, opened 2017-03-14
Fix distance given in feet, switch to meters.

spencerrecneps:
Just discovered a reference to feet in a couple of distance calculations. This PR switches to meters.

### DIFF
--- a/pfb-analysis/connectivity/census_blocks.sql
+++ b/pfb-analysis/connectivity/census_blocks.sql
@@ -4,8 +4,10 @@
 -- code to be run on table that has
 -- been imported directly from US Census
 -- blkpophu file
--- :nb_output_srid psql var must be set before running this script,
---      e.g. psql -v nb_output_srid=2163 -f census_blocks.sql
+-- :nb_output_srid psql, :block_road_buffer, and :block_road_min_length vars
+-- must be set before running this script,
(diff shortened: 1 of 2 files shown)


pfb-analysis/connectivity/census_blocks.sql
```diff
@@ -96,7 +96,7 @@ ALTER TABLE neighborhood_census_blocks DROP COLUMN IF EXISTS tmp_geom_buffer;
 ALTER TABLE neighborhood_census_blocks ADD COLUMN tmp_geom_buffer geometry(multipolygon, :nb_output_srid);
 
 UPDATE  neighborhood_census_blocks
-SET     tmp_geom_buffer = ST_Multi(ST_Buffer(geom,50));
+SET     tmp_geom_buffer = ST_Multi(ST_Buffer(geom,15));
```
CloudNiner: Since we technically allow any SRID to be run, not just ones in meters, these magic values should probably parameterized and passed to this script in the same way as the values recently modified in other scripts, using a helpful variable name added to run_connectivity.sh.

spencerrecneps:
Good point. I'll update.

spencerrecneps:
Also noticed we were missing a reference to nb_output_srid so I added that in.

PR #162, opened 2017-03-14
Manually mount i3 instance family ephemeral storage

CloudNiner:
## Overview

The fs_setup cloud-config directive was not properly mounting the i3 instance family ephemeral storage, discovered when our docker volume mount point mounted directly on the 8Gb root volume and a job ran out of disk space.

See https://bugs.launchpad.net/cloud-init/+bug/1672833 for a detailed breakdown of why this solution is necessary

## Demo

Mounted disks after boot:

screen shot 2017-03-14 at 15 12 15

From /var/log/cloud-init-output.log:

screen shot 2017-03-14 at 15 16 45

## Testing

Nothing really to test, with this configuration applied via `./scripts/infra plan && ./scripts/infra apply` the ephemeral device is properly mounted at the expected location as evidenced in the screenshots above.

Connects #163

### DIFF
--- a/deployment/terraform/cloud-config/ecs-batch-user-data.yml
+++ b/deployment/terraform/cloud-config/ecs-batch-user-data.yml
@@ -3,19 +3,14 @@
 packages:
     - awslogs
 
-fs_setup:
-  - label: nvme0n1
-    filesystem: ext4
-    extra_opts: ["-E", "nodiscard"]
-    device: /dev/nvme0n1
-    partition: auto
-
-mounts:
-  - [ /dev/nvme0n1, null ]
-  - [ /dev/nvme0n1, "/media/nvme0n1", "ext4", "defaults,nobootwait,discard", "0", "2" ]
-
 runcmd:
   - curl -o /etc/papertrail-bundle.pem https://papertrailapp.com/tools/papertrail-bundle.pem
+  # Manually mount i3 family ephemeral storage, see:
+  # https://bugs.launchpad.net/cloud-init/+bug/1672833
+  - mkfs.ext4 -E nodiscard /dev/nvme0n1
+  - mkdir -p /media/nvme0n1
+  - echo -e "/dev/nvme0n1\t/media/nvme0n1\text4\tdefaults,nofail,discard\t0\t2" >> /etc/fstab
+  - mount -a
 
 write_files:
   - path: /etc/ecs/ecs.config


deployment/terraform/cloud-config/ecs-batch-user-data.yml
```diff
@@ -3,19 +3,22 @@
...
+  # Manually mount i3 family ephemeral storage, see:
+  # https://bugs.launchpad.net/cloud-init/+bug/1672833
+  # o - Create empty dos partition table
+  # n - Add new partition
+  # p - Primary partition (In add new prompt)
+  # 1 - Partition number (In add new prompt)
+  # empty newline - First sector, accepts default: 1
+  # empty newline - Last sector, accepts default: end
+  # w - Write changes
+  - echo -e "o\nn\np\n1\n\n\nw" | fdisk /dev/nvme0n1
```
hectcastro: Is this step something that actually needs to happen? My understanding is that the block device presented should be ready for `mkfs`.

deployment/terraform/cloud-config/ecs-batch-user-data.yml
```diff
@@ -3,19 +3,22 @@
...
+  - mkfs.ext4 /dev/nvme0n1p1
```
hectcastro: Based on (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html), I think that we may want this to be:

```yaml
- mkfs.ext4 -E nodiscard /dev/nvme0n1p1
```

deployment/terraform/cloud-config/ecs-batch-user-data.yml
```diff
@@ -3,19 +3,22 @@
...
+  - mkdir -p /media/nvme0n1p1
+  - echo -e "/dev/nvme0n1p1\t/media/nvme0n1p1\text4\tdefaults,discard\t0\t0" >> /etc/fstab
```
hectcastro: For similar reasons:

```yaml
- echo -e "/dev/nvme0n1p1\t/media/nvme0n1p1\text4\tdefaults,nofail,discard\t0\t2" >> /etc/fstab
```

- `nofail`: Bomb boot sequence if drive fails to mount
- Set `fs_passno` to 2 (the last column)

> This field is used by fsck(8) to determine the order in which filesystem checks are done at boot time.  The root filesystem should be specified with a fs_passno of 1.  Other filesystems should have a fs_passno of 2.

deployment/terraform/cloud-config/ecs-batch-user-data.yml
CloudNiner: Let me try it out again. I _think(?)_ I did earlier yesterday and it didn't work, but I don't remember for sure.

CloudNiner:
Updated. Just ran a deploy via TF and the new instance came up properly configured.

PR #165, opened 2017-03-15
Run update_job_definition.py inside docker container

CloudNiner:
## Overview

Jenkins CI jobs run on bare metal. Which means the update_job_definitions.py script does as well. CI has some version of boto3 installed, but its not necessarily the one we want and botocore on CI is out of date. This encapsulates the script dependencies and runs via docker to address this.

## Testing

Configure your VM with the env variables necessary to run `./scripts/infra` as described in deployment/README.md. Then run `./scripts/infra plan`. The script should run successfully.


deployment/aws-batch/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  update-job-defs:
```
tnation14: Can we add this service to `docker-compose.yml` so that all services are defined there?

scripts/infra
```diff
@@ -61,13 +61,15 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
...
             plan)
+                echo "Updating AWS Batch job definitions"
                 pushd "../aws-batch"
                 # TODO: Only trigger this if the pfb-analysis container has changed, otherwise skip
                 #       and reuse old version
                 # TODO: This is a bit awkward. If we `plan` but never `apply`, we'll have created
                 #       an orphaned job definition that never gets used.
-                echo "Updating AWS Batch job definitions"
-                BATCH_JOB_NAME_REVISION=$(python update_job_definition.py pfb-analysis-run-job.json "${PFB_AWS_ECR_ENDPOINT}/pfb-analysis:${GIT_COMMIT}")
+                docker-compose build
```
tnation14: If we add `update-job-defs` to the top-level `docker-compose.yml`, we can just build `update-job-defs` here without changing directories.

CloudNiner: So I explicitly kept this separate for two reasons:

1. It's a custom command only used in CI, so it keeps the main docker-compose.yml clean and allows us to continue using `docker-compose up` without explicitly choosing services to start
2. It allows us to simply delete this directory (`./deployment/aws-batch/`) once Terraform supports Batch without having to delete additional dependencies sprinkled throughout the project.

With that said, I definitely see the argument for consistency. Happy to move to the main `docker-compose.yml` if that's more compelling.

tnation14: That makes a lot of sense. Since this is (hopefully) temporary, I'm fine with leaving things as they are.

Issue #124, opened 2017-03-06
Clean out unnecessary group_vars


KlaasH:
Something I noticed looking at PR #121 but which is unrelated to that work:

I think some of the variables in `ansible/group_vars/all.example` are left over from before the analysis was containerized.  Namely these:
```
postgresql_database_name: "pfb"
# User 'gis' allows the Toole analysis scripts to work as is
postgresql_database_user: "gis"
postgresql_database_password: "gis"
postgresql_database_host: "127.0.0.1"
```
and these:
```
# These mappings allow the Toole analysis scripts to work as is without requiring a password
postgresql_hba_mapping:
  - { type: "local", database: "{{ postgresql_database_name }}", user: "{{ postgresql_database_user }}", address: "", method: "trust" }
  - { type: "host", database: "{{ postgresql_database_name }}", user: "{{ postgresql_database_user }}", address: "{{ postgresql_database_host }}/32", method: "trust" }
  - { type: "host", database: "{{ postgresql_database_name }}", user: "{{ postgresql_database_user }}", address: "0.0.0.0/0", method: "md5" }
```

The rest of the postgresql and postgis vars are now just changing the versions used by the django site, but it seems reasonable to keep them rather than be using different versions in the different parts of the project.

In unrelated news, ECS now supports docker 1.12
```
# azavea.docker
# 1.11.* to match AWS ECS
docker_version: 1.11.*
```

PR #166, opened 2017-03-15
Clean out unnecessary group_vars

KlaasH:
Removes some variables from `group_vars/all.example` that became obsolete when the analysis was dockerized.  ~~Also adds a comment to note that the postgres group_var overrides don't affect the analysis but are set up to match the versions used by the analysis, for the sake of not adding potential confusion to the project by running two different versions of postgres.~~

Connects #124.

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -7,30 +7,6 @@ pfb_aws_batch_job_definition_name: "staging-pfb-analysis-run-job"
 
 # application settings
 docker_compose_version: 1.8.0
-postgresql_database_name: "pfb"
-# User 'gis' allows the Toole analysis scripts to work as is
-postgresql_database_user: "gis"
-postgresql_database_password: "gis"
-postgresql_database_host: "127.0.0.1"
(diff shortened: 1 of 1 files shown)


deployment/ansible/group_vars/all.example
```diff
@@ -7,21 +7,24 @@ pfb_aws_batch_job_definition_name: "staging-pfb-analysis-run-job"
...
+# azavea.docker
+# 1.12.* to match AWS ECS
+docker_version: 1.12.*
+docker_options: "--storage-opt dm.basesize=20G"
+
+# azavea.terraform
+terraform_version: 0.8.7
+
+
+## PostgreSQL and PostGIS parameters ##
```
CloudNiner: What's the intent for keeping postgresql/postgresql-support/postgis options around? All of the deps are now encapsulated by Docker and we don't install the roles that these variables are used in anymore.

KlaasH: Quite so.  Somehow started thinking of the django app as running on the VM.  Removed.

PR #168, opened 2017-03-15
Restore additional schemas to database setup

KlaasH:
PR #155 removed the TDB-Tools extension and replaced the functions that were in use from it, but didn't remove the `CREATE EXTENSION "tdg";` from `setup_database.sh` and didn't replace the schema modifications made by the extension, which are relied upon by later steps.

This replaces the `CREATE EXTENSION` with some `CREATE SCHEMA` commands from https://github.com/spencerrecneps/TDG-Tools/blob/nodes/TDG%20SQL%20Tools/sql/misc/misc.sql#L22, which I think is all the analysis needs to work properly.


spencerrecneps:
Sorry I missed that. Looks like you got all the right schemas in there.

PR #173, opened 2017-03-16
Feature/awf/better analysis logging

CloudNiner:
## Overview

The analysis job logs are very difficult to read in some cases. This makes a few improvements that should make reading job logs a bit easier.

- Use `wget -nv` (no verbose) to get summary info about the download, but don't report download progress
- Replace `/usr/bin/time -v` verbose output with a custom format string that only shows elapsed time and average memory usage.
- Add `AnalysisJob.logs_url` property which makes it easy to link to CloudWatch logs from the Django application

## Demo

screen shot 2017-03-16 at 08 51 20
screen shot 2017-03-16 at 08 50 56

## Testing

Rebuild analysis container, run Germantown job locally, inspect output.

Trigger a new job via Django. Wait for it to start running, then navigate to `job.logs_url`

### DIFF
--- a/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
+++ b/deployment/aws-batch/job-definitions/pfb-analysis-run-job.json
@@ -20,6 +20,9 @@
             "name": "PFB_MAINTENANCE_WORK_MEM",
             "value": "2048MB"
         }, {
+            "name": "PFB_MAX_WAL_SIZE",
+            "value": "4096MB"
+        }, {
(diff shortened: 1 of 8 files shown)


CloudNiner:
Agree. Checkpoint warnings are disabled by default now. I also significantly bumped MAX_WAL_SIZE on the AWS jobs to 4Gb via the job definition config, which should also improve things a bit.

CloudNiner:
Connects #174

Issue #98, opened 2017-02-23
Write python script to trigger many analysis jobs at once


CloudNiner:
We're receiving the final list of cities and their input neighborhoods from data analytics. We should chat with whoever is doing the work to settle on an input format for the data so that this script can efficiently read it and trigger new jobs.

Once this task is done we should be able to start trial runs of the 300 cities deliverable.

CloudNiner:
Depends on #91, #96, #100

CloudNiner:
As per a discussion with Esther and Dan, the input to this script will be a single shapefile, where each neighborhood to run is a polygon in that file. The neighborhood name and 2-letter state abbreviation will be attributes on each polygon.

dmcglone:
Sample shapefile for three cities:

[pfb_city_boundaries.zip](https://github.com/azavea/pfb-network-connectivity/files/845891/pfb_city_boundaries.zip)
Projection: WGS 1984, SRID: 4326

CloudNiner:
Awesome, looks great. Thanks for putting this together.

CloudNiner:
@dmcglone @eneedham we may have to add an additional attribute to some of the polygons in the shapefile, which would be a public URL to retrieve the appropriate OSM data from. Some larger boundaries cannot have the necessary OSM data retrieved automatically due to Overpass API limitations. We'll be creating a fix in #176, so check back in before creating the shapefile with the full list of cities.

dmcglone:
Would this be a unique attribute per city? We could potentially add extent coordinates of each polygon as a field.

CloudNiner:
Yes, I think it would be an additional, optional attribute per polygon.

We already use the polygon extent to automatically retrieve appropriate OSM data if possible. The issue is that the API endpoint we use fails if the extent we request is too large. So we need to have manual extracts prepared for the larger boundaries. We'll likely just use the Mapzen metro extracts where possible.

PR #178, opened 2017-03-20
Feature/awf/300 cities script

CloudNiner:
## Overview

This PR adds a Django management command which pulls a shapefile from a remote public URL, creates a new AnalysisBatch object, then for each of the Polygon/MultiPolygon geometries in that file, does the following:
- Create a neighborhood for the polygon, if one does not yet exist
- Create a new AnalysisJob for the neighborhood
- Triggers each AnalysisJob if AnalysisBatch creation was successful, only if the `--submit` option was provided

Also adds a management command to cancel all jobs in an AnalysisBatch

## Demo

screen shot 2017-03-20 at 09 47 53

## Testing

Create and submit a new batch with:
```
./scripts/django-manage analysis_batch_create --submit https://s3.amazonaws.com/pfb-batch-run-inputs-us-east-1/testing/pfb_city_boundaries.zip
```
Go to the AWS Batch console, which should show three new jobs in various states of progress.

Cancel all of the jobs with:
```
./scripts/django-manage analysis_batch_cancel <batch_uuid>
```
Substituting the AnalysisBatch UUID printed when the first command was run into `<batch_uuid>`.

Connects #98


src/django/pfb_analysis/management/commands/analysis_batch_create.py
```diff
@@ -0,0 +1,109 @@
...
+                        neighborhood = neighborhoods[0]
+                    except IndexError:
+                        geom = GEOSGeometry(json.dumps(feature['geometry']))
+                        neighborhood = Neighborhood.objects.create_from_geom(geom, **{
+                            'label': '{}, {}'.format(city, state),
+                            'state_abbrev': state,
+                            'organization': user.organization,
+                            'created_by': user,
+                            'modified_by': user,
+                        })
```
KlaasH: Trying to figure out what this gets over a `get_or_create` followed by `if created` setup led to discussion about Neighborhood mutability.  Per that discussion, this behavior, where changes to a boundary are ignored if a neighborhood with that name already exists, is not what we want.  For now, we should overwrite the shapefile every time.  In the future we might want to detect changes and break the boundary out into a related model so we can update a neighborhood boundary but maintain the ability to know about/recreate older versions.

src/django/pfb_analysis/models.py
```diff
@@ -25,6 +35,50 @@ def get_neighborhood_file_upload_path(instance, filename):
     return 'neighborhood_boundaries/{0}/{1}'.format(instance.name, os.path.basename(filename))
 
 
+class NeighborhoodManager(models.Manager):
+
+    def create_from_geom(self, geom, **kwargs):
```
KlaasH: Given the comment above about overwriting the boundary each time a neighborhood shows up in the batch shapefile, this should probably be a `set_boundary_file` method on the model.

src/django/pfb_analysis/models.py
```diff
@@ -58,10 +114,50 @@ def state(self):
...
+        client = boto3.client('batch')
+        if not reason:
+            reason = 'AnalysisBatch terminated by user at {}'.format(datetime.utcnow())
+
+        batch_job_ids = self.jobs.values_list('batch_job_id', flat=True)
+        for jobs in chunks(batch_job_ids, 100):
+            jobs_list = client.describe_jobs(jobs=jobs)['jobs']
+            for job in jobs_list:
+                if job['status'] in (JobState.SUBMITTED, JobState.PENDING, JobState.RUNNABLE,
+                                     JobState.STARTING, JobState.RUNNING,):
```
KlaasH: Could add a convenience attribute like `JobState.ACTIVE_STATUSES` to capture this list, which would make it easier to adjust if AWS makes changes, especially if we end up writing these 5 out in more places.

CloudNiner:
Updated, ready for another review.

CloudNiner:
Updated again, with commit that moves job cancel logic down into new method `AnalysisJob.cancel()`

KlaasH:
If we're keeping score, another point against the django-in-the-analysis solution: fiona depends on gdal, which the django container has, so the analysis container build is failing with 
```
fiona/_transform.cpp:473:22: fatal error: cpl_conv.h: No such file or directory

 #include "cpl_conv.h"

                      ^

compilation terminated.

error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```
Adding `libgdal-dev` to the `BUILD_DEPS` in the Dockerfile clears it up, though I think it also adds a bunch of compilation warnings.

CloudNiner:
Always keeping score. I'll make this change and take a look at the warnings.

CloudNiner:
Warnings seem to be just the usual when building C programs. 

Updated again.

src/django/pfb_analysis/models.py
```diff
@@ -150,6 +242,16 @@ def logs_url(self):
...
                           batch_job_id=self.batch_job_id)
 
+    def cancel(self, reason=None):
+        """ Cancel the analysis job, if its running """
+        if not reason:
+            reason = 'AnalysisJob terminated by user at {}'.format(datetime.utcnow())
+
+        if self.status in self.Status.ACTIVE_STATUSES:
+            client = boto3.client('batch')
+            client.terminate_job(jobId=self.batch_job_id, reason=reason)
```
KlaasH: `ACTIVE_STATUSES` includes `CREATED`, so trying to terminate the remote job should either be protected by `if self.batch_job_id is not None` or wrapped in a `try: ... except botocore.exceptions.ParamValidationError: ...`

CloudNiner: Whoops. Missed that. Will fixup with the if check since that's a specific thing that should be accounted for. I'm not sure we would want to catch the broader botocore error since that could potentially hide other problems. 

It probably makes sense to catch errors in the loop and continue, logging a verbose warning if a particular job fails to cancel. Job ids can still be noted and cancelled manually if necessary via console.

Issue #145, opened 2017-03-09
Add analysis status reporting to pfb-analysis scripts


CloudNiner:
When we trigger an analysis job, we don't ever know when it finishes via the Django model, unless we directly poll for batch job status. The goal here is to have the analysis job report progress back to the Django model, so we have a better approximation of when each job completes without having to poll.

Create function that sends PATCH to /api/analysis-jobs/:jobId, call before starting each of the major sections of the analysis, updating AnalysisJob.status

Will require extra vars to be passed to the analysis job ENV.

sharph:
Instead of above strategy, consider installing django app into analysis container and updating job status via management command.

Issue #126, opened 2017-03-06
Add analysis container to docker compose configuration


flibbertigibbet:
Build analysis container with `docker-compose build`.

May need to avoid running analysis job by default with `docker-compose run`.

Related to and may depend on #125.

PR #179, opened 2017-03-20
Send status updates from analysis jobs

KlaasH:
Makes analysis jobs update their status in the management app.
Pieces::
- Adds an `AnalysisJobStatusUpdate` model to track updates for a given job. It's foreign keyed to the AnalysisJob and timestamped.
- Adds a management command, `update_status` that takes `job_id`, `status`, `step`, and `message` arguments and creates an `AnalysisJobStatusUpdate` instance.
- Pulls the whole Django app into the analysis container, so that the analysis can run the management command to update its status (rather than having to hit an endpoint). See below for some notes on this.
   - Passes all the environment variables that the Django app needs to run as container environment overrides when a job is made, since the job container now needs to be able to run the django app.
- Adds a `utils.sh` script to the analysis scripts directory with an `update_status` command that wraps the management command, so other scripts can source the file and use it.
- Adds `update_status` calls at what seemed like useful spots throughout the analysis.
- Moves the analysis container into docker compose.
- Moves the analysis from `./pfb-analysis` to `./src/analysis`. I think this makes sense in general, but the immediate issue it solves is that with the project root directory as the nearest shared ancestor of the analysis and the django app, the Docker build context had to be `./`, which meant it pulled in everything, including `./data`.

Incidental changes:
- Added `end=''` to the `print` statement in `update_job_definition.py`. This might be a platform-dependent thing, but it was adding a `\r` that ended up in the `PFB_AWS_BATCH_JOB_DEFINITION_NAME_REVISION` value in `staging-pfb-config-us-east-1.tfplan`, which made terraform mad.
- Upgraded terraform in `group_vars/all.example` to 0.8.8.  Jenkins evidently got updated, so my 0.8.7 terraform refused to mess with a configuration created by a newer version.

Notes:
- There are pros and cons to the "include Django" strategy rather than the endpoint strategy
   - Pros:
      - Slightly fewer moving pieces in the code path
      - If we end up needing to dump large quantities of data from the analysis, we'll be set up to save it right into the app's database.
   - Cons:
      - More moving pieces in the deployment/configuration setup
      - Passing the environment through is a little weird and could be a source of bugs during changes or refactors.
- There's a `message` property on `AnalysisJobStatusUpdate` that I didn't end up using, but I didn't delete it because I could imagine it being useful for error statuses.  Easy to remove, though, if we don't want it.
- The "install dependencies" step of the analysis container build already took a while, and now it takes slightly longer. It also now needs the Django app's `requirements.txt`, but to avoid re-building the dependencies layer every time the Django code changes, I set it up to copy only the `requirements.txt` first and then copy the rest of the app in a later step.

**To test**
Unfortunately passing the environment to the job on creation means it won't work to create jobs in your local instance to run on staging.  Perhaps we could set up something easier to use, but what I did to test this was:
- Deploy to staging (the usual cibuild/cipublish/infra plan/infra apply)
- Log into the staging task (`ssh -A` to bastion, ssh to instance, `docker exec -ti <container id> bash`)
- Run migrations (`./manage.py migrate`)
- In `shell_plus`, do
```
gtown = Neighborhood.objects.get(name='germantown-philadelphia-pa')
job.run()
```
Then you can see the updates with `print '\n'.join([' '.join((str(u.timestamp), str(u.status), str(u.step))) for u in job.status_updates.all()])`
image

Connects #145.  Also connects #126.


src/django/pfb_analysis/management/commands/update_status.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Command(BaseCommand):
+    help = "Update status during an analysis job"
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('job_id')
+        parser.add_argument('status')
+        parser.add_argument('step', nargs='?')
```
CloudNiner: Not sure `nargs='?'` makes sense here if a default is not provided. Making all these arguments required seems sensible.

src/django/pfb_analysis/models.py
```diff
@@ -93,10 +95,43 @@ class Status(object):
...
+        """
+        # If there's no job ID that means it's brand new
+        if self.batch_job_id is None:
+            return self.Status.CREATED
+
+        if self.batch_job_status == 'FAILED':
+            return self.Status.ERROR
+        try:
+            # If the job hasn't failed and has sent any status updates, use the latest
+            return self.status_updates.last().status
```
CloudNiner: ~~Shouldn't this explicitly order by AnalysisJobStatusUpdate.timestamp?~~ Edit: Saw the ordering meta field below.

src/django/pfb_analysis/models.py
```diff
@@ -93,10 +95,43 @@ class Status(object):
...
+        except AttributeError:
+            # Not failed but no status updates means it must be in the queue
+            return self.batch_job_status
```
CloudNiner: Do we want to be mixing AnalysisJob status and batch job status here? Would it make more sense to keep those two things completely separate, and just add an additional `status` that encapsulates the job being queued?

src/django/requirements.txt
```diff
@@ -1,3 +1,5 @@
+Django==1.10.3
```
CloudNiner: Hm. This feels super awkward since now the parent container isn't the source of the Django version, and any updates to that would be overridden by this. Not sure what a good solution would be though. A comment here reminding future us to also update the parent container if we update this _might_ help.

CloudNiner: In the same vein, might want to note with a comment which are application specific requirements and which are artifiacts of having to duplicate the parent container's requirements.txt file.

src/django/pfb_analysis/models.py
```diff
@@ -142,26 +190,22 @@ def create_environment(**kwargs):
...
-        if not self.batch_job_id:
-            return None
-        client = boto3.client('batch')
-        try:
-            jobs = client.describe_jobs(jobs=[self.batch_job_id])['jobs']
-            return jobs[0]['status']
-        except KeyError:
-            logger.exception('Error retrieving AWS Batch job status for job'.format(self.uuid))
-            return None
+    def save(self, *args, **kwargs):
```
CloudNiner: If the idea is for these objects to be immutable, is this override here necessary? Is there an advantage to doing it this way over using the two field `created_at` and `modified_at` combo used in other models/apps?

src/django/pfb_analysis/models.py
```diff
@@ -142,26 +190,22 @@ def create_environment(**kwargs):
...
         except (botocore.exceptions.BotoCoreError, KeyError):
             logger.exception('Error starting AnalysisJob {}'.format(self.uuid))
 
-    def batch_job_status(self):
-        """ Return current AWS Batch job status for this job
 
-        List of available statuses: http://docs.aws.amazon.com/batch/latest/userguide/jobs.html
-        TODO: Refactor to cache in db?
+class AnalysisJobStatusUpdate(models.Model):
+    job = models.ForeignKey(AnalysisJob, related_name='status_updates', on_delete=models.CASCADE,
```
CloudNiner: Is there a case where we would want to create an instance of this model but not have a job to attach it to? e.g. should we be using null=True and blank=True here?

src/django/pfb_analysis/management/commands/update_status.py
```diff
@@ -0,0 +1,24 @@
...
+        parser.add_argument('message', nargs='?')
+
+    def handle(self, *args, **options):
+        try:
+            job = AnalysisJob.objects.get(pk=options['job_id'])
+        except (AnalysisJob.DoesNotExist, ValueError, KeyError):
+            job = None
```
CloudNiner: See below comment about having null job_ids in the AnalysisJobStatusUpdate model.  Should we instead just let this exception bubble or pass with some other warning rather than inserting a record not attached to a job?

common.yml
```diff
@@ -5,15 +5,15 @@ services:
...
-      - PGHOST=database.service.pfb.internal
-      - PGDATABASE=pfb
-      - PGPASSWORD=pfb
-      - PGPORT=5432
-      - PGUSER=pfb
       - DJANGO_ENV=development
-      - DJANGO_SECRET_KEY=secret
       - DJANGO_LOG_LEVEL=INFO
       - AWS_PROFILE=pfb
+      - PFB_DB_HOST=database.service.pfb.internal
```
CloudNiner: Maybe I missed it among all the other changes, but these changes should also be reflected in the ECS task definition files in Terraform: 

https://github.com/azavea/pfb-network-connectivity/tree/develop/deployment/terraform/task-definitions

src/django/pfb_analysis/management/commands/update_status.py
KlaasH: Still on the fence about `message` continuing to exist at all, but requiring `step` seems sensible.  How about
```
        parser.add_argument('job_id')
        parser.add_argument('status')
        parser.add_argument('step')
        parser.add_argument('message', nargs='?', default='')
```
?

src/django/pfb_analysis/models.py
KlaasH: So I remember having some idea about it being potentially useful to have unconnected status updates when I wrote it that way, but I can't remember what the idea was and it doesn't seem useful to me now.  Will change.

src/django/pfb_analysis/models.py
KlaasH: I think my thought was that if we end up using API calls the timing could be subject to enough variables that we might want to pass the timestamp with the request to keep it precise and consistent.  Hence the overridable default rather than an `auto_now_add`.  I didn't use `PFBModel` because 1) I do think they should be immutable so the `modified` fields wouldn't be useful and 2) `PFBModel`requires `user`, which would be redundant.

KlaasH: And yeah, using `default=datetime.utcnow` without the `self.timestamp = self.timestamp.replace(tzinfo=pytz.UTC)` was giving me `/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py:1430: RuntimeWarning: DateTimeField AnalysisJobStatusUpdate.timestamp received a naive datetime (2017-03-20 18:43:16.708175) while time zone support is active.` 

Thinking it makes sense to just make this `auto_now_add` and worry about skew when it becomes something we care about, if that day every comes.

CloudNiner: Cool, yeah I think just switching the existing timestamp field to use auto_now_add is good for now.

src/django/pfb_analysis/management/commands/update_status.py
CloudNiner: Whoops, yes. +1 to your proposal

common.yml
KlaasH: This actually brings this file in line with what was already in the task definition files.  I think those were working before because of fallbacks/defaults in settings.py.

src/django/pfb_analysis/models.py
KlaasH: Yeah, I think you're right that using the batch job status as the status isn't great.  It would make `AnalysisJob.Status` an incomplete list, with the extra members being outside the app and beyond our control.  I'll add `QUEUED` and use it for anything that's been sent (i.e. has a batch ID) but hasn't provided updates yet.  I think it still makes sense to check `FAILED`, though, since a job that crashes won't send a status update to tell us so, so it would look like it was stuck on the last milestone it successfully passed.

KlaasH:
I think I've addressed all the comments.  I'll rebase the fixups in before merging.

src/django/pfb_analysis/models.py
CloudNiner: 👍 agree that still checking failed is good.

flibbertigibbet:
Getting the following error on the analysis jobs API view:
image

Destroyed/recreated all docker containers; migrations have all been run.

Issue #94, opened 2017-02-23
Add analysis job list view for a user in angular app


CloudNiner:

CloudNiner:
Depends on #93

CloudNiner:
List item should contain:
- [ ] Django job ID
- [ ] Batch job ID
- [ ] Batch job status
- [ ] (?) Link to AWS Batch Job CloudWatch logs
- [ ] If status == success, links to download files (overall_scores.csv, neighborhood_ways.zip)

PR #180, opened 2017-03-20
Feature/kak/analysis job list view#94

flibbertigibbet:
Adds a list view for analysis jobs. Use it as the root view, redirected to after login. Based off of the existing `BoundaryUpload` views, which are deleted here (we will add neighborhood views instead).

Connects #94. Does not add links for downloading the logs or result files, as the API components aren't ready yet to deliver those.

image


src/angularjs/src/app/components/analysis-jobs/analysis-jobs-status.filter.js
```diff
@@ -0,0 +1,31 @@
...
+ * @description
+ * Transforms analysis job status into user friendly string
+ */
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function displayStatus() {
+        var statuses = {
+            CREATED: 'Created',
```
CloudNiner: These are defined both here and in analysis-jobs.service.js. Consider defining in a constant service and reusing them here.

flibbertigibbet: Yeah, I was wondering how to improve on this. Just copied the pattern from the extant 'batch upload' filters. It is slightly different than a reverse of the map in the service; the descriptive strings here are longer than those used for the drop-downs (defined in the service). Ideally, I think we should be able to get all of these dynamically from the API though.

CloudNiner: Yeah, from the API is probably best, but we don't have an endpoint for that yet. Moving to a constants file and making each item in a dict follow the form:
```
var status = {
  IMPORTING: {
    short: 'shortdesc',
    long: 'longer description'
  },
  ...
}
```
Might be a good middle ground for now. In the place where keys/values are reversed, just loop the new status dict to find the key you want, it's not like that's going to be a big performance hit. 

Or just leave as is and make a separate issue to add the endpoint and fix this up, but either way #179 added an additional `QUEUED` status that should be reflected here before merging.

flibbertigibbet:
Added queued status and constants file for job status options.

PR #181, opened 2017-03-21
Allow analysis to run on manually provided OSM file

CloudNiner:
## Overview

Some boundaries are too large to pull via the Overpass API. As a workaround for now, we allow a neighborhood to specify the URL that its associated OSM data is available from.

## Notes

To simplify the logic in entrypoint.sh, I updated all of the existing OSM file references to use the `.osm` file extension. The analysis now explicitly looks for this when running and fails if it doesn't find a matching file.

## Demo

screen shot 2017-03-21 at 10 11 23

## Testing

Locally:
```
# Unzipped OSM file
PFB_OSM_FILE_URL=https://s3.amazonaws.com/test-pfb-inputs/germantown/gtown_westside.osm AWS_PROFILE=pfb ./scripts/run-local-analysis https://s3.amazonaws.com/test-pfb-inputs/germantown/gtown_westside.zip pa 42

# Zipped OSM file
PFB_OSM_FILE_URL=https://s3.amazonaws.com/test-pfb-inputs/germantown/gtown_westside.osm.zip AWS_PROFILE=pfb ./scripts/run-local-analysis https://s3.amazonaws.com/test-pfb-inputs/germantown/gtown_westside.zip pa 42

# bzipped OSM file from Metro Extracts
PFB_OSM_FILE_URL=https://s3.amazonaws.com/metro-extracts.mapzen.com/philadelphia_pennsylvania.osm.bz2 AWS_PROFILE=pfb ./scripts/run-local-analysis https://s3.amazonaws.com/test-pfb-inputs/germantown/gtown_westside.zip pa 42
```

On AWS:

Go to http://localhost:9200/api/analysis_jobs/ and trigger a new job, ensuring that you enter a URL into the `OSM extract url` field. Then go to the Batch console on staging and verify that the ENV var `PFB_OSM_FILE_URL` was added to the job.


CloudNiner:
Once this is merged I'll start a new Manhattan/Brooklyn/Bronx job using the Mapzen NYC Metro Extract file.

CloudNiner:
😮 

Yeah we should definitely clip the OSM data to only the area necessary for the analysis. From my understanding of what the analysis is doing we could also clip this import to boundary_edge + 2 * max_trip_distance by default. @spencerrecneps ?

CloudNiner:
Which step had its running time go out of control?

KlaasH:
I think many of the steps were slower, but the `ANALYZE received.neighborhood_ways_net_vert;` in `build_network.sql` never actually finished even though I let it run for a few hours.

You mean half of NB_MAX_TRIP_DISTANCE like we do for the census blocks, right?  Not twice?

CloudNiner:
Whoops. Yes, half NB_MAX_TRIP_DISTANCE

spencerrecneps:
Sorry. I've been meaning to take a closer look at the NB_MAX_TRIP_DISTANCE changes that have come through but am just now getting to it. The trip distance actually refers to one-way travel, so we shouldn't be dividing it in half.

WRT the question above: yes, we can clip the OSM input to a buffer around the neighborhood boundary. No need to keep anything outside of that area. But, to reiterate, the buffer distance should be equal to NB_MAX_TRIP_DISTANCE, not divided in half.

CloudNiner:
Ok great. Thanks for clarifying. 

@KlaasH I'll update the analysis to clip OSM data to NB_MAX_TRIP_DISTANCE and remove the divide by 2 used when setting NB_BOUNDARY_BUFFER, which means I'll likely remove NB_BOUNDARY_BUFFER entirely.

KlaasH:
I'm still not on the same page about the buffer.  We're not caring about any destinations outside the boundary, right?  Just routes outside the boundary that could get you from one point inside the boundary to another?  A one-way trip that starts and ends inside the boundary can't get more than 1/2 the max distance outside of it.

CloudNiner:
I believe the analysis allows for trips to start and end outside the neighborhood boundary.

KlaasH:
Ok, yeah.  Looking back at the queries I see e.g.
```
EXISTS (
            SELECT  1
            FROM    neighborhood_boundary AS b
            WHERE   ST_Intersects(b.geom,r1.geom)
)
```
in reachable_roads_low_stress_calc.sql.  I think the constraints like that were what made me think we didn't care about destinations outside the boundary, but I see now that it's only constraining one end.  Sorry for the confusion.

spencerrecneps:
@KlaasH The analysis allows for a trip to start within the neighborhood but end outside of it. For example, a person living near the edge of the neighborhood may attend school outside of the boundary. We want to be able to model their trip, which would start within the neighborhood but might end up to MAX_TRIP_DISTANCE outside of it.

spencerrecneps:
FWIW it might not be a bad idea to leave the neighborhood buffer as a separate variable. It's possible we could decide to restrict external trips as a future refinement. I'd vote for leaving it in there and just setting it to max distance for the time being.

CloudNiner:
Updated. Added `osmconvert` to clip to computed bbox boundary before import.

http://wiki.openstreetmap.org/wiki/Osmconvert

Demo:

_Result from running command in PR testing instructions which downloads the Philadelphia metro extract_

screen shot 2017-03-22 at 13 33 59

PR #195, opened 2017-03-21
Add missing Angular constants

flibbertigibbet:
Intended to be added with #180 refactoring.


flibbertigibbet:
Merging this small fixup.

Issue #142, opened 2017-03-09
Implement "List Neighborhoods" view in Angular app


CloudNiner:
Reuse styling and table functionality from #94 

Can be linked to in app header with the title "Neighborhoods". Should only show neighborhoods attached to the user's organization.

Issue #140, opened 2017-03-09
Implement "Upload Neighborhood" view in Angular app


CloudNiner:
Change "Upload Data" button to "Upload Neighborhood" and re-use the existing "Upload Boundary" workflow/views.

Issue #139, opened 2017-03-09
AnalysisJob list view: Filtering


CloudNiner:
Filter list view table on:
- [ ] ~~django job id~~
- [ ] ~~batch job id~~
- [ ] ~~batch job status~~
- [ ] status
- [ ] neighborhood

CloudNiner:
Depends on #94

flibbertigibbet:
Currently has filters in place for neighborhood and status. Should this be changed to 'search by UUID or batch job UUID'? Would require implementing a text-field search that we currently do not have in the UI (just drop-down filtering).

CloudNiner:
I just updated the original comment to remove the search on the django job and batch job uuids, since those are additional nice-to-haves that the client isn't specifically requesting. So if we have status and neighborhood filtering then this issue is done for now as well.

flibbertigibbet:
Neighborhood filtering in place; status filtering will need to be updated since removed as filterable API field (do client-side?)

CloudNiner:
While it might be pretty inefficient (since you'd have to load the entire queryset unless you did a custom sql query), it should be possible to do a custom method filter using django-filters. That's probably a better first shot than doing client side filtering, which could be awkward due to pagination as the job list gets large.

Example of the Django filters boilerplate here:
https://github.com/azavea/climate-change-api/blob/develop/django/climate_change_api/climate_data/filters.py#L23

PR #197, opened 2017-03-21
Feature/kak/neighborhood list view#142

### LINKED ISSUE #194: Non-administrative users cannot view neighborhoods
The `Neighborhood` API endpoint has `RestrictedCreate` permissions, which should mean any user should be able to view neighborhoods, and only administrators can create/update/delete. However, only administrators can view neighborhoods.

flibbertigibbet:
Adds a neighborhood list view and upload view to the UI. Connects #142. Connects #140. Connects #139.

Also updates API permissions to fix #194. Makes neighborhoods viewable by all users; filters to user's organization.

Also also fixes filtering analysis jobs by status.

image


flibbertigibbet:
Administrators can view neighborhoods across organizations; not sure if we should update `OrgAutoFilterBackend` to remove the administrator exemption, add a new backend filter for neighborhoods, or simply show all with the organization in the UI.

CloudNiner:
We will likely have to address this later, but its probably appropriate for now for all users in an org, even administrators to be scoped to seeing neighborhoods in their org.

flibbertigibbet:
I'll add a filter for users/orgs that allows admins to view across orgs, and modify the existing to always limit to org (for analysis jobs and neighborhoods).

src/angularjs/src/app/neighborhoods/list/neighborhoods-list.controller.js
```diff
@@ -0,0 +1,88 @@
...
+            var params = _.merge({}, defaultParams, nextParams);
+            $state.go('neighborhood.list', params, {notify: false});
+            getNeighborhoods(params);
+        }
+
+        function getPrev() {
+            var params = _.merge({}, defaultParams, prevParams);
+            $state.go('neighborhood.list', params, {notify: false});
+            getNeighborhoods(params);
+        }
```
KlaasH: All this doubled code for next and previous isn't the prettiest.  I like a version with more brackets but fewer lines better:
```
function NeighborhoodListController($state, $stateParams, $scope, Pagination, Neighborhood) {
    var ctl = this;

    var defaultParams = {
        limit: null,
        offset: null
    };

    initialize();

    function initialize() {
        ctl.pages = {
            'next': {
                hasPage: false
            },
            'previous': {
                hasPage: false
            }
        };
        ctl.getPage = getPage;
        ctl.neighborhoods = [];

        ctl.deleteUpload = deleteUpload;
        ctl.filters = {};

        getNeighborhoods();
    }

    function getNeighborhoods(params) {
        params = params || $stateParams;
        Neighborhood.query(params).$promise.then(function(data) {
            ctl.neighborhoods = _.map(data.results, function(obj) {
                return new Neighborhood(obj);
            });

            _.forEach(['previous', 'next'], function (direction) {
                if (data[direction]) {
                    ctl.pages[direction].hasPage = true;
                    ctl.pages[direction].params = Pagination.getLinkParams(data[direction]);
                } else {
                    ctl.pages[direction] = { hasPage: false };
                }
            });
        });
    }

    function getPage(direction) {
        var params = _.merge({}, defaultParams, ctl.pages[direction].params);
        $state.go('neighborhoods.list', params, {notify: false});
        getNeighborhoods(params);
    }
```

Then the template lines would be
```
        <button class="btn-success btn" ng-click="neighborhoodList.getPage('previous')"
                ng-disabled="!neighborhoodList.pages.previous.hasPage">
```
and
```
        <button class="btn-success btn" ng-click="neighborhoodList.getPage('next')"
                ng-disabled="!neighborhoodList.pages.next.hasPage">
```

src/angularjs/src/app/neighborhoods/list/neighborhoods-list.controller.js
```diff
@@ -0,0 +1,88 @@
...
+
+        function getNext() {
+            var params = _.merge({}, defaultParams, nextParams);
+            $state.go('neighborhood.list', params, {notify: false});
+            getNeighborhoods(params);
+        }
+
+        function getPrev() {
+            var params = _.merge({}, defaultParams, prevParams);
+            $state.go('neighborhood.list', params, {notify: false});
```
KlaasH: This should be `neighborhoods` (with an S)

src/django/pfb_network_connectivity/filters.py
```diff
@@ -13,6 +13,16 @@ class OrgAutoFilterBackend(filters.BaseFilterBackend):
...
+        if queryset.model == Organization:
+            return queryset.filter(uuid=request.user.organization_id)
+        else:
+            return queryset.filter(organization=request.user.organization)
+
+
+class OrgOrAdminAutoFilterBackend(filters.BaseFilterBackend):
+    """Filter that allows non-admin users to see only their own organization's objects."""
+
+    def filter_queryset(self, request, queryset, view):
```
KlaasH: This change surfaced an existing bug in the AnalysisJob list--AnalysisJob doesn't have an `organization` attribute.  Nor should it, since it's foreign keyed to Neighborhood, which does, so it would be redundant.  But that means filtering on `organization` in these two filters throws an error (which was hidden before for users from the admin org, since they got the unfiltered results).  Simplest thing is probably to add another special clause, i.e.
```
if queryset.model == AnalysisJob:
    return queryset.filter(neighborhood__organization=request.user.organization)
```
Though it would be nice if there were a more general way to do it.

src/angularjs/src/app/neighborhoods/list/neighborhoods-list.controller.js
flibbertigibbet: The pagination code came from the pre-existing view, and is duplicated across all of the other paginated views. I'm disinclined to mess with it. If you really think this is something to address, how about creating an issue to factor out pagination?

src/django/pfb_network_connectivity/filters.py
flibbertigibbet: I haven't been able to see that issue yet, as the analysis job endpoint is broken since #179.

CloudNiner: Is one of you able to review #200? The fix for #179 lives there.

flibbertigibbet:
Kept working on this branch to add in the neighborhood create view. Can break that out into a separate PR if needed.

src/django/pfb_network_connectivity/filters.py
flibbertigibbet: Ah, okay. I'll take a look.

CloudNiner:
Can you update this PR with the issues it now encompasses since it seems to have expanded in scope?

flibbertigibbet:
Updated with backend filtering fix for analysis job status and assorted small fixes. Think this is good to go for another look.

src/angularjs/src/app/neighborhoods/list/neighborhoods-list.controller.js
KlaasH: Fair enough.  Probably not worth messing with.  I'll make an issue if it gets under my skin again.

src/angularjs/src/app/neighborhoods/create/neighborhoods-create.html
```diff
@@ -0,0 +1,45 @@
+<pfb-navbar></pfb-navbar>
+
+<div class="container">
+  <div class="row">
+    <div class="col-md-10 col-md-push-1">
+      <h1>Upload Data</h1>
```
KlaasH: Should say "Upload Neighborhood" (might not be the final text, but better than "Data", which doesn't really apply).

Issue #188, opened 2017-03-21
Add AnalysisJob.running_time property


CloudNiner:
AnalysisJob.run() should set the first AnalysisJobStatusUpdate.

This property should show the job running time in seconds as the difference between the earliest and latest status update

Issue #182, opened 2017-03-21
Log number of census blocks in a job to console


CloudNiner:
During analysis run

PR #200, opened 2017-03-22
Feature/awf/analysis job running time

CloudNiner:
## Overview

Bah, apparently I never made a PR for this.

This adds the following properties to AnalysisJob:
- `running_time` 
- `start_time`

It also returns these properties in the API. 

Also fixes the issue with status no longer being a DB field, causing FilterSet issues. I fixed this in the simplest way possible by removing `status` filtering. If we want this back we'll need to explore other solutions later.

### Notes

Added a psql command in the analysis to print the number of census blocks in the analysis after import.

## Testing Instructions

Trigger a new AnalysisJob in the API, then view that job's detail view. You'll see the new properties, without errors.

Connects #182, #188

### DIFF
--- a/src/analysis/import/import_neighborhood.sh
+++ b/src/analysis/import/import_neighborhood.sh
@@ -102,6 +102,9 @@ then
                 AS boundary WHERE NOT ST_DWithin(blocks.geom, boundary.geom, \
                 ${NB_BOUNDARY_BUFFER});"
         echo "DONE: Finished removing blocks outside buffer"
+        echo "Census Blocks in analysis:"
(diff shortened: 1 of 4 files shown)


src/analysis/import/import_neighborhood.sh
```diff
@@ -102,6 +102,9 @@ then
                 AS boundary WHERE NOT ST_DWithin(blocks.geom, boundary.geom, \
                 ${NB_BOUNDARY_BUFFER});"
         echo "DONE: Finished removing blocks outside buffer"
+        echo "Census Blocks in analysis:"
+        psql -h "${NB_POSTGRESQL_HOST}" -U "${NB_POSTGRESQL_USER}" -d "${NB_POSTGRESQL_DB}" \
+            -c "SELECT count(*) as total_census_blocks FROM neighborhood_census_blocks;"
```
flibbertigibbet: That's handy.

Issue #199, opened 2017-03-22
Jobs failing in staging environment due to connection error


CloudNiner:
See: https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=/aws/batch/job;stream=staging-pfb-analysis-run-job--102--04cac264/15b96328-9eb3-4e3b-af5a-056180f0c6ce/caf09218-d728-42a0-a361-3b6503c8bf06

Relevant stacktrace:
```
12:42:17
File "/usr/local/lib/python2.7/dist-packages/django/db/backends/postgresql/base.py", line 176, in get_new_connection
12:42:17
connection = Database.connect(**conn_params)
12:42:17
File "/usr/local/lib/python2.7/dist-packages/psycopg2/__init__.py", line 164, in connect
12:42:17
conn = _connect(dsn, connection_factory=connection_factory, async=async)
12:42:17
django.db.utils.OperationalError: FATAL: password authentication failed for user "pfb"
12:42:17
FATAL: password authentication failed for user "pfb"
```

PR #201, opened 2017-03-22
Add options for running development jobs

KlaasH:
## Overview

Since we can't run local development jobs on the staging AWS Batch cluster after PR #179, provides two interim options:
- A management command to run a job from ECS "Run new task".  Requires that the neighborhood you want to run already exist.
- To make it easier to run the analysis locally, it prints the command that will do it, complete with job ID and shapefile URL.  The more complete version of this approach, where the django container actually connects to the host Docker daemon and starts the analysis job, seems like it would add a lot more work for not that much gain.  Indeed, opt-in (i.e. you have to copy the command and run it to make it go rather than having to cancel the run if you don't want it) seems like possibly the right choice for local development.

### Demo

Local, assuming you have a `germantown-philadelphia-pa` neighborhood:
image

On staging:
image

Actually this is failing with "STOPPED (CannotPullContainerError: Error: image pfb:7a01b08)". Will investigate further.

### Notes

Pretty soon we should have front-end tools to create jobs and neighborhoods, so the process on staging should be pretty smooth.  But the problem of not having a full production-like environment for individual development remains.  The most complete approach--individual app and batch clusters for each developer--would be some work to set up and maintain, not to mention the potential for increased AWS costs.

## Testing Instructions

See Demo, but here are the commands for the local version:
In `shell_plus`:
```
user = PFBUser.objects.get(email='systems+pfb@azavea.com')
gtown = Neighborhood.objects.get(name='germantown-philadelphia-pa', organization=user.organization)
job.run()
```
Then that will print a command that you can copy and run in the VM to run the analysis and save updates to the job.

Connects #199

### DIFF
--- a/src/django/pfb_analysis/management/commands/run_analysis_job.py
+++ b/src/django/pfb_analysis/management/commands/run_analysis_job.py
@@ -1,0 +1,28 @@
+from django.core.management.base import BaseCommand
+from django.contrib.auth import get_user_model
+
+from pfb_analysis.models import AnalysisJob, Neighborhood
+
+
+class Command(BaseCommand):
+    help = """ Create and run an AnalysisJob for the given neighborhood
+
+    The neighborhood name must match an existing Neighborhood for the user's organization.
+
+    """
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('neighborhood')
+        parser.add_argument('--user', default='systems+pfb@azavea.com', type=str)
+
+    def handle(self, *args, **options):
+        UserModel = get_user_model()
+        user = UserModel.objects.get(email=options['user'])
+        neighborhood = Neighborhood.objects.get(name=options['neighborhood'],
+                                                organization=user.organization)
+        job = AnalysisJob.objects.create(neighborhood=neighborhood,
+                                         created_by=user,
+                                         modified_by=user)
+        job.run()
+        self.stdout.write('Started job {} for {}'.format(job.batch_job_id, neighborhood.name))
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -293,14 +293,23 @@ class AnalysisJob(PFBModel):
             logger.warn('Attempt to re-run job: {}. Skipping.'.format(self.uuid))
             return
 
-        client = boto3.client('batch')
-
         # Since we run django manage commands in the analysis container, it needs a copy of
         # all the environment variables that this app needs, most of which are conveniently
         # prefixed with 'PFB_'
         # Set these first so they can be overridden by job specific settings below
         environment = {key: val for (key, val) in os.environ.items()
                        if key.startswith('PFB_') and val is not None}
+        # For the ones without the 'PFB_' prefix, send the settings rather than the original
+        # environment variables because the environment variables might be None, which is not
+        # acceptable as a container override environment value, but the settings values will be set
+        # to whatever they default to in settings.
+        environment.update({
+            'DJANGO_ENV': settings.DJANGO_ENV,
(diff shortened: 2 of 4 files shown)


KlaasH:
Note: the changes to `environment` will conflict with #181, but should be easy to reconcile.  I wanted to use the environment for the log statement rather than write the param names out a bunch more times.

src/django/pfb_analysis/management/commands/run_analysis_job.py
```diff
@@ -0,0 +1,28 @@
...
+    def handle(self, *args, **options):
+        UserModel = get_user_model()
+        user = UserModel.objects.get(email=options['user'])
+        neighborhood = Neighborhood.objects.get(name=options['neighborhood'],
+                                                organization=user.organization)
+        job = AnalysisJob.objects.create(neighborhood=neighborhood,
+                                         created_by=user,
+                                         modified_by=user)
+        job.run()
+        self.stdout.write('Started jobs {} for {}'.format(job.batch_job_id, neighborhood.name))
```
CloudNiner: Probably should use singular 'job' here in the format message.

src/django/pfb_analysis/models.py
```diff
@@ -159,30 +159,42 @@ def create_environment(**kwargs):
...
+        # For the ones that aren't, send the settings rather than the original environment
+        # variables because the environment variables might be None, which is not acceptable
+        # as a container override environment value, but the settings values will be set
+        # to whatever they default to in settings.
+        environment.update({
+            'DJANGO_ENV': settings.DJANGO_ENV,
+            'DJANGO_LOG_LEVEL': settings.DJANGO_LOG_LEVEL,
+            'AWS_DEFAULT_REGION': settings.AWS_REGION,
+        })
+        environment.update(**{key: val for (key, val) in os.environ.items()
```
CloudNiner: No need to use the `**` here, update takes a dict directly without having to expand to kwargs.

src/django/pfb_analysis/models.py
```diff
@@ -159,30 +159,42 @@ def create_environment(**kwargs):
...
+                              if key.startswith('PFB_') and val is not None})
+
+        # Workaround for not being able to run development jobs on the actual batch cluster:
+        # bail out with a helpful message
+        if settings.DJANGO_ENV == 'development':
```
CloudNiner: Many thumbs up. This is a great interim solution that gets us 90% of the way there with 10% of the effort.

src/django/pfb_network_connectivity/settings.py
```diff
@@ -196,7 +196,7 @@
         },
     },
     'loggers': {
-        'django': {
```
CloudNiner: The configuration for the `django` logger should be left in, as that captures all messages from the `django` namespace. It makes sense for the `pfb_analysis` logger to duplicate the configuration of the `django` logger

src/django/pfb_analysis/models.py
```diff
@@ -184,33 +184,42 @@ def create_environment(**kwargs):
             logger.warn('Attempt to re-run job: {}. Skipping.'.format(self.uuid))
             return
 
-        client = boto3.client('batch')
-
-        # Since we run django manage commands in the analysis container, it needs a copy of
-        # all the environment variables that this app needs, most of which are conveniently
-        # prefixed with 'PFB_'
-        # Set these first so they can be overridden by job specific settings below
```
CloudNiner: Should we be clobbering these values with settings from `os.environ`? I would think we'd want it the other way around, where custom, explicitly set options overwrite whatever defaults we pull from the environment.

KlaasH: Good point. I don't know if the PFB variables coming into the Django container from the environment are likely to end up needing overrides, but certainly we shouldn't clobber them if they do...

PR #206, opened 2017-03-23
Add missing CANCELLED job status

CloudNiner:
## Overview

Missed this additional status in the JS after adding cancelled status in #202


flibbertigibbet:
Ah, okay, I guess that status was what the missing migration was for.

PR #209, opened 2017-03-24
Feature/awf/analysis export dest tables

### LINKED ISSUE #157: Add tables to export
The following tables should be made available via export:
- neighborhood_census_blocks
- neighborhood_colleges
- neighborhood_community_centers
- neighborhood_medical
- neighborhood_parks
- neighborhood_retail
- neighborhood_schools
- neighborhood_social_services
- neighborhood_supermarkets
- neighborhood_universities

CloudNiner:
## Overview

Exports additional destination tables from the analysis to GeoJSON.

## Demo

**neighborhood_census_blocks:**
screen shot 2017-03-24 at 08 10 15

### Notes

No issues installing `gdal-bin` via apt, but the version is quite old (1.10.x) and so some features were missing that required workarounds (comments in code).

A bunch of the data and geometries are null in the Germantown analysis. We may have to investigate this later.

## Testing Instructions

Run local analysis job, inspect files in the output data dir `./data/output/local-....` after completion. There should be a file for each of the items listed in #157 

Closes #157

### DIFF
--- a/src/analysis/scripts/export_connectivity.sh
+++ b/src/analysis/scripts/export_connectivity.sh
@@ -7,7 +7,6 @@ set -e
 cd $(dirname "$0")
 source ./utils.sh
 
-NB_OUTPUT_SRID="${NB_OUTPUT_SRID:-2163}"
 NB_POSTGRESQL_HOST="${NB_POSTGRESQL_HOST:-127.0.0.1}"
 NB_POSTGRESQL_DB="${NB_POSTGRESQL_DB:-pfb}"
 NB_POSTGRESQL_USER="${NB_POSTGRESQL_USER:-gis}"
@@ -98,7 +97,6 @@ function ec_export_destination_geojson() {
   # Use geom_poly because its the source field in all dest tables -- geom_pt is only
   #   derived field in some tables
   ogr2ogr -f GeoJSON "${FILENAME}" \
-          -s_srs "EPSG:${NB_OUTPUT_SRID}" \
           -t_srs EPSG:4326 \
           "PG:host=${NB_POSTGRESQL_HOST} dbname=${NB_POSTGRESQL_DB} user=${NB_POSTGRESQL_USER}" \
           -sql "select *, geom_poly from ${EXPORT_TABLENAME}"
--- a/src/analysis/scripts/export_connectivity.sh
+++ b/src/analysis/scripts/export_connectivity.sh
@@ -7,7 +7,7 @@ set -e
 cd $(dirname "$0")
 source ./utils.sh
 
-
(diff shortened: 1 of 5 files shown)


src/analysis/scripts/export_connectivity.sh
```diff
@@ -95,8 +124,28 @@ then
         # Export neighborhood_ways as SHP
         ec_export_table_shp "${OUTPUT_DIR}" "neighborhood_ways"
 
+        # Export census blocks
+        ec_export_table_shp "${OUTPUT_DIR}" "neighborhood_census_blocks"
+        ec_export_table_geojson "${OUTPUT_DIR}" "neighborhood_census_blocks"
```
CloudNiner: This file will be massively large for pretty much any larger analysis, and we shouldn't be displaying it on a web map anyways.

The Germantown one is 4.2mb and geojson.io struggles.

src/analysis/scripts/export_connectivity.sh
```diff
@@ -75,6 +75,35 @@ function ec_export_table_csv() {
          -c "\COPY ${EXPORT_TABLENAME} TO '${FILENAME}' WITH (FORMAT CSV, HEADER)"
 }
 
+function ec_export_table_geojson() {
+  OUTPUT_DIR="${1}"
+  EXPORT_TABLENAME="${2}"
+
+  FILENAME="${OUTPUT_DIR}/${EXPORT_TABLENAME}.geojson"
+  ogr2ogr -f GeoJSON "${FILENAME}" \
+          -t_srs EPSG:4326 \
```
KlaasH: Is there a reason the other geojson function includes `-s_srs "EPSG:${NB_OUTPUT_SRID}"` and this one doesn't?

CloudNiner: I ran into an error where for some of the destinations tables ogr2ogr threw an error about not finding appropriate geometries. I can remove it and test again, maybe the error was fixed via further munging of these commands.

KlaasH: I'm inclined to say we should err on the side of explicitness--we're confident we know what SRID the tables are in, so might as well tell `ogr2ogr` and not rely on it to figure it out.  Then again if the errors you were seeing were actually related to something else there might be advantages to letting it autodetect, e.g. in case we do ever end up making tables in other coordinate systems for some reason.

CloudNiner: Setting `-s_srs` doesn't seem to be necessary for now. I pushed a commit to remove it. It can be added back in later if it causes trouble for other neighborhoods.

PR #212, opened 2017-03-24
Create analysis job view

### LINKED ISSUE #141: Implement "Run Analysis" view in Angular app
Create a duplicate view using the same style as the "Upload Neighborhood" view

Depends on #140, neighborhoods must exist to start a new analysis.

KlaasH:
## Overview

Adds a "Run Analysis" view to the management app, for kicking off analysis jobs.

### Demo

image

image

### Notes

- I added an `all` route to the neighborhoods endpoint because the droplist in this view and in the filter on the list view shouldn't be paginated.  Neighborhoods aren't very big, so I didn't bother to limit the fields returned, but we could if it seems laggy at all.
- I added a `neighborhood_label` SerializerMethodField to AnalysisJob so that we can show the label in the list.  Including the whole neighborhood with the job is pretty easy for GET requests but makes saving a bit more complicated, and I think label and ID (which is already there, in the `neighborhood` field) are all we're likely to need.

## Testing Instructions

With the AngularJS app running, go to http://localhost:9301/#/analysis-jobs/ and click 'Run Analysis'.  Pick a neighborhood and submit the job and it should put you back on the list view with your new job at the top, with status "Created".  Your Django container will have logged the command you can use to run that analysis job locally, so copy that and run it if you want to see the status change.

Resolves #141.

### DIFF
--- a/src/angularjs/src/app/analysis-jobs/create/analysis-jobs-create.controller.js
+++ b/src/angularjs/src/app/analysis-jobs/create/analysis-jobs-create.controller.js
@@ -1,0 +1,46 @@
+/**
+ * @ngdoc controller
+ * @name pfb.analysisJobs.create.controller:AnalysisJobCreateController
+ *
+ * @description
+ * Controller for creating/running an analysis job
+ *
+ */
+(function() {
+    'use strict';
+
+    /** @ngInject */
+    function AnalysisJobCreateController($state, $filter, toastr, AnalysisJob, Neighborhood) {
+        var ctl = this;
+
+        function initialize() {
+            ctl.neighborhoods = Neighborhood.all();
+            ctl.neighborhoods.$promise.then(function(data) {
+                ctl.neighborhoods = data;
+            });
+        }
+
+        initialize();
+
+        ctl.create = function() {
+            var submitToast = toastr.info('Submitting analysis job. Please wait...',
+                                          {autoDismiss: false});
+
+            var job = new AnalysisJob({
+                neighborhood: ctl.neighborhood.uuid,
+                osm_extract_url: ctl.osmUrl
+            });
+            job.$save().then(function() {
+                toastr.clear(submitToast);
+                toastr.success('Successfully submitted analysis job');
+                $state.go('analysis-jobs.list');
+            }).catch(function(error) {
+                toastr.clear(submitToast);
+                toastr.error('Error submitting analysis job: ' + error);
+            });
+        };
+    }
+    angular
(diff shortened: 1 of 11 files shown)


KlaasH:
The "pick an arbitrary number that's larger than the number of neighborhoods we ever expect to see" thing is bothering me, I guess for aesthetic reasons since there's no actual risk that a limit like 10,000 would ever cause a problem.

The endpoint already has LimitOffsetPagination, so I could just use the existing endpoint with `limit=10000` to get the whole list.  But instead I've gone in the other direction and added field limiting so that the new endpoint (which I also renamed to `list`) returns only UUID, label, and state.  Which might make a difference if the list gets into the hundreds, though it also might not.

Thoughts?

CloudNiner:
Rather than having a separate limited list endpoint, is there any disadvantage to having the regular list endpoint (at /api/neighborhoods/) just return UUID, label and state? Or alternatively, those three fields plus whatever other fields we want to allow filtering/sorting on? The the detail still returns all the available neighborhood fields.

flibbertigibbet:
We added a custom pagination parameter for 'all' records [here](https://github.com/azavea/ashlar/blob/develop/ashlar/pagination.py).

KlaasH:
Ok, yeah, that approach looks good.  I did a version that just changes the bits we want to change then calls `super`, though being able to call `super().paginate_queryset` required one slightly-squirrely bit in `get_limit`.  See what you think.

src/django/pfb_network_connectivity/settings.py
```diff
@@ -222,7 +222,7 @@
         'rest_framework.authentication.TokenAuthentication',
         'rest_framework.authentication.SessionAuthentication',
     ),
-    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
+    'DEFAULT_PAGINATION_CLASS': 'pfb_network_connectivity.pagination.OptionalLimitOffsetPagination',
```
flibbertigibbet: We might not want to make this the default. Or, be careful to override with the required-limit version for endpoints that might be overloaded trying to fetch all results (thinking of the analysis jobs endpoint in particular).

KlaasH: Yeah, hard to know whether we'll want no limits, restricted limits, or neither (we don't currently have a `max_limit` set on the default paginator) for other endpoints, but making the `limit=all` option opt-in makes sense.  I put back the default and set the modified one as the pagination class on NeighborhoodViewSet.

PR #214, opened 2017-03-24
Feature/awf/django overall scores

### LINKED ISSUE #185: Save analysis results table overall_scores back to AnalysisJob
Consider `overall_scores` property on model as a JSONField

We'll need to support filtering/sorting by overall scores results across many AnalysisJobs

CloudNiner:
## Overview

Add overall_scores field to AnalysisJob as jsonb, and push analysis results from CSV back to Django app, so we can easily sort/filter on these scores later.

### Demo

screen shot 2017-03-24 at 10 50 20

## Testing Instructions

- Start new AnalysisJob for Germantown, either via API UI or Django shell
- Copy the run-local-analysis command printed by AnalysisJob.run() in the django docker console
- Run job in VM
- When job is finished, overall_scores should have data for the job id you just ran, inspect in Django console
- Overall scores should not print in analysis_jobs list view in API

Closes #185


src/django/pfb_analysis/management/commands/load_overall_scores.py
```diff
@@ -0,0 +1,68 @@
...
+                  'from file {}'.format(job_uuid, csv_filename))
+            return
+
+        with open(csv_filename, 'r') as csv_file:
+            reader = csv.DictReader(csv_file)
+            results = {}
+            for row in reader:
+                key_column_value = row.pop(key_column)
+                metric = self.clean_metric_dict(row.copy())
+                results[key_column_value] = metric
```
KlaasH: The `notes` fields include newlines.  I was going to say we should strip them here, but actually the script that writes them is intending for them not to be there but needs a `'g'` flag added to the end of all the `regexp_replace` calls: https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/connectivity/overall_scores.sql#L24

CloudNiner: Good catch, I'll update the `regexp_replace` functions

PR #215, opened 2017-03-24
Add API view to cancel AnalysisJobs

### LINKED ISSUE #213: Add /api/analysis_job/:job_id/cancel/ endpoint
Should call AnalysisJob.cancel()

Ensure app permissions work correctly

CloudNiner:
## Overview

Adds API view to cancel in progress AnalysisJobs

### Demo

screen shot 2017-03-24 at 15 21 40

### Notes

I chose to do no error handling of exceptions fired by the AnalysisJob.cancel method. Theoretically, we could handle errors that happen within the boto calls, but the only HTTP status to return that _maybe_ makes more sense than the default 500 error is 503. And I'm not sure its worth it at this point to account for that.

## Testing Instructions

Get a token from http://localhost:9200/#/users/edit/:userid/ by clicking the 'refresh token' button.

Trigger a new analysis job via API views, curl, postman, etc. Note the new job's ID.

POST http://localhost:9200/api/analysis_jobs/:jobid/cancel/ via curl, postman, etc. The response should be the analysis job object, with `status: 'CANCELLED'`.

Closes #213

### DIFF
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -266,12 +266,19 @@ class AnalysisJob(PFBModel):
         if not reason:
             reason = 'AnalysisJob terminated by user at {}'.format(datetime.utcnow())
 
-        logger.info('Cancelling job: {}'.format(self))
-        if self.status in self.Status.ACTIVE_STATUSES and self.batch_job_id is not None:
-            client = boto3.client('batch')
-            client.terminate_job(jobId=self.batch_job_id, reason=reason)
-        if self.status != self.Status.CANCELLED:
-            AnalysisJobStatusUpdate.objects.create(job=self, status=self.Status.CANCELLED, step='')
+        if self.status in self.Status.ACTIVE_STATUSES:
+            logger.info('Cancelling job: {}'.format(self))
+            old_status = self.status
+            self.update_status(self.Status.CANCELLED)
+            if self.batch_job_id is not None:
+                try:
+                    client = boto3.client('batch')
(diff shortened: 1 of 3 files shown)


CloudNiner:
I did try cancelling a batch job, and it worked as expected. You make a good point though, as its not a situation that's accounted for, since terminate_job returns immediately. I'll look at this again, a fix should be relatively straightforward.

CloudNiner:
Updated with a second commit to attempt to better address your comment. See commit log for detailed writeup.

src/django/pfb_analysis/models.py
```diff
@@ -329,14 +337,25 @@ def create_environment(**kwargs):
...
-            AnalysisJobStatusUpdate.objects.create(job=self,
-                                                   status=self.Status.QUEUED,
-                                                   step=self.Status.QUEUED)
+            self.update_status(self.status.QUEUED)
         except (botocore.exceptions.BotoCoreError, KeyError):
             logger.exception('Error starting AnalysisJob {}'.format(self.uuid))
 
+    def update_status(self, status, step='', message=''):
+        if self.status != self.Status.CANCELLED:
+            AnalysisJobStatusUpdate.objects.create(job=self,
```
KlaasH: The same note applies the the way I originally wrote the status updating command, but it occurs to me now that this could be written as `self.status_updates.create(status=status, step=step, message=message)`, which is shorter and seems slightly more correct.

CloudNiner: True. will update to this then merge.

PR #217, opened 2017-03-27
Add API view for AnalysisJob results

### LINKED ISSUE #211: Add `/api/analysis_jobs/:job_id/results/` endpoint
This endpoint should return:
- A dict of the overall scores
- A list of urls to download the destinations json files at
- Download link to neighborhood ways
- Download link to neighborhood_census_blocks
- Number of census blocks in the analysis

Depends on #157 #183 #185

CloudNiner:
## Overview

Adds an API view to give the user access to all available results from a single AnalysisJob.

Accessible at `/api/analysis_jobs/:jobId/results/`. If AnalysisJob.status != COMPLETE, then the endpoint returns 404 Not Found.

### Demo

screen shot 2017-03-27 at 09 37 46
screen shot 2017-03-27 at 09 38 13

## Notes

Will wait to merge until I can add the number of census blocks in the analysis to this endpoint, which is being added separately in #183 

## Testing Instructions

Run a local job to completion via the AnalysisJob.run() method. Then query the results endpoint mentioned above, and you should be able to view the overall scores and download the related results files.

Closes #211

### DIFF
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -236,6 +236,19 @@ class AnalysisJob(PFBModel):
         return '{}--{}--{}'.format(definition_name[:30], revision, str(self.uuid)[:8])
 
     @property
+    def census_blocks_url(self):
+        return self._s3_url_for_result_resource('neighborhood_census_blocks.zip')
+
+    @property
+    def destinations_urls(self):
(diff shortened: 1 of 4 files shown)


CloudNiner:
Updated with `census_block_count`, ready for review.

Also made census_block_count a read-only field in the serializer as followup to #183.

PR #218, opened 2017-03-27
Save census block count to AnalysisJob

### LINKED ISSUE #183: Analysis scripts should report number of blocks back to AnalysisJob

KlaasH:
## Overview

Adds a 'set_job_attr' management command and uses it to save the census
block count onto a job during import.

### Demo

image

### Notes

It prints a warning but doesn't raise an except for missing or invalid job, which is preferable in development where jobs run directly don't have valid job IDs and might make sense in production as well.  Other failures result in exceptions.

## Testing Instructions

Check out, run migrations, and run a job.  The census block count attribute should be populated as soon as that part of the import process is done.

Resolves #183.


src/django/pfb_analysis/models.py
```diff
@@ -197,6 +197,7 @@ class Status(object):
                                                 'e.g. http://a.com/foo.osm or ' +
                                                 'http://a.com/foo.osm.bz2')
     overall_scores = JSONField(db_index=True, default=dict)
+    census_block_count = models.IntegerField(blank=True, null=True)
```
CloudNiner: PositiveIntegerField is likely preferable here, but shouldn't be a big deal either way.

https://docs.djangoproject.com/en/1.10/ref/models/fields/#positiveintegerfield

PR #222, opened 2017-03-28
Feature/overall scores update

spencerrecneps:
## Overview

I apologize for not finalizing this sooner. This PR establishes the final schema for the scoring outputs. I've moved the old overall_scores logic to an intermediate output (score_inputs) and created a new overall_scores table with the scores that should be used on the website.

### Demo

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

 * Run run_connectivity.sh
 * View output in overall_scores table

Connects #XXX


spencerrecneps:
Just noticed I accidentally committed some commenting out of run_connectivity that I did during testing. Will revert and re-commit. Standby.

spencerrecneps:
OK. Should be fixed. Sorry for the noise.

CloudNiner:
Thanks for sending this. We'll need to shuffle things around, still hope to get you a login to the staging account sometime tomorrow so you can start testing with these changes applied. I'll review this a bit later today.

In the meantime, can you update https://github.com/azavea/pfb-network-connectivity/issues/157 with the final list of destinations tables we'll need to have available.

src/analysis/scripts/export_connectivity.sh
```diff
@@ -149,10 +153,13 @@ then
...
         # ec_export_table_shp "${OUTPUT_DIR}" "neighborhood_connected_census_blocks"
 
+        # Export neighborhood_score_inputs as CSV
+        ec_export_table_csv "${OUTPUT_DIR}" "neighborhood_score_inputs"
+
         # Export neighborhood_overall_scores as CSV
         ec_export_table_csv "${OUTPUT_DIR}" "neighborhood_overall_scores"
         # Send overall_scores to Django app
-        update_overall_scores "${OUTPUT_DIR}/neighborhood_overall_scores.csv"
+        update_overall_scores "${OUTPUT_DIR}/neighborhood_score_inputs.csv"
```
CloudNiner: In your PR comment you stated that the new overall_scores is the one that should be displayed on the site. Do we need to make this intermediate table available and if so for what purpose?

spencerrecneps: neighborhood_score_inputs would only need to be available as an export. It doesn't need to be referenced by the website.

src/analysis/connectivity/overall_scores.sql
```diff
@@ -1,275 +1,90 @@
...
+--   and converts to scores for each of the
+--   subcategories. Then, combines the
+--   subcategory scores into an overall category
+--   score. Finally, combines category scores into
+--   a single master score for the entire
+--   neighborhood.
 ----------------------------------------
 DROP TABLE IF EXISTS generated.neighborhood_overall_scores;
 
 CREATE TABLE generated.neighborhood_overall_scores (
```
spencerrecneps: I can make this change. Will commit something ASAP.

src/analysis/connectivity/score_inputs.sql
```diff
@@ -0,0 +1,2636 @@
...
+-- population
+-------------------------------------
+-- median pop access ratio
+INSERT INTO generated.neighborhood_score_inputs (
+    category, score_name, score, notes, human_explanation, use_pop
+)
+SELECT  'People',
+        'Median ratio of access to population',
+        quantile(pop_ratio,0.5),
+        regexp_replace('Ratio of population accessible by low stress
```
spencerrecneps: This raises the question of where the website is going to be pulling descriptions of each score from. Would it make the most sense to include that as a column in this table (as we were doing previously?) Or is that best handled separate from the tabular data?

In other words, should I be adding a "human readable" column that contains text ready to published on the website?

src/analysis/connectivity/score_inputs.sql
```diff
@@ -0,0 +1,2636 @@
...
+            to population accessible overall, expressed as
+            the median of all census blocks in the
+            neighborhood','\n\s+',' '),
```
spencerrecneps: Thanks for catching that.

spencerrecneps:
Just wanted to provide a little more background on why I split things into the intermediate and final tables. The team has gone back and forth on the strengths/weaknesses of various ways to aggregate the census block data across a whole neighborhood. The intermediate table is basically a collection of various options for aggregating. The final table simply takes the aggregation method that we selected for each category. So part of the final table is basically a subset of the intermediate table.

The final difference involves taking the aggregated scores and then normalizing them to a scale of 0-100, and finally rolling up all the scores into a single master score for the whole neighborhood (this also on a scale of 0-100).

src/analysis/connectivity/score_inputs.sql
```diff
@@ -0,0 +1,2636 @@
...
+    use_grocery BOOLEAN,
+    use_social_svcs BOOLEAN,
+    use_parks BOOLEAN,
+    use_trails BOOLEAN,
+    use_comm_ctrs BOOLEAN,
+    use_transit BOOLEAN
+);
+
+
+-------------------------------------
```
spencerrecneps: I can see the logic in that. I've generally tried to followed a pattern of naming each script for the table which it creates (which as you note fails for the access_*.sql scripts).

The one concern I have with this approach is that the final scoring requires that we combine scores from several related categories. For instance, the core services score is a composite of doctors, dentists, hospitals, etc. If we break out this scoring into each access_*.sql file, we'll still end up having to do some calculations within overall_scores.sql.

I'll mull this as I make these other changes.

src/analysis/scripts/export_connectivity.sh
CloudNiner: Ok. This is fine as is, we'll update this as necessary once this PR is merged.

src/analysis/connectivity/score_inputs.sql
CloudNiner: That makes sense, and I saw some examples of that while perusing the PR. Just wanted to provide an alternative, as coming back to this massive multiple-thousands of lines file for further development could be frustrating in the future.

src/analysis/connectivity/score_inputs.sql
CloudNiner: Yes, great question. We're leaning towards maintaining the list of available scores, categories, and human-readable strings in the main database. This is a bit more flexible and extensible, and we should be able to (at least partially) guard against the metadata for each score getting out of sync with the list of scores being calculated by the analysis.

This would mean that all of the associated metadata in the intermediate table (old overall_scores table) could go away, and the new overall_scores table can just contain the score name key and value. 

It would be helpful if the score name set via the analysis was a short, unique, 'snake case' key that can double as the lookup key in the main database to retrieve the associated score metadata (category, label, long description, etc).  

If you're willing to make this change, we can communicate later this week on how to get the score metadata prepared.

spencerrecneps: Sounds good to me. I'll make the change.

CloudNiner:
@spencerrecneps shoot us a quick comment on this PR once you've addressed all comments/are satisfied with changes and I'll take another look.

spencerrecneps:
I think this should do it for changes to address your previous comments. I've used the category field in the overall_scores table as the snake_case identifier that you can rely on for matching with the main database.

src/analysis/connectivity/overall_scores.sql
```diff
@@ -23,7 +23,7 @@ CREATE TABLE generated.neighborhood_overall_scores (
 INSERT INTO generated.neighborhood_overall_scores (
     category, score_original, score_normalized, human_explanation
 )
-SELECT  'People',
+SELECT  'people',
```
CloudNiner: Is the category field now misnamed? Should it just be a `score_identifier` or `score_name` instead? It seems to me that a category would be something like `opportunity`, not `opportunity_employment` which is a specific score or metric?

Also, since we'll be saving category, human_explanation, and any other metadata separately, the human_explanation field here can be removed, and also be removed from the neighborhood_score_inputs table. I'm ok with leaving them in for now if we want to wait to remove until we setup the replacement source for this information.

spencerrecneps: Changed to score_id. I'd like to leave the explanation fields in there for now. We may actually want to leave them in there for good as a reference for users who download the scores.

CloudNiner: 👍

CloudNiner: Don't see a new commit for this change? Forget to push?

spencerrecneps: Indeed. Sorry.

CloudNiner:
I pulled this branch down and ran it locally, hit this error in score_inputs.sql:
```
ERROR:  relation "neighborhood_transit" does not exist at character 830
STATEMENT:  INSERT INTO generated.neighborhood_score_inputs (
            category, score_name, score, notes, human_explanation
        )
        SELECT  'Opportunity',
                'Average transit bike shed access ratio',
                CASE    WHEN SUM(pop_high_stress) = 0 THEN 0
                        ELSE SUM(pop_low_stress)::FLOAT / SUM(pop_high_stress)
                        END,
                regexp_replace('Ratio of population with low stress access
                    compared to total population within the bike shed distance
                    of transit stations in the neighborhood expressed as an average of
                    all transit stations in the neighborhood','\n\s+',' ','g'),
                regexp_replace('On average, transit stations in the neighborhood are
                    connected by the low stress access to this percentage people
                    within biking distance.','\n\s+',' ','g')
        FROM    neighborhood_transit
        WHERE   EXISTS (
                    SELECT  1
                    FROM    neighborhood_boundary AS b
                    WHERE   ST_Intersects(neighborhood_transit.geom_pt,b.geom)
                );
psql:../connectivity/score_inputs.sql:2567: ERROR:  relation "neighborhood_transit" does not exist
LINE 16: FROM    neighborhood_transit
                 ^
ERROR:  relation "neighborhood_transit" does not exist at character 856
STATEMENT:  INSERT INTO generated.neighborhood_score_inputs (
            category, score_name, score, notes, human_explanation
        )
        SELECT  'Opportunity',
                'Median transit population shed ratio',
                quantile(pop_ratio,0.5),
                regexp_replace('Ratio of population with low stress access to transit stations
                    in the neighborhood to total population within the bike shed
                    of each transit stations expressed as a median of all
```

CloudNiner:
If you're ok with fixing that in a separate PR, I'll merge this so I can get started on the associated changes we need to make to the main application to support the schema changes made here.

CloudNiner:
I just pushed a commit to this branch that temporarily disables the load_overall_scores export, since that was causing issues and will need to be fixed separately.

spencerrecneps:
Ah yes. Forgot to warn you of the transit errors. They should be harmless, since it just leaves a null value in the transit score fields. It's expecting a transit table, but we still haven't enabled the GTFS input. We should be creating an empty transit table in any case for cities who don't upload GTFS.

I'll add in a separate PR once this is merged.

Issue #143, opened 2017-03-09
Architect analysis tile generation


CloudNiner:
How do we do this? Separate job? Same job? What additional software is necessary?

Need to output tiles for some number of sensible zoom levels depending on boundary size (2-5 levels) each time we run an analysis. Output uses the geom from neighborhood_ways table, polylines colored by one of the columns in the table.

CloudNiner:
Consider asking other folks with tiling experience at Azavea for suggestions on how to handle this.

e.g. Geotrellis or people who worked on Tiler Perry

Task inputs:
- shapefile with linestring geometries, with the fields to style by as attributes

Task outputs:
- Static image tiles stored to the job's `results` S3 bucket, one tile layer, only saving a subset of sensible zoom levels for that particular location (base the zoom level range on boundary size?)

CloudNiner:
@spencerrecneps The analysis database table we need to export for this is `neighborhood_ways` correct? Which columns should we be styling/displaying on that table? What are the possible values, and which values are 'good' and 'bad'?

spencerrecneps:
@CloudNiner sorry I missed the mention above. Looks like you got this sorted out? There are actually two stress values that should be shown: ft_seg_stress and tf_seg_stress. The way I generally show these is to draw two lines (one for each side of the road) and offset them slightly.

CloudNiner:
Thanks. Not quite what we're doing, added #261 to address it.

PR #225, opened 2017-03-29
Generate tiles for neighborhood_ways stress ratings

KlaasH:
## Overview

Adds a container, called `tilemaker`, that takes a job ID and an AWS bucket and looks in the corresponding results folder for a `neighborhood_ways.zip`, which it downloads and converts to raster tiles then uploads back to a `tiles` directory in the results folder.

### Demo

image

### Notes

- I renamed all the `BATCH_JOB` variables, methods, etc. to `BATCH_ANALYSIS_JOB`, since I wanted to make another and call it `BATCH_TILEMAKER_JOB`.
- Just like the analysis job, it can't be submitted to AWS Batch from development, but it prints the command you can run to do it locally.
- ~Not sure exactly how we'll want to schedule/trigger this, so I left that part out for now.  Could be a bit of logic in `update_status` that calls `generate_tiles` when the status changes to `COMPLETE`.~
- As suggested in line comments, `AnalysisJob.run()` calls `generate_tiles`, which schedules the job with a `dependsOn` parameter so that it won't run until the analysis job is done.
- ~Calling this WIP because I haven't tried deploying it to staging yet.~

## Testing Instructions

- Change your `group_vars/all` based on the changes to `all.example`
- Run `./scripts/setup` from the host machine to do the whole provisioning and rebuilding.
- Run an analysis job or find a finished one whose results were successfully uploaded.  Get its ID and run:
```
AWS_STORAGE_BUCKET_NAME='<your_username>-pfb-storage-us-east-1' \
  PFB_JOB_ID='<job_id>' PFB_S3_RESULTS_PATH='results/<job_id>' \
  PFB_S3_TILES_PATH='results/<job_id>/tiles' docker-compose run tilemaker
```
If you don't want to copy/paste the job ID that many times, you can also make a job in the angular site--it will print the analysis command and the tiling command to the log, so you can run them in succession.

Connects #143.


src/tilemaker/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+    exit 1
+fi
+
+TL_SHAPEFILE_NAME="${TL_SHAPEFILE_NAME:-neighborhood_ways}"
+TL_MIN_ZOOM="${TL_MIN_ZOOM:-8}"
+TL_MAX_ZOOM="${TL_MAX_ZOOM:-17}"
+
+PFB_TEMPDIR=`mktemp -d`
+
+TL_AWS_RESULTS_PATH="s3://${AWS_STORAGE_BUCKET_NAME}/results/${PFB_JOB_ID}"
```
CloudNiner: It might be better if this was passed in as part of the ENV? The AnalysisJob already manages the path a job's files exist at, and keeping AnalysisJob as the single source of truth for that is probably good, rather than having it defined here in a second spot.

src/tilemaker/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+mkdir -p /data
+ogr2ogr -t_srs EPSG:4326 -f GeoJSON "/data/${TL_SHAPEFILE_NAME}.json" \
+    "${PFB_TEMPDIR}/${TL_SHAPEFILE_NAME}.shp"
+
+# Get bounds
+TL_BOUNDS=$(geojson-extent < "/data/${TL_SHAPEFILE_NAME}.json" spaces)
+
+# Make tiles and upload them to S3
+tl copy -z "${TL_MIN_ZOOM}" -Z "${TL_MAX_ZOOM}" -b "${TL_BOUNDS}" \
+    "mapnik:///opt/tl-export/styles/${TL_SHAPEFILE_NAME}_style.xml" \
```
CloudNiner: Static xml file seems good for now. I don't recall anything in the spec about individual cities needing to be able to tweak their analysis's styling.

src/django/pfb_analysis/models.py
```diff
@@ -348,17 +361,53 @@ def create_environment(**kwargs):
...
+            jobQueue=settings.PFB_AWS_BATCH_ANALYSIS_JOB_QUEUE_NAME,
+            containerOverrides=container_overrides)
         try:
             self.batch_job_id = response['jobId']
             self.save()
             self.update_status(self.status.QUEUED)
         except (botocore.exceptions.BotoCoreError, KeyError):
             logger.exception('Error starting AnalysisJob {}'.format(self.uuid))
 
+    def generate_tiles(self):
```
KlaasH: I made EXPORTED and TILING statuses, the former being the new final status of the analysis container.  The tiling container sets TILING when it starts and COMPLETE when it's done.

src/django/pfb_network_connectivity/settings.py
```diff
@@ -271,27 +271,41 @@
...
+# Same setup for tilemaker jobs, but without the exception-throwing
+PFB_AWS_BATCH_TILEMAKER_JOB_QUEUE_NAME = os.getenv('PFB_AWS_BATCH_TILEMAKER_JOB_QUEUE_NAME')
+PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION = os.getenv('PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION')
+PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME = os.getenv('PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME')
+if PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME and not PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION:
+    try:
+        revision = get_latest_job_definition(PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME)['revision']
+    except NoActiveJobDefinitionRevision:
+        pass
+    PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION = '{}:{}'.format(PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME,
```
KlaasH: Yeah, you're right.  No reason for this to act any differently from the analysis job setup.  I changed it to match.

src/tilemaker/Dockerfile
```diff
@@ -0,0 +1,79 @@
+FROM buildpack-deps:xenial
+
+##### Install Node
+# Copied from https://github.com/nodejs/docker-node/blob/master/6.10/Dockerfile because we
+# need G++ 5 to successfully compile Mapnik but Node's standard Debian containers don't have it.
```
KlaasH: I figured since the Node containers use [buildpack-deps](https://hub.docker.com/_/buildpack-deps/) `jessie` images it would be easiest to use the same thing but from a newer base image.  There might be alternatives, like upgrading g++ on the Node image or figuring out a way to get Mapnik installed from binaries such that the packages that use it won't try to compile it.  But this seemed easier than either of those, and made the crashes stop, so I went with this.

src/tilemaker/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+# Install tl and other utilities
+RUN npm install -g --save \
+    tl mbtiles mapnik \
+    tilelive tilelive-file tilelive-http tilelive-mapbox tilelive-mapnik \
+    tilelive-s3 tilelive-modules \
+    tilelive-tmsource tilelive-tmstyle tilelive-utfgrid tilelive-vector tilelive-carto \
+    tilejson geojson-extent
+
+# Customized version of gejson-extent until upstream gets changed
+RUN npm install -g git+https://github.com/KlaasH/geojson-extent.git#4f405faca
```
KlaasH: Gone now.  Turned out this choked on the NYC file anyway, so I changed to using `ogrinfo`.

src/tilemaker/entrypoint.sh
KlaasH: This was hard-coded in the analysis export script, too, but yeah, I pulled it out into AnalysisJob. Might be better in settings, but then again another layer might be overkill.

src/tilemaker/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+##### Include the django container so we can send status updates
+# Copy django requirements (will copy the rest later, so that any change to
+# the code doesn't trigger re-running the whole provisioning step)
+COPY ./django/requirements.txt /opt/pfb/django/requirements.txt
+
+RUN apt-get install -y \
+    python-dev \
+    python-pip
+
+# Copy django requirements (will copy the rest later, so that any change to
```
CloudNiner: Looks like a copy/paste error, same code duplicated on lines 77-79 above? Should one set be removed?

CloudNiner:
Should also double-check the build issues before merging.

src/tilemaker/Dockerfile
KlaasH: Whoops, yup, made a mess of that.  Was copying pieces from the analysis dockerfile but rearranging them to make the "include Django container" part self-contained.

PR #232, opened 2017-03-30
Angular AnalysisJob detail view

### LINKED ISSUE #208: Add download links to AnalysisJob files in admin UI
Probably best on a separate AnalysisJob "detail" view. 

For now can be an unstyled list of links.

Files to make available (via S3 signed URLs) when job is complete are at:
```
PFB_S3_STORAGE_BUCKET/results/:analysis_job_uuid/*
```

CloudNiner:
## Overview

Adds a functional (but not super pretty) AnalysisJob detail view to the admin UI. 

User can see all properties of the AnalysisJob, can refresh the view to check for updates, and can cancel an in progress job. 

### Demo

screen shot 2017-03-30 at 15 01 41
screen shot 2017-03-30 at 15 01 50

### Notes

Didn't get fancy with the cancel button. If a job is cancellable it'll cancel, otherwise silently do nothing. I think it's fine for now.

Same with dressing up the rows, or trying to neatly format text. Don't see value in it yet, easy to clean up later.

I included links in the data section to all files requested for download in #157, including a zipped connected_census_blocks.csv

## Testing Instructions

Start AnalysisJob from UI in development. Go to detail page. Cancel it, to prove that works. Start another job. Run it via console. Go to detail page for that job and click refresh while job is in progress to see it update. When job is complete the results will populate after clicking refresh.

Closes #208

### DIFF
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -249,6 +249,10 @@ class AnalysisJob(PFBModel):
         }
 
     @property
+    def overall_scores_url(self):
+        return self._s3_url_for_result_resource('neighborhood_overall_scores.csv')
+
+    @property
     def logs_url(self):
         url = ('https://console.aws.amazon.com/cloudwatch/home?region={aws_region}' +
                '#logStream:group=/aws/batch/job;prefix={batch_job_name}/{batch_job_id}' +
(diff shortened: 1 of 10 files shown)


src/django/pfb_analysis/models.py
```diff
@@ -240,6 +240,10 @@ def census_blocks_url(self):
         return self._s3_url_for_result_resource('neighborhood_census_blocks.zip')
 
     @property
+    def connected_census_blocks_url(self):
+        return self._s3_url_for_result_resource('neighborhood_connected_census_blocks.csv.zip')
```
KlaasH: All this repetition of the list of exports is begging to be consolidated into a config object in settings.  Then these model and view pieces could loop over it and the analysis could get it as (yet another) environment variable.  Or something along those lines.  Not on this PR, though.  We should open an issue.

PR #234, opened 2017-03-31
Feature/awf/django score model

### LINKED ISSUE #224: Add Django model for analysis score metadata
Should contain at least the following properties for each score:
- slug (used to match the score name generated by the analysis)
- category
- short label
- human readable (long) description

Probably easiest to just get the full list of scores and associated metadata from spencer and load this list via fixture, keeping the score metadata version controlled that way.

When displaying the list of overall_scores in the AnalysisJob model, we'll want to set diff the available list of scores against the saved ones in the AnalysisJob model instance, and raise a warning if there's any inconsistencies.

CloudNiner:
## Overview

This is the other followup to #222 in which it was decided to move the score metadata into the Django application so it lives in a single place. This PR adds the model and initial data fixture to load the initial set of analysis score metadata.

### Notes

Did not add additional checks or score metadata, since its not critically important. We can do so once we have a better idea of how this will be displayed to the application user.

## Testing Instructions

Run `./scripts/django-manage loaddata analysis-score-metadata` and ensure it runs successfully.

Closes #224


src/django/pfb_analysis/fixtures/analysis-score-metadata.json
```diff
@@ -0,0 +1,182 @@
+[
+  {
+    "model": "pfb_analysis.AnalysisScoreMetadata",
```
CloudNiner: @spencerrecneps this is the file you'll need to update once you're ready to finalize the score metadata. I just filled in most fields with placeholders for now.

Opened #235 as a reminder.

CloudNiner:
So scores can go out of sync with fixtures as well, because each time you load a fixture it truncates the table before reloading, but an append only strategy works fine in either case. In which case the primary benefit of the fixture is that its easier to give an external client a JSON file than it is to point them a source code file. 

Updates to the score metadata would require a deploy+mgmt command (no migration) in the case of fixtures, or just a deploy in the case of static object.

I don't really feel strongly about either strategy, neither has any compellingly strong advantages/disadvantages.

CloudNiner:
Yes, I will add loaddata to scripts/update. Great point.

I'll delete the helper method for now, agree its not super useful until we figure out access patterns.

PR #240, opened 2017-03-31
Move existing views to /admin

### LINKED ISSUE #193: Move existing angular admin views to `/admin`
We want to reserve the root view for the public site

flibbertigibbet:
Move existing Angular views from site root to /admin.
Closes #193.

### Demo

image

## Testing Instructions

 * rebuild angularjs container
 * views under http://localhost:9301 should all route to http://localhost:9301/admin

### DIFF
--- a/src/angularjs/src/app/index.route.js
+++ b/src/angularjs/src/app/index.route.js
@@ -9,32 +9,32 @@
     function routerConfig($stateProvider, $urlRouterProvider) {
         $stateProvider
             .state('login', {
-                url: '/login/',
+                url: '/admin/login/',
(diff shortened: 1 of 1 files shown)


src/angularjs/src/app/index.route.js
```diff
@@ -9,32 +9,32 @@
     function routerConfig($stateProvider, $urlRouterProvider) {
         $stateProvider
             .state('login', {
-                url: '/login/',
+                url: '/admin/login/',
```
CloudNiner: Since login would log the user into the entire application, maybe this should continue to live at `/login`

PR #265, opened 2017-04-10
Fix join syntax in results URLs method

KlaasH:
I converted results path from a hard-coded string in a few places to a
method on AnalysisJob, but got the syntax wrong when I changed
`_s3_url_for_result_resource` for the new setup.


KlaasH:
Trivial, merging.

PR #267, opened 2017-04-11
Add score and latest-updated sorting to AnalysisJob endpoint

### LINKED ISSUE #242: Add sorting options to AnalysisJob API endpoint
- [ ] Last updated (job completion datetime)
- [ ] Highest Rated
- [ ] Lowest Rated
- [ ] Alphabetical

### LINKED ISSUE #260: Add latest=true filter to analysis job list endpoint
It should only return the most recently completed AnalysisJob for each Neighborhood

KlaasH:
## Overview

Adds "overall score" and "latest updated" sorting and filtering to the AnalysisJobs endpoint.

### Demo

- Latest updated. Sorting by `created_at` already worked, but with the addition of `latest=True`, it now works as a "neighborhoods by last analysis run" endpoint: http://localhost:9200/api/analysis_jobs/?latest=True&ordering=-created_at
- Highest rated: http://localhost:9200/api/analysis_jobs/?latest=True&ordering=-overall_score
- Lowest rated: http://localhost:9200/api/analysis_jobs/?latest=True&ordering=overall_score
- Alphabetical by neighborhood: http://localhost:9200/api/analysis_jobs/?latest=True&ordering=neighborhood__label (note the double-underscore.  the `neighborhood_label` field in the results is a MethodField added by the serializer, so not sortable)

### Notes

~Issue #242 calls for "Last updated (job completion datetime)".  This implements it as job creation datetime.  I just started writing up a case for doing it by job creation, but then realized that would mean the scores for a neighborhood disappear as soon as a new job is started for it, which would not be great.  I made a run at pulling AnalysisJobStatusUpdate information into the query and didn't get there, but I'll have another go in issue #260.~

I pushed a new version that switches to doing the `latest` filtering in code and behaves thus:
- If there are successful jobs for a neighborhood, include only the one that finished most recently
- If there are none, include the one with the most recent status update
- If there are no status updates for any job for a neighborhood (i.e. only CREATED jobs), return the most recently modified job

Note also, if this gets applied after a status filter (which seems to be the case. I haven't found documentation, but it seems like it gets applied second regardless of parameter ordering), it will be based on the set of jobs it gets, so it'll follow the fallback logic and return the most recently updated of those jobs, even if there are other jobs for the same neighborhoods with different statuses that are newer.

## Testing Instructions

See the Demo section for example queries.  The new ordering options and the `latest=True` filter should work fine in conjunction with the previously-existing filters (e.g. by status) and ordering options.

Resolves #242, resolves #260.

### DIFF
--- a/src/django/pfb_analysis/filters.py
+++ b/src/django/pfb_analysis/filters.py
@@ -24,6 +24,7 @@ class AnalysisJobFilterSet(filters.FilterSet):
         if value:
             matches = [m.pk for m in queryset.all() if m.status == value]
             queryset = queryset.filter(pk__in=matches)
+
         return queryset
 
     def filter_latest(self, queryset, name, value):
--- a/src/django/pfb_analysis/filters.py
+++ b/src/django/pfb_analysis/filters.py
@@ -1,0 +1,40 @@
+import logging
+
+from rest_framework import filters
+import django_filters
+from django.db.models import Case, When, Q, Max, Value, BooleanField
+
+from .models import AnalysisJob
+
+
+logger = logging.getLogger(__name__)
+
+
+class AnalysisJobFilterSet(filters.FilterSet):
+    """ Filters for AnalysisJob:
+      - status, to handle job status being defined as a property and not as database field
+      - latest, to return only the most recent analysis job for each neighborhood
+    """
+
+    status = django_filters.ChoiceFilter(choices=AnalysisJob.Status.CHOICES,
+                                         method='filter_status')
+    latest = django_filters.BooleanFilter(method='filter_latest')
+
+    def filter_status(self, queryset, name, value):
+        if value:
+            matches = [m.pk for m in queryset.all() if m.status == value]
+            queryset = queryset.filter(pk__in=matches)
+        return queryset
+
+    def filter_latest(self, queryset, name, value):
+        if type(value) is bool:
+            queryset = queryset.annotate(is_latest=Case(
+                When(Q(created_at=Max('neighborhood__analysis_jobs__created_at')),
+                     then=Value(True)),
+                default=Value(False),
+                output_field=BooleanField())).filter(is_latest=value)
+        return queryset
+
+    class Meta:
+        model = AnalysisJob
+        fields = ['neighborhood', 'batch', 'status', 'latest']
--- a/src/django/pfb_analysis/filters.py
+++ b/src/django/pfb_analysis/filters.py
@@ -2,9 +2,8 @@ import logging
 
 from rest_framework import filters
 import django_filters
-from django.db.models import Case, When, Q, Max, Value, BooleanField
 
-from .models import AnalysisJob
+from .models import AnalysisJob, AnalysisJobStatusUpdate, Neighborhood
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


src/django/pfb_analysis/filters.py
```diff
@@ -0,0 +1,62 @@
...
+        return queryset
+
+    def filter_latest(self, queryset, name, value):
+        """ Filters down to the latest successful analysis for each neighborhood, but falls back
+        to the latest modified if there are no successful analysis jobs for a neighborhood.
+
+        This means that if it's applied on top of a status filter, it follows the fallback
+        path and returns the latest job with the given status.
+
+        Runs tons of queries, but the query expression required to do this would be gnarly.
```
CloudNiner: Meh. Think I agree that its probably fine for now, but worth investigating if it gets nasty later.

src/django/pfb_analysis/views.py
```diff
@@ -12,24 +12,26 @@
...
+from .serializers import AnalysisJobSerializer, NeighborhoodSerializer
+from .filters import AnalysisJobFilterSet
+from .functions import ObjectAtPath
+
 
 class AnalysisJobViewSet(ModelViewSet):
     """For listing or retrieving analysis jobs."""
 
-    queryset = AnalysisJob.objects.all()
+    queryset = AnalysisJob.objects.all().annotate(
```
CloudNiner: Presumably this annotation is so that the sorting works, right?

This annotation is presumably useful elsewhere. Consider moving the logic to a custom AnalysisJobManager in models.py.

src/django/pfb_analysis/functions.py
```diff
@@ -0,0 +1,14 @@
+from django.db.models.expressions import Func
+
+
+class ObjectAtPath(Func):
```
CloudNiner: Huh. I'm a bit surprised this doesn't exist already.

src/django/pfb_analysis/views.py
```diff
@@ -12,24 +12,26 @@
...
+        overall_score=ObjectAtPath('overall_scores', ('overall_score', 'score_normalized')))
```
KlaasH: Yeah, I agree, but https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#jsonfield is focused on filtering and I don't see anything that would fit the bill at https://docs.djangoproject.com/en/1.10/ref/models/expressions/

The double-underscore syntax has been extended to traversing JSONField values in some contexts, e.g.
```
AnalysisJob.objects.filter(overall_scores__overall_score__score_normalized__gt=1)
```
works, but
```
AnalysisJob.objects.annotate(overall_score=F('overall_scores__overall_score__score_normalized'))
```
says "FieldError: Cannot resolve keyword 'overall_score' into field. Join on 'overall_scores' not permitted."

src/django/pfb_analysis/filters.py
KlaasH: Definitely would be up for another run at it at some point, since the same logic would probably enable moving the status filter into the query as well, but yeah, it seems reasonable to settle for this for the moment.

src/django/pfb_analysis/views.py
KlaasH: Good point.  It seemed weird here, but I wasn't sure where it should go.  Will move.

src/django/pfb_analysis/functions.py
KlaasH: Yeah, I agree, but https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#jsonfield is focused on filtering and I don't see anything that would fit the bill at https://docs.djangoproject.com/en/1.10/ref/models/expressions/

The double-underscore syntax has been extended to traversing JSONField values in some contexts, e.g.
```
AnalysisJob.objects.filter(overall_scores__overall_score__score_normalized__gt=1)
```
works, but
```
AnalysisJob.objects.annotate(overall_score=F('overall_scores__overall_score__score_normalized'))
```
says "FieldError: Cannot resolve keyword 'overall_score' into field. Join on 'overall_scores' not permitted."

CloudNiner: I have some hunches on why that F expression is failing, but either way I wouldn't expect them to work for jsonb fields since the SQL query syntax is different and jsonb is postgresql specific. Either way I think the solution presented here is ok for now.

PR #268, opened 2017-04-11
Codify SES Send Email permissions in Terraform

### LINKED ISSUE #238: Add SES:SendRawEmail permission to EC2 container instance profile
Was temporarily added to the EC2 StagingContainerInstanceProfile via the AWS console

CloudNiner:
## Overview

Email permissions are needed by the EC2 instance to properly send user emails. These were quickfixed earlier on by adding the necessary permissions in the console. Now the same permissions are codified in Terraform

### Demo

screen shot 2017-04-11 at 15 41 11

### Notes

I _believe_ I made these changes in an acceptable way, as an inline policy for both ECS/EC2. Permission seems to be necessary on both. 

## Testing Instructions

Can checkout branch and run `./scripts/infra plan` to verify no permissions changes in the plan, the changes in the PR are live in the staging environment.

Closes  #238

### DIFF
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -27,12 +27,15 @@ data "aws_iam_policy_document" "container_instance_ecs_assume_role" {
   }
 }
 
-data "aws_iam_policy_document" "container_instance_ses_send_email" {
+data "aws_iam_policy_document" "ses_send_email" {
   statement {
     effect = "Allow"
 
     resources = ["*"]
-    actions   = ["ses:SendRawEmail"]
+    actions   = [
+      "ses:SendEmail",
+      "ses:SendRawEmail"
+    ]
   }
 }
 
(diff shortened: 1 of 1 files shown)


CloudNiner:
@azavea/operations can you take a look?

deployment/terraform/iam.tf
```diff
@@ -80,10 +83,10 @@ resource "aws_iam_role_policy_attachment" "ecs_policy" {
   policy_arn = "${var.aws_ecs_service_role_policy_arn}"
 }
 
-resource "aws_iam_role_policy" "ses_send_email" {
-  name   = "SESSendEmail"
+resource "aws_iam_role_policy" "ecs_ses_send_email" {
```
hectcastro: I don't think this one is necessary. This role is assumed by the ECS service to make API calls on our behalf. It shouldn't need to send emails.

deployment/terraform/iam.tf
```diff
@@ -99,6 +102,12 @@ resource "aws_iam_role" "container_instance_ec2" {
   assume_role_policy = "${data.aws_iam_policy_document.container_instance_ec2_assume_role.json}"
 }
 
+resource "aws_iam_role_policy" "ec2_ses_send_email" {
```
hectcastro: This one is definitely necessary because it gets assumed by the EC2 instance running the application.

deployment/terraform/iam.tf
CloudNiner: Tested...correct. Updated PR.

deployment/terraform/iam.tf
```diff
@@ -99,6 +96,12 @@ resource "aws_iam_role" "container_instance_ec2" {
...
+  name = "EC2SESSendEmail"
```
hectcastro: I think that it would be good to inject `var.environment` into this name so that the staging and production environments don't collide when inside the same account.

CloudNiner: Good idea. Done.

PR #275, opened 2017-04-12
Feature/awf/django save neighborhood geom

### LINKED ISSUE #190: Save neighborhood geometry in Django DB on model create

CloudNiner:
## Overview

Bah. This took longer than expected. Due to the abstractions in the Django file API, we can only rely on a very specific subset of File operations when working with a FileField. This means that in order to protect against the underlying storage API changing, we need to read the entire file to disk so we can parse it with Fiona or even something like Django's built-in DataSource API, because those only support Python file objects or a string path to a file on disk.

## Notes

Added a data migration to parse any existing Neighborhoods, catching all errors and printing exceptions to console so that migrations don't fail if there's an error loading up a Neighborhood's shapefile.

## Testing Instructions

1. Run migrations. All of your existing neighborhoods should now have geoms attached.

2. Create a new neighborhood boundary via the API, then inspect it in the django shell, `neighborhood.geom` should be non-null.

Closes #190


src/django/pfb_analysis/models.py
```diff
@@ -101,6 +103,7 @@ def set_boundary_file(self, geom):
                         zip_handle.write(os.path.join(tmpdir, shpfile), shpfile)
             boundary_file = File(open(zip_filename))
             self.boundary_file = boundary_file
+            self.geom = geom
```
CloudNiner: Shortcut by directly setting geom here, so that we don't create a shapefile from the geom to upload, then in save recreate the shapefile locally again just so we can get the geom back from it that we already have.

src/django/pfb_network_connectivity/models.py
```diff
@@ -5,7 +5,7 @@
 from __future__ import unicode_literals
 import uuid
 
-from django.db import models
+from django.contrib.gis.db import models
```
CloudNiner: This seemed easiest, no issues with swapping the base class. Other option was to have two PFBModel base classes, one inheriting from gis model and one from the regular one, but that was way more effort for essentially no gain.

Issue #114, opened 2017-03-01
Parameterize buffer distance in coverage.sql


CloudNiner:
It doesn't look like this file is currently in use anywhere? Can you verify @spencerrecneps ?

spencerrecneps:
Not currently. It's in need of some changes before I pull it into the final results. I'll parameterize appropriately when I make my updates.

CloudNiner:
👍

CloudNiner:
Leaving open for now and assigning to you. We can revisit this issue later.

CloudNiner:
@spencerrecneps coverage.sql is still not used. Can that file be deleted if its not in use and this issue closed?

spencerrecneps:
Yes. There are several files that are obsolete and can be removed. I'll do that and submit a PR.

lydiascarf:
Closing this as it's about a file that doesn't exist anymore

PR #276, opened 2017-04-12
Remove unused files relating to coverage.

spencerrecneps:
## Overview

Removes files that were not being used in the analysis. They had been left over from the early stages of the project when we were testing various measures.


CloudNiner:
Closes #114

PR #277, opened 2017-04-12
Offset segments to show both directions in neighborhood_ways tiles

### LINKED ISSUE #261: Update tile generation to show both ft_seg_stress and tf_seg_stress
By drawing both, each slightly offset to one side of the road or the other. Hopefully this can be done with styling...

KlaasH:
## Overview

Updates the styling for the stress score tiles to
- only rely on the `SEG_STR` scores (formerly it also looked at `INT_STR`)
- display the `FT` and `TF` scores separately by offsetting them in opposite directions.

### Demo

image

### Notes

- I also changed the tile generation command to produce some output, rather than none, as it was doing, or tens of thousands of lines, as it was doing before that.
- The sample above is from a York, PA, analysis, and seems to be missing some roads. I opened the exported neighborhood_ways shapefile in QGIS and confirmed that they're not there.

## Testing Instructions

Rebuild the `tilemaker` container and run it.  You can create a job locally and run the analysis then the tilemaker commands that get printed to the logs, or you could run tile generation for an already-finished analysis, along the lines of
```
AWS_STORAGE_BUCKET_NAME='<username>-pfb-storage-us-east-1' PFB_JOB_ID='<job_id>' PFB_S3_RESULTS_PATH='results/<job_id>' PFB_S3_TILES_PATH='results/<job_id>/tiles' docker-compose run tilemaker
```

Resolves #261.

### DIFF
--- a/src/tilemaker/entrypoint.sh
+++ b/src/tilemaker/entrypoint.sh
@@ -37,12 +37,15 @@ ogr2ogr -overwrite -t_srs EPSG:4326 -f "ESRI Shapefile" "/data/" \
 TL_BOUNDS=$(ogrinfo -so -al "/data/${TL_SHAPEFILE_NAME}.shp" \
     | grep Extent | sed -E 's/.*\((.*), (.*)\) - \((.*), (.*)\).*/\1 \2 \3 \4/')
 
-# Make tiles and upload them to S3
+# Make tiles and upload them to S3.
+# 'tl' only has totally-quiet or very-verbose (every tile) output options, so this filters
+# the output to show the first line then every 1000th line thereafter.
 /usr/bin/time -f "\nTIMING: %C\nTIMING:\t%E elapsed %Kkb mem\n" \
-tl copy --quiet --min-zoom "${TL_MIN_ZOOM}" --max-zoom "${TL_MAX_ZOOM}" \
+tl copy --min-zoom "${TL_MIN_ZOOM}" --max-zoom "${TL_MAX_ZOOM}" \
     --bounds "${TL_BOUNDS}" \
     "mapnik:///opt/pfb/tilemaker/styles/${TL_SHAPEFILE_NAME}_style.xml" \
-    "s3://${AWS_STORAGE_BUCKET_NAME}/${PFB_S3_TILES_PATH}/{z}/{x}/{y}.png"
+    "s3://${AWS_STORAGE_BUCKET_NAME}/${PFB_S3_TILES_PATH}/{z}/{x}/{y}.png" \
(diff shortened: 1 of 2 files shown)


spencerrecneps:
@KlaasH can you elaborate on the missing roads? Is there an example you can point me to so I can make sure we're not missing something on the import?

spencerrecneps:
@KlaasH Ah never mind. I didn't notice the base map had roads drawn. I see what you're saying. I'll investigate. My guess is those are alleys, which we don't include.

spencerrecneps:
Yep. Those are tagged as alleys. We should be OK.

KlaasH:
That screenshot was most of my screen, so it should be viewed zoomed in a little vs. how it's shown inline on GitHub.  I made the gap depend on zoom level because one pixel looked too small when zoomed in but too big when zoomed out.

But also, the testing instructions could be more complete.  How about this?

- Make a directory and copy your generated tiles into it:
```
mkdir raster_preview  && cd raster_preview
aws --profile pfb s3 sync s3://USERNAME-pfb-storage-us-east-1/results/JOB_ID/tiles .
```
- Copy this file into the same directory, as `index.html`:
```
<!DOCTYPE html>
<html>
    <head>
        <title>Vector Tiles Example</title>
        <meta charset="utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0">

        <link rel="stylesheet" href="https://unpkg.com/leaflet@1.0.3/dist/leaflet.css" />
        <script src="https://unpkg.com/leaflet@1.0.3/dist/leaflet.js"></script>
    </head>

    <body style='margin:0'>
        <div id="map" style="width: 100vw; height: 100vh; background: white"></div>
        <script type="text/javascript">
var map = L.map('map');

var baseLayer = L.tileLayer('http://server.arcgisonline.com/ArcGIS/rest/services/Canvas/World_Light_Gray_Base/MapServer/tile/{z}/{y}/{x}', {
    attribution: 'Tiles &copy; Esri &mdash; Esri, DeLorme, NAVTEQ',
    maxZoom: 17,
    minZoom: 3,
    noWrap: true  
}).addTo(map);

var rasterLayer = L.tileLayer("/{z}/{x}/{y}.png", {}).addTo(map);

map.setView({ lat: 40, lng: -75 }, 9);
</script>
    </body>

</html>
```
- Install and run the Node mini-server (assumes you have Node installed)
```
npm install http-server
./node_modules/http-server/bin/http-server
```
- Go to the URL it prints (probably http://127.0.0.1:8080).  It should show a map with your tiles on it.

PR #278, opened 2017-04-12
Closes #273: Save job names/defs to AnalysisJob model

### LINKED ISSUE #273: AnalysisJob.logs_url reports wrong URL
This is because the  `AnalysisJob.analysis_job_name` is a derived property, which includes `settings.PFB_AWS_BATCH_ANALYSIS_JOB_DEFINITION_NAME_REVISION`. This exposes a larger issues, which is that that setting value changes across deployments. Any computed property of AnalysisJob that uses these settings should instead be saved to the db model when a job is triggered.

CloudNiner:
## Overview

The `settings.PFB_AWS_BATCH_ANALYSIS_JOB_DEFINITION_NAME_REVISION` and `settings.PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION` values get incremented with every deployment, and we use this value to effectively track a few different things in the AnalysisJob model. By always lazily generating job names/definitions/etc via `@property` we lose the association to the actual values when a deployment occurs because revision is incremented. This PR changes the inner workings of the `job_name` properties so that we lazily save the proper value the first time these properties are accessed, and then just return that afterwards.

### Notes

Unable to use ModelField(default=callback) option
because we need the job uuid to create the job name, and args cannot be provided to the callback.

Other option was to make job name fields DB fields instead of
properties directly, but that requires additional save logic
and a data migration that would take more time to write,
and we don't filter/sort on these fields anyways. Could
still migrate to that later if necessary, as this puts the
first half of that machinery in place.

This solution feels a bit smelly and I'm not totally sold on it, in that we don't typically write Django models this way, and still has the disadvantage of being unable to filter/sort via ORM. But, the extra code required if we were to go for the solution in the paragraph above also feels meh (example in #275) because we have to duplicate code in data migrations. And it would take a bit longer to implement, using time which can be spent on other more pressing things ATM. 

## Testing Instructions

Run migrations. Create an analysis job via API, then access `job.analysis_job_name` and `job.tilemaker_job_name` via the Django shell. Should get the same job names as before. Access the `_` versions afterwards, and will see that the values are saved.

Closes #273


src/django/pfb_analysis/serializers.py
```diff
@@ -27,7 +27,9 @@ def get_neighborhood_label(self, obj):
 
     class Meta:
         model = AnalysisJob
-        exclude = ('created_at', 'modified_at', 'created_by', 'modified_by', 'overall_scores',)
+        exclude = ('created_at', 'modified_at', 'created_by', 'modified_by', 'overall_scores',
+                   'analysis_job_definition', 'tilemaker_job_definition',
+                   '_analysis_job_name', '_tilemaker_job_name',)
```
KlaasH: Time to switch to a whitelist?  I believe it would be
```
        fields = ('uuid', 'createdAt', 'modifiedAt', 'running_time', 'start_time', 'status',
                  'neighborhood_label', 'neighborhood', 'overall_score', 'batch_job_id',
                  'osm_extract_url', 'census_block_count', 'batch')
```
Not shorter, but not longer, and whitelists are easier to read and sometimes produce fewer unwelcome surprises.

CloudNiner: Went for it, then got exceptions. Added separate issue #282 to address later if time.

PR #279, opened 2017-04-12
Add filters for analysis job by neighborhood name

### LINKED ISSUE #243: Add 'neighborhood.name' search/filtering to AnalysisJob list in API
Used by search view dropdown

flibbertigibbet:
## Overview

Filter AnalysisJob API endpint by neighborhood name (slug) or label, either as exact match or substring.
Closes #243.

## Notes

Here's some relevant [django filter docs](https://django-filter.readthedocs.io/en/latest/ref/filters.html#).

## Testing Instructions

 * Create one or more analysis jobs
 * Should be able to query analysis jobs by neighborhood name (slug) or label
 * Should be able to query for exact match
 * Should be able to query for substring using `__contains` syntax, like: http://localhost:9200/api/analysis_jobs/?neighborhood__label__contains=somestring

### DIFF
--- a/src/django/pfb_analysis/filters.py
+++ b/src/django/pfb_analysis/filters.py
@@ -59,4 +59,9 @@ class AnalysisJobFilterSet(filters.FilterSet):
 
     class Meta:
         model = AnalysisJob
-        fields = ['neighborhood', 'batch', 'status', 'latest']
+        fields = {'neighborhood': ['exact', 'in'],
+                  'neighborhood__name': ['exact', 'contains'],
+                  'neighborhood__label': ['exact', 'contains'],
+                  'batch': ['exact', 'in'],
+                  'status': ['exact', 'in'],
+                  'latest': ['exact', 'in']}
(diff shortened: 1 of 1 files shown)


src/django/pfb_analysis/filters.py
```diff
@@ -59,4 +59,9 @@ def filter_latest(self, queryset, name, value):
 
     class Meta:
         model = AnalysisJob
-        fields = ['neighborhood', 'batch', 'status', 'latest']
+        fields = {'neighborhood': ['exact', 'in'],
+                  'neighborhood__name': ['exact', 'contains'],
+                  'neighborhood__label': ['exact', 'contains'],
+                  'batch': ['exact', 'in'],
+                  'status': ['exact', 'in'],
+                  'latest': ['exact', 'in']}
```
KlaasH: `'in'` doesn't work for `status`, since it's a method filter.  Adding e.g. `status__in=COMPLETE,TILING` gets quietly ignored.  Another reason to aspire to a query-based status filter.

`latest` is also a method filter, but fortunately `'in'` doesn't really make sense in that case anyway, since it's boolean.

flibbertigibbet: Good catch. It's unfortunate that the directory field structure requires the fields to map to some filter type list. I tried using the `__all__` magic keyword to simply enable all filter types whether they made sense or not, but that broke things as it apparently doesn't work with cross-relationship fields included.

Added `in` as an option since I noticed the methods were using `__in` filters, so that seemed to be the current behavior, to do an inexact match. Do things still work as expected without a filter method type suffix? If so, I'll just leave them with just `exact`.

flibbertigibbet:
Dropped the `in` filters for `status` and `latest`.

PR #284, opened 2017-04-12
Export transit geojson to S3/Admin UI

CloudNiner:
## Overview

Exports new transit geojson to S3 and makes available in AnalysisJob detail view in admin UI

## Testing Instructions

 * Create and run a job all the way through to completion, analysis+tilemaker
 * View transit results in AnalysisJob detail view in admin UI


CloudNiner:
This is trivial, merging so it can be included in the initial attempt at a production deployment which I'm about to do.

Issue #192, opened 2017-03-21
Add production environment


CloudNiner:
The timeline for this is as follows:

We will continue to use staging for initial testing of the batch analysis process with Spencer/Jess for now.

PFB expects that our first batch analysis run of the initial set of cities will take place starting April 17. 

We'd need to have a production stack ready to go by then, so this task should happen in the April 6-14 timeframe.

CloudNiner:
Until #151 is addressed, we'll need to manually scale the Batch ECS cluster production instances while we're running the first batch.

CloudNiner:
Start after April 10

CloudNiner:
Production stack setup and ready for first analysis batch.

Manual deployments will be required for now, our Jenkins environment is currently not configured to run builds for the master branch. Manual deployment steps are the same as for staging, and instructions are in the deployment README.

PR #286, opened 2017-04-13
Production deployment

CloudNiner:
## Overview

Provides the necessary updates to end up with a successful production environment on AWS.

Commit https://github.com/azavea/pfb-network-connectivity/commit/d32e6a5ebe056702ed119e220a473f6fdd840cfd is live on the production stack.

### Demo

screen shot 2017-04-13 at 09 49 00

### Notes

Got error:
```
* aws_ecs_service.pfb_app_http: InvalidParameterException: The target group with targetGroupArn arn:aws:elasticloadbalancing:us-east-1:950872791630:targetg$
oup/tf-tg-6defea6dd9c9c22ca88749d107/54fbe71805823763 does not have an associated load balancer.
        status code: 400, request id: 32e019c6-2041-11e7-9b66-c5845ed8449f "ProductionHTTPAppServer"
```
while launching stack but I believe that was fallout from the previous error where the IAM Role BatchManageJobs failed to create because it already existed. After fixing the IAM role error, deleting the pfb_app_http ECS task and re-applying, the stack launched successfully.

Will be opening a separate PR to remove the `batch_analysis_compute_environment_arn` tfvar as a followup. It is only passed into Django, which does not use it. 

Do we want to setup the Jenkinsfile to automatically push new commits to `master` to the production environment? Probably. That can be a separate PR as well.

## Testing Instructions

Navigate to https://pfb.azavea.com. Password in fileshare. Should be able access the same features as staging.

NOTE: The batch container instance is currently disabled, so jobs can be created, but they will not run.

@azavea/operations can you take a look?

Connects #192

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -30,6 +30,7 @@ You will be prompted to enter your AWS credentials, along with a default region.
 If you're deploying new application code, first set the commit to deploy, then build and push containers:
 ```bash
 vagrant@vagrant-ubuntu-trusty-64:~$ export GIT_COMMIT="<short-commit-to-deploy>"
+vagrant@vagrant-ubuntu-trusty-64:~$ export ENVIRONMENT="staging|production"
 vagrant@vagrant-ubuntu-trusty-64:~$ export PFB_AWS_ECR_ENDPOINT="<aws-account-id>.dkr.ecr.us-east-1.amazonaws.com"
 vagrant@vagrant-ubuntu-trusty-64:~$ ./scripts/cibuild && ./scripts/cipublish
 ```
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -38,7 +38,7 @@ vagrant@vagrant-ubuntu-trusty-64:~$ ./scripts/cibuild && ./scripts/cipublish
 Next, use the `infra` wrapper script to lookup the remote state of the infrastructure and assemble a plan for work to be done:
 
 ```bash
-vagrant@vagrant-ubuntu-trusty-64:~$ export PFB_SETTINGS_BUCKET="staging-pfb-config-us-east-1"
+vagrant@vagrant-ubuntu-trusty-64:~$ export PFB_SETTINGS_BUCKET="${ENVIRONMENT}-pfb-config-us-east-1"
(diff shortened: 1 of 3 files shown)


CloudNiner:
I performed the following checklist to launch the new production environment:
- [ ] Copy staging tfvars to new file
- [ ] Update relevant variables in the new production tfvars file
- [ ] Create production config bucket
- [ ] Upload production tfvars file to the appropriate config bucket keypath, with encryption set
- [ ] Create production AWS Batch resources following instructions [here](https://github.com/azavea/pfb-network-connectivity/tree/master/deployment#aws-batch)
- [ ] Apply new terraform changes by following instructions [here](https://github.com/azavea/pfb-network-connectivity/tree/master/deployment#terraform)
- [ ] Wait for stack to launch
- [ ] Add PFB AWS account NS records to Azavea account so that temporary production DNS resolves correctly
- [ ] Run migrations via ECS management task
- [ ] Generate and update root account password via Django shell on ECS instance worker
- [ ] Add root account password to password manager

hectcastro:
Nice work. In addition to the items in your checklist, I did the following:

- [x] Enable versioning on Terraform state management S3 buckets
- [x] Added bucket policies for Terraform state management S3 buckets so that only SSE operations are allowed

---

> but I believe that was fallout from the previous error where the IAM Role BatchManageJobs failed to create because it already existed. After fixing the IAM role error, deleting the pfb_app_http ECS task and re-applying, the stack launched successfully.

I think that is true as well.

> Do we want to setup the Jenkinsfile to automatically push new commits to master to the production environment? Probably. That can be a separate PR as well.

The high level (and slightly aggressive) approach we're taking for Raster Foundry is to publish `release/*` directly into production. This gives us a way to append any tweaks that are necessary for the release to be successful before finalizing it.

CloudNiner:
Hmm, deploying to production on release branches seems a bit aggressive for a client project, particularly this one, and for code that isn't well covered by tests. We also want to be deliberate about when we cut releases to the production environment so we can alert the client.

Auto-deployment when master is updated seems like a good middle-ground between your suggestion and manual deployments?

CloudNiner:
Either way, any changes to the deployment process in Jenkins will be a separate PR, so I'm going to merge this one awhile. Thanks for taking a look.

PR #292, opened 2017-04-14
Include neighborhood object in job api results

### LINKED ISSUE #264: Update AnalysisJobSerializer with more info
Instead of returning neighborhood uuid and neighborhood_label in AnalysisJobSerializer, neighborhood should be a nested object which contains:
- [ ] uuid
- [ ] label
- [ ] state
- [ ] geom point

AnalysisJobSerializer should also provide the 'overall_score' metric saved in the overall_scores field

KlaasH:
## Overview

Switches the analysis-jobs endpoint from including UUID and label as properties on the job to including `neighborhood` as an object, with those details plus state and point geometry inside.

Since DRF doesn't have easy automatic behavior for saving inlined related objects, I borrowed an idea from MUB and made neighborhood on the AnalysisJob serializer into a field that expects UUID for create/save operations but includes the serialized object when serializing for presentation.

### Demo

Example result:
```
        {
            "uuid": "acc421de-2832-4252-87e0-5f1c828452bb",
            "createdAt": "2017-04-07T01:12:12.827062Z",
            "modifiedAt": "2017-04-07T04:09:36.021285Z",
            "running_time": 11461,
            "start_time": "2017-04-07T01:16:32.001600Z",
            "status": "COMPLETE",
            "neighborhood": {
                "uuid": "08f827a6-c21a-4e45-959e-42cc509bb6fc",
                "name": "gtown-westside",
                "label": "Germantown, PA",
                "state_abbrev": "PA",
                "organization": "813f56de-1800-4b50-8515-a073eb32970d",
                "geom_pt": {
                    "type": "Point",
                    "coordinates": [
                        -75.17746130993164,
                        40.02685249307617
                    ]
                }
            },
            "overall_score": 23.0,
            "batch_job_id": "0981846f-f061-4910-820b-ad896f4220db",
            "osm_extract_url": null,
            "census_block_count": 2700,
            "batch": null
        },
```

### Notes

I made a new serializer for the partial neighborhood included with the analysis job (`NeighborhoodSummarySerializer`).  It's extremely similar to `NeighborhoodGeoJsonSerializer`, except that it specifies all the fields as read-only and it's not a Geo serializer so the point just goes in with other attributes rather than in a Feature/geometry/properties structure.  I figured since this endpoint isn't returning GeoJSON, there's no making the inlined neighborhood GeoJSON-like, but I could be persuaded to toss out `NeighborhoodSummarySerializer` and just use `NeighborhoodGeoJsonSerializer` if that seems better.

## Testing Instructions

- The `analysis_jobs` endpoint should work and include neighborhood information in the results.
- Creating a job in the front-end or the API should work as before, with `neighborhood: <uuid>` creating the association to the neighborhood.
- The Analysis Job detail admin view should show the neighborhood label.

Resolves #264.

### DIFF
--- a/src/django/pfb_analysis/serializers.py
+++ b/src/django/pfb_analysis/serializers.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from rest_framework import serializers
 
 from pfb_analysis.models import AnalysisJob, Neighborhood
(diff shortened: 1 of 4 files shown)


src/django/pfb_analysis/views.py
```diff
@@ -70,6 +70,7 @@ def results(self, request, pk=None):
 class NeighborhoodMixin(APIView):
     """Shared properties of the neighborhood viewsets."""
 
+    queryset = Neighborhood.objects.all()
```
flibbertigibbet: Tried putting this property on the mixin when adding it initially, but it would error. Thought this simply wouldn't generalize because `ModelViewSet` and `ReadOnlyModelViewSet` separately inherit from `GenericAPIViewSet` and not from each other, but curiously, this does seem to work now.

src/django/pfb_analysis/serializers.py
```diff
@@ -1,38 +1,58 @@
...
 from rest_framework import serializers
 
 from pfb_analysis.models import AnalysisJob, Neighborhood
 from pfb_network_connectivity.serializers import PFBModelSerializer
 from rest_framework_gis.serializers import GeoFeatureModelSerializer
 
 
-class AnalysisJobSerializer(PFBModelSerializer):
+class PrimaryKeyReferenceRelatedField(serializers.PrimaryKeyRelatedField):
+    """ A custom relational field for read-only objects
```
KlaasH: I feel like docstring style diversity in the wild is somewhat high, and I had gotten muddled from seeing various patterns that differ slightly but are all obviously intentional.  I've now perused PEP 257 again and attempted to bring this file (not the project, just the file I happen to be modifying) into line with it.

src/django/pfb_analysis/views.py
KlaasH: I wonder if that was fallout from the ordering.  Or did you get the crash even with the mixin first?  Anyway yeah, this seems to be working.

PR #294, opened 2017-04-14
Overall scores in census blocks

spencerrecneps:
## Overview

Adds new field "overall_score" to neighborhood_census_blocks. This is the field we can use to generate tiles for viewing census block scores.


spencerrecneps:
Since this hasn't been merged yet I went ahead and pushed a few more commits to this branch related to breaking out the retail scoring into its own category.

PR #299, opened 2017-04-17
Load prototype styles and navbar

### LINKED ISSUE #291: Load prototype into AngularJS site

KlaasH:
## Overview

- Removes all the old styles and replaces them with the styles from the prototype.
- Does a rough first pass at modifying the navbar to show admin links on admin views and public links otherwise.
- Loads the html for the home view but doesn't implement any of the behavior.

### Demo

screen shot 2017-04-17 at 12 55 51-fullpage
_Looks a lot like the prototype. Surprise!_

### Notes

- Removing all the old styles broke the existing admin site.  It's still functional, including the filters, but it doesn't look good.

screen shot 2017-04-17 at 13 00 41

I tried briefly to fix it--i.e. find the specific styles from the old site that are missing and pull them in--but didn't succeed and figured it should go into another card rather than delaying this blocker.

- I've been having some staleness issues that I haven't managed to track down--e.g. switching to this branch from develop leaves me with not-fully-broken admin pages, but rebuilding and restarting the container clears them out.  Again, there's something to track down in the build setup but it seems like not a big enough issue to hold everything up.

## Testing Instructions

Rebuild and start the angular container and view the site.  Going to http://localhost:9301/ should now show the public site home page rather than forwarding you to login.
Going to http://localhost:9301/#/login/ and logging in should get you to the admin pages (sans styling) and admin header.

Resolves #291.


src/angularjs/src/app/components/navbar/navbar.directive.js
```diff
@@ -35,7 +31,11 @@
             templateUrl: 'app/components/navbar/navbar.html',
             controller: 'NavbarController',
             controllerAs: 'navbar',
-            bindToController: true
+            bindToController: true,
+            scope: {
+                admin: '@',
```
CloudNiner: Fine to start, but this is state you could theoretically acquire from an AuthService of some kind.

KlaasH: Yeah, there are nuances to be worked out in how the navbar behaves and what it shows when.  The parameter isn't intended so much as "is there an admin user logged in" as "is this an admin page".  So maybe the `$state` service rather than `AuthService` for that part.  Though showing a login link for users who aren't logged in and a link to the admin views for users who are might be a sensible thing to do, and would probably hang on `AuthService`.

KlaasH:
Noticed this is making the "Run Analysis" button disappear.  Trying to figure out why.

KlaasH:
Figured out where the "Run Analysis" button went--it was in an `ng-if` based on a property of the navbar controller, but the scopes had shifted and it was no longer available to the list view's template.  More details about that and what I did to fix it in the commit message for 61c3100.

@CloudNiner, could you take another look, since I just went and changed a bunch more stuff?

CloudNiner:
Sorry I missed your "updated" comment yesterday. This LGTM

PR #300, opened 2017-04-17
Add boundary population category to neighborhood_overall_scores

spencerrecneps:
## Overview

Adds two new entries to neighborhood_overall_scores. One is the total population of the boundary area (summary of pop10 attribute in census blocks). The second is a categorization of the city:

- 1=large
- 2=medium
- 3=small

Current breaks are set at >= 500,000 (large city) and >= 100,000 (medium city).

The score_id is `population_total` for the population and `population_category` for the category.

## Testing Instructions

1. Run analysis
2. Check for new scores

### DIFF
--- a/src/analysis/connectivity/overall_scores.sql
+++ b/src/analysis/connectivity/overall_scores.sql
@@ -11,10 +11,11 @@
 --
 -- variables:
 --   :total=100
---   :people=20
+--   :people=15
 --   :opportunity=25
---   :core_services=30
+--   :core_services=25
 --   :recreation=10
+--   :retail=10
 --   :transit=15
 ----------------------------------------
 DROP TABLE IF EXISTS generated.neighborhood_overall_scores;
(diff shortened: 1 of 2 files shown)


src/analysis/connectivity/overall_scores.sql
```diff
@@ -349,3 +350,32 @@ SELECT  'overall_score',
...
+                        WHERE   ST_Intersects(b.geom,neighborhood_census_blocks.geom)
+                    )
+        ),
+        'Total population of boundary';
+
+-- population category
+INSERT INTO generated.neighborhood_overall_scores (
+    score_id, score_original, human_explanation
+)
+SELECT  'population_category',
```
CloudNiner: While this is useful for documentation purposes, we don't need to create this score here. We can configure these buckets at the application/UI display level, which will be far more flexible.

src/analysis/connectivity/overall_scores.sql
```diff
@@ -349,3 +350,32 @@ SELECT  'overall_score',
...
+SELECT  'population_total',
+        (
+            SELECT SUM(pop10) FROM neighborhood_census_blocks
+            WHERE   EXISTS (
+                        SELECT  1
+                        FROM    neighborhood_boundary AS b
+                        WHERE   ST_Intersects(b.geom,neighborhood_census_blocks.geom)
+                    )
+        ),
+        'Total population of boundary';
```
CloudNiner: Can you also update the json file so that they contain these new scores, along with the description text?

spencerrecneps:
Done

Issue #251, opened 2017-04-06
Implement compare feature on neighborhood list page


KlaasH:
Depends #248.

flibbertigibbet:
Relates to and possibly depends on #249. There should be a modal centered over the map view displaying an overview of comparison city scores when at least one comparison city selected.

Issue #250, opened 2017-04-06
Implement sorting/filtering UI controls on neighborhood list page


KlaasH:
Has to be server-side, since the list will be paginated.

Depends on #248.

CloudNiner:
Depends on #242 #243

PR #314, opened 2017-04-18
Feature/kak/neighborhood angular list#248

### LINKED ISSUE #248: Implement list on neighborhood list page
With pagination, possibly infinite scroll.

flibbertigibbet:
## Overview

Implement public neighborhood list view.

### Demo

image

### Notes

Things not done here:
 - Show an actual score (data not in neighborhood API response)
 - Implement "shape preview" image, whatever that's supposed to be
 - Implement sorting/filtering or compare (to be done with #250 and #251)
 - Implement section headers grouping by first letter of alphabet (if done, would need to be conditional on sort type and depends on #250)
 - Make content available without login (should be public view, but API requires authentication)

Regarding the styling: the navbar seems to overlap the page content somewhat, cutting off the top of the list header.

## Testing Instructions

 * At http://localhost:9301/#/places/
 * Also should be able to get there via 'All locations' navbar link
 * Should list neighborhoods with correct info and working pagination
 
Closes #248

### DIFF
--- a/src/angularjs/src/app/places/list/places-list.controller.js
+++ b/src/angularjs/src/app/places/list/places-list.controller.js
@@ -40,7 +40,7 @@
         }
 
         function getPlaces(params) {
-            params = params || $stateParams;
+            params = params || _.merge({}, $stateParams, defaultParams);
             AnalysisJob.query(params).$promise.then(function(data) {
 
                 ctl.places = _.map(data.results, function(obj) {
--- a/src/angularjs/src/app/places/list/places-list.controller.js
+++ b/src/angularjs/src/app/places/list/places-list.controller.js
@@ -11,12 +11,13 @@
 
     /** @ngInject */
     function PlaceListController($state, $stateParams, $scope, Pagination, AuthService,
-                                 Neighborhood) {
+                                 Neighborhood, AnalysisJob) {
         var ctl = this;
 
         var defaultParams = {
             limit: null,
-            offset: null
+            offset: null,
+            latest: 'True'
         };
         var nextParams = {};
         var prevParams = {};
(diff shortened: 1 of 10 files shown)


src/angularjs/src/app/places/list/places-list.controller.js
```diff
@@ -0,0 +1,84 @@
...
+            ctl.places = [];
+
+            ctl.filters = {};
+
+            getPlaces();
+        }
+
+        function getPlaces(params) {
+            params = params || $stateParams;
+            Neighborhood.query(params).$promise.then(function(data) {
```
flibbertigibbet: Was thinking we'd otherwise need to separately query job status for each neighborhood on the front end, which isn't great, or add some query parameter to optionally get latest status from the neighborhood endpoint.

Querying for jobs instead does have some implications: only neighborhoods for which at least one job has been attempted will show in the list, and all our searching/filtering by neighborhood attributes need to work on the jobs endpoint.

flibbertigibbet:
Updated to query by job:
image

src/angularjs/src/app/places/list/places-list.controller.js
```diff
@@ -0,0 +1,90 @@
...
+            ctl.getPrev = getPrev;
+            ctl.places = [];
+
+            ctl.filters = {};
+
+            getPlaces();
+        }
+
+        function getPlaces(params) {
+            params = params || $stateParams;
```
KlaasH: `defaultParams` needs to be pulled in here or you don't get the `latest=True` (which is probably more obvious on my box, with lots of repeat runs, than on some...).  Perhaps 
```
params = _.merge(defaultParams, $stateParams, params);
```
(Side question: what is `$stateParams` accomplishing here?  Is that to accept params in the page URL and pass them through to the query?)

flibbertigibbet: Good catch. Yes, `$stateParams` is used elsewhere for passing around ID filters and such. Useless for the list here at the moment, but maybe it might come in handy.

PR #315, opened 2017-04-18
Don't try to export features with empty geometry

### LINKED ISSUE #295: Glendale AZ job fails with reprojection error in neighborhood_pharmacies geojson export
Glendale job failed because the neighborhood_pharmacies table threw a reprojection error when attempting to export that table as geojson.

https://staging.pfb.azavea.com/#/admin/analysis-jobs/d95d0f71-21e3-4088-8fc8-6df10226ec47/

[AWS Log](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=/aws/batch/job;stream=staging-pfb-analysis-run-job--219--d95d0f71/c57348fa-5460-4d25-b37e-0040ecd44931/f16fe022-dabd-463a-ba2c-232b8734a082)

Full log:
```
14:03:06
EXPORTING: neighborhood_pharmacies

14:03:06
Warning 1: More than one geometry column was found in the result of the SQL request. Only last one will be used

14:03:06
Warning 1: More than one geometry column was found in the result of the SQL request. Only last one will be used

14:03:06
translation from sql statement.
translation from sql statement.

14:03:06
Failed to reproject feature 11 (geometry probably out of source or destination SRS).
Failed to reproject feature 11 (geometry probably out of source or destination SRS).

14:03:06
ERROR 1: Terminating translation prematurely after failed

14:03:06
ERROR 1: non-convergent inverse meridional dist

14:03:06
LOG: unexpected EOF on client connection with an open transaction

14:03:06
Updating job status: ERROR Failed See job logs for more details.

14:03:06
LOG: received fast shutdown request
```

KlaasH:
## Overview

Protects against empty geometry, which makes `ogr2ogr`, and hence the export script, crash.

### Demo

A little hard to pick out between the "More than one geometry column" warnings, but here's the command that gets run by `export_connectivity.sh` failing then succeeding in its "before" and "after" forms.

image

### Notes

- I looked into downloading the `pgdata` for jobs off the batch worker.

Per [this answer](https://unix.stackexchange.com/a/327100) referencing [this line of the `scp` code](https://github.com/openssh/libopenssh/blob/master/ssh/scp.c#L375), scp doesn't support agent forwarding.  The strategy I did get working ([from here](https://serverfault.com/a/37664)) was opening one connection to the bastion that forwards the ssh port on the batch instance to another port, which you can then use to establish scp connections.
```
ssh -A -L 5022:<batch instance IP>:22 ec2-user@bastion.staging.pfb.azavea.com
scp -P 5022 ec2-user@localhost:<remote_path> <local_path>
```
The problem with that is that the remote files are owned by `root`, and `ec2-user` can't read them.  Which can be addressed either by tarring them up first, e.g. `ssh -p 5022 ec2-user@localhost "sudo tar czf /media/nvme0n1/pgdata/JOB_ID.tar.gz /media/nvme0n1/pgdata/JOB_ID"` then downloading that (it's created readable by other users) or by running a remote sudo command to change the permissions of the directory (which doesn't seem ideal, but then again the analysis is finished so it won't really be interfering with anything).

I think uploading to S3 and downloading from there makes the most sense.  The "tar then download" strategy, in particular, seems not ideal because it uses disk space, disk I/O, and CPU on the worker instance.  Anything will use at least some resources, but copying to S3 from EC2 is very fast, and once that's done you no longer have to rely on the EC2 instance.

- Speaking of which, it turned out none of this was actually relevant to my debugging of this issue because the batch worker instance was replaced yesterday so the working files from the actual analysis run are gone.  Only took about half an hour locally, though.

- The non-pharmacy is getting inserted at https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/connectivity/destinations/pharmacies.sql#L23.  This issue could also be addressed there, but I'm not that clear on what that's doing and why (it pulls a bunch of multipolygons with no attributes at the line mentioned, then adds point geometries for those and also pulls a bunch of points), so I went for the more obvious fix.

## Testing Instructions

I don't know how to reproduce this except with Glendale.  I could try to extract my database, since the one on staging is gone.

Resolves #295.

### DIFF
--- a/src/analysis/scripts/export_connectivity.sh
+++ b/src/analysis/scripts/export_connectivity.sh
@@ -80,7 +80,7 @@ function ec_export_table_geojson() {
   EXPORT_TABLENAME="${2}"
 
   FILENAME="${OUTPUT_DIR}/${EXPORT_TABLENAME}.geojson"
-  ogr2ogr -f GeoJSON "${FILENAME}" \
+  ogr2ogr -f GeoJSON "${FILENAME}" -skipfailures \
           -t_srs EPSG:4326 \
           "PG:host=${NB_POSTGRESQL_HOST} dbname=${NB_POSTGRESQL_DB} user=${NB_POSTGRESQL_USER}" \
           -sql "select * from ${EXPORT_TABLENAME}"
@@ -95,7 +95,7 @@ function ec_export_destination_geojson() {
   # Our version of ogr2ogr isn't new enough to specify the geom column :(
   #   Instead, ogr2ogr states it takes the "last" geom column, so we manually specify
   #   it here to ensure we always take the one we want
-  ogr2ogr -f GeoJSON "${FILENAME}" \
+  ogr2ogr -f GeoJSON "${FILENAME}" -skipfailures \
           -t_srs EPSG:4326 \
           "PG:host=${NB_POSTGRESQL_HOST} dbname=${NB_POSTGRESQL_DB} user=${NB_POSTGRESQL_USER}" \
           -sql "select *, geom_pt from ${EXPORT_TABLENAME}"


src/analysis/scripts/export_connectivity.sh
```diff
@@ -98,7 +98,7 @@ function ec_export_destination_geojson() {
   ogr2ogr -f GeoJSON "${FILENAME}" \
           -t_srs EPSG:4326 \
           "PG:host=${NB_POSTGRESQL_HOST} dbname=${NB_POSTGRESQL_DB} user=${NB_POSTGRESQL_USER}" \
-          -sql "select *, geom_pt from ${EXPORT_TABLENAME}"
+          -sql "select *, geom_pt from ${EXPORT_TABLENAME} where not ST_IsEmpty(geom_pt)"
```
CloudNiner: I would expect the other `ec_export_table_geojson` function to fail at some point because of a null geometry, but I'm not sure its possible to filter on null geometries if you don't know what the geometry column title is. Maybe `--skipfailures` would work instead? Or a broad geom filter on the bounding box of the entire world using `clipsrc`?

spencerrecneps:
I'm confused as to how there's a NULL geometry in the geom_pt field. I'm pretty sure the scripts add centroids from geom_poly to the geom_pt field, so in theory a NULL value would only occur if both fields are empty, but then how would that feature have been included in OSM in the first place?

src/analysis/scripts/export_connectivity.sh
KlaasH: Yup, `-skipfailures` is way better.  Covers more cases, plus it logs errors but still exits normally.  Pushed a commit to switch to that.  I'll rebase the first one out of existence before merging.

CloudNiner:
👍

KlaasH:
@spencerrecneps 
Seems the geometry isn't null, it's "EMPTY".  I.e. a polygon with no actual points.  In the Glendale database, this query:
```
select osm_id, name, st_asewkt(way) from neighborhood_osm_full_polygon WHERE amenity='pharmacy';
```
returns 22 rows, two of which are
```
-6325983 | CVS        | SRID=32612;POLYGON((392996.3 3710034.12,392996.87 3710076.08,393001.42 3710075.97,393006.83 3710075.83,393026.54 3710075.69,393026.24 3710052.9,393028.04 3710052.87,393027.81 3710035.65,393026.46 3710033.22,393023.16 3710032.11,393007.76 3710032.32,393007.78 3710033.97,392996.3 3710034.12))
-6325983 | CVS        | SRID=32612;POLYGON EMPTY
```
The first one might not be directly relevant, but I thought it interesting that the ID was duplicated.  I would guess that's a bit of bad data in OSM, though an `osm2pgrouting` error would be a possible fallback guess.

Issue #247, opened 2017-04-06
Implement map on homepage


CloudNiner:
Static map showing all cities, geojson probably fine

Depends on #246

PR #316, opened 2017-04-18
Reusable map component with basic configuration

CloudNiner:
## Overview

This is part 1 of a two part set to add the neighborhoods map at the bottom of the public homepage. I stopped here since this PR adds a HomeController and adds the MapDirective that can be used elsewhere. I will continue work on displaying the neighborhoods on the map in a separate branch/PR.

### Demo

screen shot 2017-04-18 at 17 09 37

### Notes

Kept the map directive as simple as possible. Added options to set map options/bounds/zoom/center/baselayer but anything else needs to be dealt with using the pfbMapReady function which provides the map object to the wrapping controller. This MapDirective should be sufficient for the individual static maps in the cards all over the site as well as for the larger more interactive maps.

@designmatty should we use a default basemap other than Stamen toner lite?

## Testing Instructions

 * Load up page, see map

Connects #247

### DIFF
--- a/src/angularjs/.eslintrc
+++ b/src/angularjs/.eslintrc
@@ -10,6 +10,7 @@
     "module": true,
     "URI": true,
     "_": true,
+    "L": true,
     "inject": true
   },
   "rules": {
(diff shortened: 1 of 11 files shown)


designmatty:
Sorry just seeing this. Yeah I think stamen actually works with their color scheme.

PR #319, opened 2017-04-19
Clip OSM extract by census block extent

### LINKED ISSUE #308: Truncate OSM data on import to boundary + NB_MAX_TRIP_DISTANCE

CloudNiner:
## Overview

Clips OSM extract by the extent of the trimmed census blocks which is equal to the extent of the boundary + buffer. In the default case, buffer == max_trip_distance. This leads to only slightly more OSM data than is necessary, which was not necessarily the case in the old extent calculation. It also allows us to reduce the amount of data we're pulling from the Overpass API.

### Demo

OSM extent for a local Germantown job using the MapZen Philadelphia OSM extract:
screen shot 2017-04-19 at 11 11 32

### Notes

`osmconvert` allows clipping via a `.poly` file, which is an arcane format which would have required extensive processing to create using the union of the clipped census blocks, and would only be slightly better in the 'average' case. In the best case, where the boundary+buffer is almost square, the extent would be almost equal to using a `.poly` file.

## Testing Instructions

- Run analysis job and generate neighborhood ways tiles
- Display tiles on map, following [instructions](https://github.com/azavea/pfb-network-connectivity/pull/277#issuecomment-293647993)
- Ensure output approximates the extent of the boundary+buffer

Closes #308

### DIFF
--- a/src/analysis/import/clip_osm.sql
+++ b/src/analysis/import/clip_osm.sql
@@ -1,0 +1,31 @@
+----------------------------------------
+-- INPUTS
+-- location: neighborhood
+-- :nb_boundary_buffer psql var must be set before running this script,
+--      e.g. psql -v nb_boundary_buffer=1000 -f clip_osm.sql
+----------------------------------------
+
+
+DELETE FROM neighborhood_ways AS ways
+    USING neighborhood_boundary AS boundary
+    WHERE NOT ST_DWithin(ways.geom, boundary.geom, :nb_boundary_buffer);
+
+DELETE FROM neighborhood_ways_intersections AS intersections
+    USING neighborhood_boundary AS boundary
+    WHERE NOT ST_DWithin(intersections.geom, boundary.geom, :nb_boundary_buffer);
+
+DELETE FROM neighborhood_osm_full_line AS lines
+    USING neighborhood_boundary AS boundary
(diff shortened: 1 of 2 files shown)


CloudNiner:
Fair point. While the updated clipped boundary is far smaller than what you show in the screenshot above, there would still be a significant amount of unnecessarily retained OSM ways. Let me see if I can put something together quickly. It's not trivial as you suggest, since there are quite a few OSM tables that get imported and used.

KlaasH:
So you're right, I hadn't registered that the old bbox was including a buffer the size of the boundary itself in every direction.  I ran both calculations for New York and got
```
Old bbox: -74.8181710000005, 40.0372210000001, -73.1409279999990, 41.3577549999998
New bbox: -74.25909,40.477399,-73.6631569999994,40.9415279999999
```
which looks like this:
image

But the screenshot of the neighborhood_ways tiles looks a lot more like the inner box, so that's making me confused.

CloudNiner:
Hmm you're right. I wonder if somewhere else in the analysis filters the box that the analysis is done on, maybe within the graph build or elsewhere? I'll dig a bit this am

CloudNiner:
Updated to clip used OSM data to boundary + buffer:

screen shot 2017-04-20 at 12 38 29
screen shot 2017-04-20 at 12 37 09

Issue #254, opened 2017-04-06
Implement neighborhood detail page


KlaasH:
View/route, layout, and text portions.  The map is #255.

PR #321, opened 2017-04-19
Load Neighborhood geojson on map

### LINKED ISSUE #247: Implement map on homepage

CloudNiner:
## Overview

Part 2 after #316. Adds the neighborhoods as geojson points to the map.

- Not styled
- Link to neighborhood detail view not hooked up. Can be done as part of #254 

@designmatty we'll need to think about how we want to style the markers/popups in this view

### Demo

screen shot 2017-04-19 at 14 40 00

### Notes

I intentionally left the map bounds set to CONUS because the final application will have points across the US.

Refactors the neighborhood map logic into a self-contained directive

Closes #247

### DIFF
--- a/src/angularjs/src/app/home/neighborhood-map.directive.js
+++ b/src/angularjs/src/app/home/neighborhood-map.directive.js
@@ -1,0 +1,62 @@
+
+(function() {
+
+    /* @ngInject */
+    function NeighborhoodMapController(Neighborhood) {
+        var ctl = this;
+        ctl.map = null;
+
+        ctl.$onInit = function () {
+            ctl.mapOptions = { scrollWheelZoom: false };
+            ctl.boundsConus = [[24.396308, -124.848974], [49.384358, -66.885444]];
+            ctl.baselayer = L.tileLayer(
+                'https://stamen-tiles.a.ssl.fastly.net/toner-lite/{z}/{x}/{y}.png', {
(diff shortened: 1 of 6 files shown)


CloudNiner:
Whoops. Realized I didn't implement the count in the map overlay callout. Working on that now.

CloudNiner:
Updated!

flibbertigibbet:
I'm seeing locations as points and not boundaries. Any idea why that is?

src/angularjs/src/app/home/neighborhood-map.directive.js
```diff
@@ -0,0 +1,63 @@
...
+            });
+
+            function onEachFeature(feature, layer) {
+                // TODO: Style marker and popup
+                // TODO: Add link to neighborhood detail in popup
+                layer.on({
+                    click: function () {
+                        var popup = L.popup()
+                            .setLatLng([
+                                feature.geometry.coordinates[1],
```
flibbertigibbet: Any chance of having a missing or broken geometry here we'd want to guard against?

CloudNiner: Hmm, would hurt to guard. Can add that.

CloudNiner:
Yeah, neighborhoods are points because we only return centroids from the endpoint.

PR #323, opened 2017-04-19
Generate census block tiles

### LINKED ISSUE #297: Generate and export  neighborhood_census_block tiles using "overall_score" column in neighborhood_overall_scores

### LINKED ISSUE #313: Update colors used in neighborhood ways static tile export
These are the colors to be used

Color for "high-stress" edges: Pantone 485 (red) `#e2231a`
Color for "low-stress" edges: Pantone Cyan (blue) `#00aeef`

KlaasH:
## Overview

Updates styling for neighborhood_ways tile layer and adds a neightborhood_census_blocks tile layer.

~The latter runs but produces only blank tiles, I'm not sure why.  Here's the shapefile and the Mapbox Studio Classic project I've been working with.~ Issues resolved. Not sure whether they were entirely projection-related or if I fixed another issue and broke the projection config simultaneously. Anyway, resolved.

### Demo

image

That's Glendale, AZ.  I packaged up the tiles and a demo page and put it on the fileshare at `PeopleForBikes_NetworkConnectivity/data/glendale_tile_demo.tar.gz`
Download, extract, change to directory, and run `./run.sh`, and it should launch a local http-server and print the URL to view it at.  (Unless you don't have node installed, in which case it will not do anything useful.)

### Notes

- I went with a 10-step color ramp because dividing the 100-point score scale into 11ths seemed didn't seem worth the trouble.
- I set it to 70% opacity, which makes it pretty bright but doesn't completely obscure the map underneath.
- I tried adding border styling, but since the borders are pretty much always roads, it just ended up getting in the way of seeing the base map streets and the segment stress.
- I set the default MAX_ZOOM to 18.  The maxZoom on the home page map controller is 18, and I would think the detail page would have the same value (or at least not less).  If that takes too much time or space, we could consider reducing the maxZoom on the maps, but we need to keep them in sync so people don't lose the layers when they zoom in all the way...

## Testing Instructions

Running tile generation is the same as before--it should just happen on staging, copy and run the command printed to the logs on development--but it should now do two tiling steps and instead of putting the output directly in `PFB_S3_TILES_PATH` it puts them in `PFB_S3_TILES_PATH/neighborhood_ways` and `PFB_S3_TILES_PATH/neighborhood_census_blocks`.

Resolves #297, resolves #313.

### DIFF
--- a/src/tilemaker/styles/neighborhood_census_blocks_style.xml
+++ b/src/tilemaker/styles/neighborhood_census_blocks_style.xml
@@ -17,48 +17,48 @@
 
 <Style name="neighborhood_census_blocks" filter-mode="first">
   <Rule>
-    <Filter>([OVERALL_SC] &gt;= 90) and ([OVERALL_SC] &lt; 100)</Filter>
-    <PolygonSymbolizer fill="#00aeef" fill-opacity="0.7" />
+    <Filter>([OVERALL_SC] &gt;= 54) and ([OVERALL_SC] &lt;= 100)</Filter>
+    <PolygonSymbolizer fill="#00aeef" fill-opacity="0.65" />
   </Rule>
   <Rule>
-    <Filter>([OVERALL_SC] &gt;= 80) and ([OVERALL_SC] &lt; 90)</Filter>
-    <PolygonSymbolizer fill="#1993d8" fill-opacity="0.7" />
+    <Filter>([OVERALL_SC] &gt;= 48) and ([OVERALL_SC] &lt; 54)</Filter>
+    <PolygonSymbolizer fill="#1993d8" fill-opacity="0.65" />
   </Rule>
   <Rule>
-    <Filter>([OVERALL_SC] &gt;= 70) and ([OVERALL_SC] &lt; 80)</Filter>
-    <PolygonSymbolizer fill="#3378c2" fill-opacity="0.7" />
+    <Filter>([OVERALL_SC] &gt;= 42) and ([OVERALL_SC] &lt; 48)</Filter>
+    <PolygonSymbolizer fill="#3378c2" fill-opacity="0.65" />
   </Rule>
   <Rule>
-    <Filter>([OVERALL_SC] &gt;= 60) and ([OVERALL_SC] &lt; 70)</Filter>
-    <PolygonSymbolizer fill="#4c5eac" fill-opacity="0.7" />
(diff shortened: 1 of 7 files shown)


CloudNiner:
Where's the documentation for the style language you're using in the XML file?

src/tilemaker/styles/neighborhood_census_blocks_style.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="utf-8"?>
```
CloudNiner: In diffing this file against the neighborhood_ways_style.xml, I see a few differences:
- Census block xml is missing a few items in the Parameters block: scale, metatile, tilejson, scheme
- Census block xml Layer definition has a different `srs` attribute. Presumably it should be the same as ways (WGS84 lng/lat for display on webmap)

I could easily see either of these issues leading to tiles not displaying how you'd expect, e.g. no complete definition for how to render tiles, or tiles being rendered in an unexpected projection

CloudNiner:
Can you get ahold of @designmatty real quick and demo the layers both individually and combined for him on your local machine? He wanted to take a look at the colors once they were implemented.

spencerrecneps:
I wonder if it makes sense to stop the color ramp somewhere around 60 or 70. We'll basically never have a block that hits 100, even in the best locations. We're muddying the differences down around 40-60, which is probably a range where the differentation really matters. Any thoughts? I've sent an email to Jennifer to get her thoughts. It kinda depends on whether we want to highlight the differences from census block to census block or whether we want to illustrate that virtually everywhere is "deficient" to some extent.

spencerrecneps:
Jennifer is on board with reducing the threshold for the top category as long as this can be changed easily in the future. I'm assuming this is a simple tweak in the XML file, if not let me know. Can we set it at >=54 for the top bucket. Everything else can scale from there. So the buckets would be:

1. 0-6
2. 6-12
3. 12-18
4. 18-24
5. 24-30
6. 30-36
7. 36-42
8. 42-48
9. 48-54
10. 54+

KlaasH:
0-54 ramp implemented.  Also neighborhood_ways lines widened a bit and opacity set to 65% per Matt's comments on the initial choices.

Here's a new screenshot:
image

CloudNiner:
👍 sounds good.

spencerrecneps:
That looks good. We may want to adjust the thresholds up or down once we have a bunch of cities to look at but that should be a quick change. Thanks!

Issue #322, opened 2017-04-19
Descending sort on analysis job scores orders nulls first


flibbertigibbet:
Ordering for highest scores first with: http://localhost:9200/api/analysis_jobs/?ordering=-overall_score

Returns jobs with no score first. Jobs missing scores should sort to end of descending list.

CloudNiner:
We now filter so that we only show completed jobs in the UI. It's not possible for a job to be complete and still have null scores.

PR #324, opened 2017-04-19
Feature/kak/places detail#254

### LINKED ISSUE #250: Implement sorting/filtering UI controls on neighborhood list page

### LINKED ISSUE #254: Implement neighborhood detail page

flibbertigibbet:
## Overview

Add places detail view and implement sorting and filtering for places list.
Closes #250 and closes #254.

### Demo

Neighborhood name dropdown filter:
image

A detail view:
image

Detail view when there's no scores to see:
image

### Notes

Did not attempt to implement download dropdown. Not sure what should go in there; there's _a lot_ of links available in the response, at various nesting levels.

## Testing Instructions

 * Go to http://localhost:9301/#/places/
 * Try filtering by neighborhood name: should be able to type stuff for autocomplete, or click things
 * Try changing sort order (note #322 may make for some odd-looking score sort results)
 * Click a neighborhood in the results list to go to details page; should have scores listed
 * Clicking a neighborhood without a valid completed job shows a message about 'no results available'

Closes #250 
Closes #254

### DIFF
--- a/src/angularjs/src/app/places/detail/places-detail.html
+++ b/src/angularjs/src/app/places/detail/places-detail.html
@@ -19,7 +19,7 @@
         </div>
     </div>
 
-    <!-- Everythign in sidebar-scrollable will scroll. -->
+    <!-- Everything in sidebar-scrollable will scroll. -->
     <div class="sidebar-scrollable">
         <section>
             <div class="metric-details">
@@ -34,20 +34,19 @@
                         </label>
                     </div>
                     <div class="column text-right">
-                        <div class="dropdown">
-                          <button class="btn btn-default dropdown-toggle btn-s" type="button" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
-                            Download
-                            <span class="caret"></span>
-                          </button>
-                          <ul class="dropdown-menu" aria-labelledby="dropdownMenu1">
-                            <li>
-                                <a href="#" class="compare-remove">Shapefile</a>
-                            </li>
-                            <li>
-                                <a href="#" class="compare-remove">CSV</a>
-                            </li>
-                          </ul>
-                        </div>
+
+                    <div class="btn-group dropdown" uib-dropdown is-open="status.isopen"
+                        ng-if="placeDetail.downloads">
+                      <button id="single-button" type="button" class="btn btn-default btn-s" uib-dropdown-toggle ng-disabled="disabled" id="downloadButton">
+                        Download <span class="caret"></span>
+                      </button>
+                      <ul class="dropdown-menu" uib-dropdown-menu role="menu"
+                        aria-labelledby="downloadButton">
+                        <li ng-repeat="item in placeDetail.downloads">
+                            <a role="menuitem" href="{{item.url}}"
+                                class="compare-remove">{{item.label}}</a>
+                        </li>
+                      </ul>
                     </div>
                 </div>
                 <div class="row" ng-if="!placeDetail.lastJobScore">
--- a/src/angularjs/src/app/places/detail/places-detail.html
+++ b/src/angularjs/src/app/places/detail/places-detail.html
@@ -1,0 +1,76 @@
+<pfb-navbar></pfb-navbar>
+
+<!-- Sidebar -->
+<!-- ATTENTION: note the use of location-overview class -->
+<div class="sidebar location-overview">
+
+    <!-- Everything in .sidebar-header will not scroll.  -->
+    <div class="sidebar-header">
+        <div class="row">
+            <div class="column-8">
+                <h2 class="sidebar-title">{{placeDetail.place.label}}</h2>
+                <h4 class="sidebar-title">{{placeDetail.place.state_abbrev}}</h4>
+                <div class="location-timestamp"><span>Last updated:</span>
+                    {{placeDetail.place.modifiedAt | date:'MMMM dd, yyyy'}}</div>
+            </div>
+            <div class="column text-right">
+                <a class="btn btn-primary" ui-sref="places.list">Back to search</a>
+            </div>
+        </div>
+    </div>
+
+    <!-- Everythign in sidebar-scrollable will scroll. -->
+    <div class="sidebar-scrollable">
+        <section>
+            <div class="metric-details">
+                <div class="row" ng-if="placeDetail.lastJobScore">
+                    <div class="column">
+                        <label class="network-score large flex-row columns-center">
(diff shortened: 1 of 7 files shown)


CloudNiner:
Let's start by putting the following downloads into the download dropdown:
- `ways_url`, title: 'Neighborhood Ways (shp)'
- `census_blocks_url`, title: 'Census Blocks (shp)'
- `connected_census_blocks_url`, title: 'Connected Census Blocks (csv)'
- `overall_scores_url`, title: 'Overall Scores (csv)'

src/angularjs/src/app/places/detail/places-detail.html
```diff
@@ -0,0 +1,76 @@
...
+        <section>
+            <div class="metric-details">
+                <div class="row" ng-if="placeDetail.lastJobScore">
+                    <div class="column">
+                        <label class="network-score large flex-row columns-center">
+                            {{placeDetail.lastJobScore | number:0}}
+                            <span class="h3">Network Score</span>
+
+                            <!-- Tooltips how it works: If you have a div with class .tooltip add a title. The content of the title will display in the toolip using css. -->
+                            <div class="tooltip" title="This is a tooltip that keeps sdfjhskdjfh dfskjhas kjhdfsa kjdfsah sdfakjh kjfdsah "><i class="icon-info-circled"></i></div>
```
CloudNiner: Maybe tooltips are a separate task? We should probably check in with @designmatty and @jcahail to see if there's anything they had in mind for the tooltips at this point

src/angularjs/src/app/places/list/place-list.html
```diff
@@ -10,20 +10,32 @@ <h2 class="sidebar-title">City/Town Search</h2>
...
+                    <label for="neighborhood-filter">Filter by city or town name</label>
+                    <ui-select
+                        name="neighborhood-filter"
+                        ng-model="placeList.neighborhoodFilter"
+                        title="Filter by Neighborhood">
+
+                        <ui-select-match allow-clear="true" placeholder="Neighborhood" >
+                            {{$select.selected.label}}
+                        </ui-select-match>
+                        <ui-select-choices repeat="neighborhood.uuid as neighborhood in
```
CloudNiner: I wonder if a select dropdown in this style will get super unwieldy when there are 300+ neighborhoods in this list? @designmatty should we just revert to a standard dropdown that only shows matching results once the user has started to type?

src/angularjs/src/app/places/list/place-list.html
```diff
@@ -69,8 +81,7 @@ <h2 class="sidebar-title">City/Town Search</h2>
         <section>
```
CloudNiner: Looks like the section headers aren't implemented yet, I made a separate issue for that: #326

src/angularjs/src/app/places/list/places-list.controller.js
```diff
@@ -34,13 +41,40 @@
...
+            ctl.neighborhoodFilter = null;
+
+            ctl.sortBy = sortingOptions[0]; // default to alphabetical order
+            ctl.sortingOptions = sortingOptions;
+
+            ctl.getPlaces = getPlaces;
 
             getPlaces();
+            loadOptions();
+            $scope.$watch(function(){return ctl.neighborhoodFilter;}, filterNeighborhood);
```
CloudNiner: This watcher should probably be manually cleaned up in the `$scope.destroy` method

src/angularjs/src/app/places/list/places-list.controller.js
```diff
@@ -14,6 +14,13 @@
                                  Neighborhood, AnalysisJob) {
         var ctl = this;
 
+        var sortingOptions = [
+            {value: 'neighborhood__label', label: 'Alphabetical'},
```
CloudNiner: @jcahail the scope changes doc states: "The examples on the wireframe are by default sorted by city in alphabetical order; CHANGE to sort alphabetically by state". So we sort first by state, then secondary sort by city name?

src/angularjs/src/app/places/list/place-list.html
designmatty: Like a typeahead? I agree that a standard dropdown will become very difficult to use once we have that many options.

flibbertigibbet: Hm, okay. This is basically the same dropdown as used by the admin portion of the site. For now at least, the list is limited to neighborhoods owned/uploaded by the user's organization, although for this public page, the set of neighborhoods available should be different.

src/angularjs/src/app/places/list/places-list.controller.js
flibbertigibbet: There's likely a number of filters in the admin portion of the site that need that cleanup as well then.

src/angularjs/src/app/places/list/place-list.html
CloudNiner: Yeah, if there's going to be a large list here, implementing this name filter as a standard typeahead instead of this ui-select filter might be better.

We need to address the permissions of what neighborhoods are returned from the request used here separately, that will have to wait until we revamp the permissions next sprint, so that doesn't need to be done now.

src/angularjs/src/app/places/list/places-list.controller.js
flibbertigibbet: According to [this discussion](http://stackoverflow.com/questions/25113884/should-angular-watch-be-removed-when-scope-destroyed), Angular should be cleaning up watchers on its own during the scope destruction.

src/angularjs/src/app/places/list/places-list.controller.js
flibbertigibbet: If so, we'll need to modify the back-end filter to support that.

flibbertigibbet:
Added download options to dropdown. In fixing drop-down to make it function [using angular-ui bootstrap](http://angular-ui.github.io/bootstrap/versioned-docs/1.3.3/), some styling went awry @designmatty:

image

src/angularjs/src/app/places/list/places-list.controller.js
CloudNiner: Indeed, and can be a separate task.

flibbertigibbet:
Opened #327 and #328 for the sort order modifications and typeahead implementation.

designmatty:
Don't stress the minor styling issues. Just make some issues for me and I'll take care of them.

PR #332, opened 2017-04-20
Bump terraform version to 0.9.3

tnation14:
## Overview

Bump terraform version to 0.9.3, which removes the `lock` configuration parameter and uses `role` instead of `roles` in the `aws_iam_instance_profile` resource.

### Demo

n/a

### Notes

Terraform also adds the `treat_missing_data` option to work with a new [CloudWatch feature](http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data). The default of `missing` seemed appropriate so I left it as is.

## Testing Instructions

 * Ensure that you can run `scripts/infra plan` according to the directions in the `README`

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -16,4 +16,4 @@ docker_options: "--storage-opt dm.basesize=20G"
 docker_compose_version: 1.9.0
 
 # azavea.terraform
-terraform_version: 0.9.1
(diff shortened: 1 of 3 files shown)


hectcastro:
Is there anything blocking this?

tnation14:
I think this needs to be merged  at the same time as azavea/cicero-live#139

Issue #305, opened 2017-04-18
Fix styling in admin views


KlaasH:
Removing all the old styles in #299 broke some things in the admin views, most glaringly the tables:

![image](https://cloud.githubusercontent.com/assets/6598836/25138292/25240e0a-2428-11e7-90eb-a217e27b49f0.png)

Hopefully it's just a matter of restoring a file or two from the old styles, but I wasn't immediately able to figure out which ones, didn't want to make a mess of the styles by including a bunch of cruft, and didn't want to hold up development of the public site.  So it needs another pass.

PR #346, opened 2017-04-24
Feature/kak/typeahead search#328

### LINKED ISSUE #328: Switch to typeahead for text search fields
Instead of using `ui-select-match` with a dropdown, use typeahead for search boxes. (Available in [angular-ui bootstrap](http://angular-ui.github.io/bootstrap/versioned-docs/1.3.3/).

See discussion on #324.

### LINKED ISSUE #329: Styling broken by angular-ui dropdown
Modifying dropdown HTML to get them to function with angular-ui bootstrap library breaks some styling: button no longer flush right, and list items appear below some other content.

See discussion on #324.

flibbertigibbet:
## Overview

Banish `ui.select` control from Angular app. Use Bootstrap typeahead instead. The select control offers a scrolling dropdown that can grow quite large; the typeahead control gets limited to the first eight matches by default.

### Demo

Also modified neighborhood filter control to include the state in the autocomplete label:
reworked_dropdown

### Notes

Note #305 is still outstanding (admin styling is broken).

## Testing Instructions

 * Rebuild angular container
 * Try out filters for admin analysis jobs: http://localhost:9301/#/admin/analysis-jobs/
 * Also for user management: http://localhost:9301/#/admin/users/
 * Also also for public 'places' list: http://localhost:9301/#/places/

Closes #328
Incidentally closes #329 (by removing the control)

### DIFF
--- a/src/angularjs/bower.json
+++ b/src/angularjs/bower.json
@@ -20,7 +20,6 @@
     "moment": "2.17.1",
     "animate.css": "3.5.2",
     "angular-bootstrap": "1.3.3",
-    "ui-select": "angular-ui-select#^0.19.5",
     "urijs": "1.18.7",
     "lodash": "4.6.1",
     "ng-file-upload": "^12.2.13",
--- a/src/angularjs/src/app/components/filters/analysis-job-filter.html
+++ b/src/angularjs/src/app/components/filters/analysis-job-filter.html
@@ -2,27 +2,23 @@
     <th>ID</th>
     <th>
         <div class="dropdown">
-            <ui-select ng-model="ctl.statusFilter" title="Filter by Status">
-                <ui-select-match allow-clear="true" placeholder="Status" >
-                    {{$select.selected}}
-                </ui-select-match>
-                <ui-select-choices repeat="status in ctl.statuses | filter: $select.search">
-                    <div ng-bind-html="status | highlight: $select.search"></div>
-                </ui-select-choices>
-            </ui-select>
+        <input name="neighborhood-filter" ng-model="ctl.statusFilter" title="Filter by Status"
+            placeholder="Status" type="text" class="form-control"
+            uib-typeahead="status for status in ctl.statuses |
+                           filter: $viewValue |
+                           limitTo: 8">
         </div>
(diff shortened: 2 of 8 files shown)


src/angularjs/bower.json
```diff
@@ -20,7 +20,6 @@
     "moment": "2.17.1",
     "animate.css": "3.5.2",
     "angular-bootstrap": "1.3.3",
-    "ui-select": "angular-ui-select#^0.19.5",
```
KlaasH: :tada:

src/angularjs/src/app/components/filters/analysis-job-filter.html
```diff
@@ -2,27 +2,23 @@
...
-                                           ctl.neighborhoods | filter: {uuid : $select.search}">
-                    <div ng-bind-html="neighborhood.label | highlight: $select.search"></div>
-                </ui-select-choices>
-            </ui-select>
+            <input title="Filter by Neighborhood" placeholder="Neighborhood"
+                ng-model="ctl.neighborhoodFilter"
+                type="text" class="form-control"
+                uib-typeahead="neighborhood as (neighborhood.label + ', ' +
+                               neighborhood.state_abbrev) for neighborhood in ctl.neighborhoods |
+                               filter: {label : $viewValue} |
```
KlaasH: Do we want to include state in this filter, such that typing e.g. "PA" would get you everything in PA?  I forget how to do that with `uib-typeahead`, but I believe I did it once.  (If so, also on the public site).

flibbertigibbet: I tried adding state to the filter search (both to the previous control and to this typeahead) but it wasn't do-able from within the template. I believe it would require munging the values in the controller to make them into a single field to pass to the control. No, the state was not in the filter search previously (or in the displayed label). I believe @jcahail said it's fine if we just filter on label (and not state).

KlaasH: Ok yeah, I didn't find the example I thought I remembered.  Just fine as is.

Issue #337, opened 2017-04-20
Implement thumbnail maps


flibbertigibbet:
To be displayed by city cards on home page and by each place in places list.

flibbertigibbet:
Existing map directive doesn't seem to work well with multiple instances on a page.

flibbertigibbet:
Going to call this done with #337. Loading the shape outline should be done with #357.

PR #347, opened 2017-04-25
Feature/kak/compare page#253

### LINKED ISSUE #253: Implement compare page
- [ ] Neighborhood cards
- [ ] Placeholder when n<3

Add/remove is issue #252.

flibbertigibbet:
## Overview

Adds neighborhood comparison page.

### Demo

image

### Notes

The styling for the 'remove' button at the top of each column appears to be broken; it was in the DOM, but not visible. Added button classes to get it to show up, but the icon is missing and there's no text.

Does not implement the list page selection/navigation, which is #251.

Does not implement the map thumbnail card, which needs #337.

Does not implement the footer cards with the metric descriptions (have no descriptions to display yet).

## Testing Instructions

 * Navigate there directly with one or more neighborhood UUIDs: http://localhost:9301/#/compare///bab4bbb5-d209-49be-a577-0b8ecb00f9bd

Closes #253


src/angularjs/src/app/compare/compare.controller.js
```diff
@@ -0,0 +1,84 @@
...
+        function clearSelection(num) {
+            var path = '/compare/';
+            if (num === 0) {
+                path += '/' + $stateParams.place2 + '/' + $stateParams.place3;
+            } else if (num === 1) {
+                path += $stateParams.place1 + '//' + $stateParams.place3;
+            } else {
+                path += $stateParams.place1 + '/' + $stateParams.place2 + '/';
+            }
+            $location.path(path);
```
KlaasH: I think this would be better as a state transition rather than a `$location` update.  I believe
```
var newParams = _.extend({}, $stateParams);
newParams['place' + (num + 1)] = '';
$state.go('compare', newParams);
```
would do it.

src/angularjs/src/app/compare/compare.controller.js
```diff
@@ -0,0 +1,84 @@
...
+
+        function getPlace(num, uuid) {
+            if (!uuid) {
+                return;
+            }
+
+            ctl.places[num] = {};
+
+            Neighborhood.query({uuid: uuid}).$promise.then(function(data) {
+                ctl.places[num].neighborhood = new Neighborhood(data);
```
KlaasH: Filling the object in pieces makes it render in chunks.  I think it would be better to assign the whole thing to the controller at once.  (Switching to a shared service that returns the whole `place` would also imply setting it on the controller all at once.)

flibbertigibbet: I agree it looks odd to see the columns load separately, split seconds apart. Then again, it seems to generally be preferred web design to present content sooner, piecemeal, than to leave the user looking at an empty page. @CloudNiner maybe have thoughts on how best to load here?

CloudNiner: Hmm, I don't feel strongly about this one either way. I might lean slightly towards displaying all three at the same time for two reasons:
1. Each individual piece of content is less valuable until all the content is presented
2. Loading each block individually makes the page load as a whole feel a bit janky as content pops in piecemeal

If one of the blocks fails to load, you could still show the remaining blocks, and have some type of error block in place of the ones that failed to load. I think that would look the same in either case.

Not to punt down the chain, but @designmatty may have a preference as this feels more of a UI/UX issue.

src/angularjs/src/app/compare/compare.controller.js
flibbertigibbet: That's so much better.

flibbertigibbet:
Okay, moved module under places (URL now like http://localhost:9301/#/places/compare//bab4bbb5-d209-49be-a577-0b8ecb00f9bd//). Delay results display until all loaded. Also added thumbnail map, since that's gotten merged.

src/angularjs/src/app/compare/compare.controller.js
designmatty: This is definitely one of those very edge-case situations which doesn't have a right or wrong solution. I think I would prefer to show them once all have loaded, and we could potentially show the user some loading spinner while they are loading.

src/angularjs/src/app/places/compare/compare.controller.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        function getPlaces(uuids) {
+            var promises = _.map(uuids, function(uuid, offset) {
+                return getPlace(offset, uuid);
+            });
+
+            // do not display any place until all places have been retrieved
+            $q.all(promises).then(function(results) {
+                ctl.places = results;
```
KlaasH: It was mainly the individual cards loading in chunks that looked bad to me, so loading each place into `ctl.places[offset]` as it arrives would be fine as well, I think.  Not sure which is better, but the difference is minor and this is here, so this is good.

flibbertigibbet: Delaying display in any case would require knowing when all promises have either returned or failed, which as far as I can see would still require a promises-aware refactor, which is why I went with this.

Issue #350, opened 2017-04-25
Make results path in storage bucket public


flibbertigibbet:
To support pulling tiles and GeoJSON layers into Angular app maps.

Issue #349, opened 2017-04-25
Support CORS for S3 resources


flibbertigibbet:
The GeoJSON URLs provided in the job results endpoint under `destinations_urls` need to support CORS in order for them to be retrieved client-side for map display.

https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#using-cors

flibbertigibbet:
Closed in #354.

PR #352, opened 2017-04-25
Feature/place detail map#255

### LINKED ISSUE #255: Implement map on neighborhood detail page
Depends #254.

flibbertigibbet:
## Overview

Adds map to public places detail view with a control containing the results GeoJSON layers.

### Demo

A popup:
pfb_university_map_popup

### Notes

Things not done:
 - Display neighborhood bounds polygon, as there is currently no endpoint for that (will we add one?)
 - Properly zoom/fit map to neighborhood area. If we add bounds polygon, can be accomplished with that
 - Style results markers/popups. At the least, should vary marker color by layer to distinguish
 - Display results tile layer (depends on #350).

At least for my test run for Philly, many of the results were missing values for low/high stress and pop. score; there were some for some schools in the 'university' layer (as in popup pictured above).

## Testing Instructions

 * As a workaround for #349, go to your S3 results bucket and enable CORS
 * Visit places detail page
 * Should get a map with a layer control in upper right, with a layer for each metric
 * Should be able to turn layers on and off
 * Should have GeoJSON properties in layer marker popups

Closes #255

### DIFF
--- a/src/angularjs/src/app/places/map/places-map.directive.js
+++ b/src/angularjs/src/app/places/map/places-map.directive.js
@@ -1,9 +1,10 @@
 (function() {
 
     /* @ngInject */
-    function PlacesMapController($log, $scope) {
+    function PlacesMapController($filter, $http, $sanitize, $scope) {
         var ctl = this;
         ctl.map = null;
+        ctl.layerControl = null;
 
         ctl.$onInit = function () {
             ctl.mapOptions = {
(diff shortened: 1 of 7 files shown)


src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,121 @@
...
+                return $sanitize(snippet);
+            }
+        }
+    }
+
+    function PlacesMapDirective() {
+        var module = {
+            restrict: 'E',
+            scope: {
+                pfbPlacesMapLayers: '='
```
CloudNiner: Probably want a one-way binding here instead

src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,121 @@
...
+        var ctl = this;
+        ctl.map = null;
+        ctl.layerControl = null;
+
+        ctl.$onInit = function () {
+            ctl.mapOptions = {
+                scrollWheelZoom: true
+            };
+
+            // TODO: set center and zoom level by zooming to fit geojson polygon bounds
```
CloudNiner: Since we're now pulling actual geojson, we should probably take care of this TODO.

src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,121 @@
...
+            ctl.mapCenter = [39.963277, -75.142971];
+            ctl.baselayer = L.tileLayer(
+                'https://stamen-tiles.a.ssl.fastly.net/toner-lite/{z}/{x}/{y}.png', {
+                    attribution: 'Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.',
+                    maxZoom: 18
+                });
+
+            $scope.$watch(function(){return ctl.pfbPlacesMapLayers;}, setLayers);
```
CloudNiner: Rather than setup a `$watch` here you should just be able to implement `ctl.$onChanges(changes)` if you switch pfbPlacesMapLayers to a one-way binding.

https://docs.angularjs.org/api/ng/service/$compile#life-cycle-hooks

src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,121 @@
...
+                            ctl.map.openPopup(popup);
+                        }
+                    }
+                });
+            }
+
+            /**
+             Convert GeoJSON properties into HTML snippet for Leaflet popup display.
+             */
+            function buildLabel(properties) {
```
CloudNiner: Ok for now, depending on how fancy the popup gets we might want to switch to an actual ng template. We have plenty of examples of this in other projects.

src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                    return ['<li>',
+                            key.replace(/_/g, ' '),
+                            ': ',
+                            (val ? val : '--'),
+                            '</li>'
+                            ].join('');
+                }).join('');
+                snippet += '</ul>';
+                return $sanitize(snippet);
```
CloudNiner: 👍

src/angularjs/src/app/places/map/places-map.directive.js
flibbertigibbet: My idea is to fix this up when adding pull of neighborhood bounds to zoom to fit that single area, instead of trying to zoom to fit all 13 results bounds.

src/angularjs/src/app/places/map/places-map.directive.js
flibbertigibbet: Agreed. I imagine we'll want to change up the contents of this considerably.

src/angularjs/src/app/places/map/places-map.directive.js
flibbertigibbet: Created #357 for that

flibbertigibbet:
Okay, if it makes sens to go with #357 handling zoom/bounds, I think this should be good to go now.

src/angularjs/src/app/places/map/places-map.directive.js
```diff
@@ -0,0 +1,127 @@
...
+            ctl.baselayer = L.tileLayer(
+                'https://stamen-tiles.a.ssl.fastly.net/toner-lite/{z}/{x}/{y}.png', {
+                    attribution: 'Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.',
+                    maxZoom: 18
+                });
+        };
+
+        ctl.$onChanges = function(changes) {
+            // set map layers once received from parent scope (paret-detail.controller)
+            if (changes.pfbPlacesMapLayers.currentValue && ctl.map) {
```
CloudNiner: Should check for `changes.pfbPlacesMapLayers` before checking for `changes.pfbPlacesMapLayers.currentValue`, this would currently throw an error if key `pfbPlacesMapLayers` was not in the list of changes for a given digest cycle

PR #354, opened 2017-04-25
Make results bucket public

### LINKED ISSUE #350: Make results path in storage bucket public

### LINKED ISSUE #353: Add endpoint for simplified neighborhood bounds
Return neighborhood bounds as simplified GeoJSON, for use in client-side maps.

KlaasH:
## Overview

Makes the results bucket public and adds CORS permissions so that the feature geojson files and the tile layers can be accessed by the front-end.

### Notes

Even without of setting up CloudFront we might be able to get more fine-grained with the permissions policy, but for now this does the job.

## Testing Instructions

I applied this to the staging site, so any file in the `staging-pfb-storage-us-east-1` bucket should now be accessible at the key path appended to https://s3.amazonaws.com/staging-pfb-storage-us-east-1/.

The static tile layer URLs would be `https://s3.amazonaws.com/staging-pfb-storage-us-east-1/results/<JOB_ID>/tiles/<LAYER_NAME>/{z}/{x}/{y}.png`, where `LAYER_NAME` is "neighborhood_ways" or "neighborhood_census_blocks".

Resolves #353, resolves #350.

### DIFF
--- a/deployment/terraform/storage.tf
+++ b/deployment/terraform/storage.tf
@@ -17,7 +17,16 @@ resource "aws_s3_bucket" "static" {
 
 resource "aws_s3_bucket" "storage" {
   bucket = "${lower("${var.environment}")}-pfb-storage-${var.aws_region}"
-  acl    = "private"
+  acl    = "public-read"
+  policy = "${data.aws_iam_policy_document.anonymous_read_storage_bucket_policy.json}"
+
+  cors_rule {
+    allowed_origins = ["*"]
+    allowed_methods = ["GET", "HEAD"]
+    max_age_seconds = 3000
+    allowed_headers = ["Authorization"]
+    expose_headers  = ["ETag"]
+  }
 
   tags {
     Environment = "${var.environment}"
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -58,6 +58,27 @@ data "aws_iam_policy_document" "batch_manage_jobs" {
   }
 }
 
+data "aws_iam_policy_document" "anonymous_read_storage_bucket_policy" {
+  policy_id = "S3StorageAnonymousReadPolicy"
+
+  statement {
+    sid = "S3ReadOnly"
+
+    effect = "Allow"
+
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    actions = ["s3:GetObject"]
+
+    resources = [
+      "arn:aws:s3:::${lower("${var.environment}")}-pfb-storage-${var.aws_region}/*",
+    ]
+  }
+}
+
 #
 # Custom policies
 #


deployment/terraform/storage.tf
```diff
@@ -17,7 +17,29 @@ resource "aws_s3_bucket" "static" {
 
 resource "aws_s3_bucket" "storage" {
   bucket = "${lower("${var.environment}")}-pfb-storage-${var.aws_region}"
-  acl    = "private"
+  acl    = "public-read"
+  policy = <<POLICY
```
hectcastro: There is an `aws_iam_policy_document` resource that can be used to capture policy definitions:

```hcl
data "aws_iam_policy_document" "read_only_bucket_policy" {
  policy_id = "S3AnonymousReadOnlyPolicy"

  statement {
    sid = "S3ReadOnly"

    effect = "Allow"

    principals {
      type        = "AWS"
      identifiers = ["*"]
    }

    actions = ["s3:GetObject"]

    resources = [
      "arn:aws:s3:::${var.bucket_name}/*",
    ]
  }
}
```

Then `data.aws_iam_policy_document.read_only_bucket_policy.json` can be used to render it where needed.

KlaasH: Cool.  But I'm not quite getting how to then attach the policy to the resource in a way that supplies the `var.bucket_name` value.

Taking out the `resources` clause from the policy_document definition then doing 
```
resource "aws_s3_bucket_policy" "storage_anonymous_read" {
  bucket = "${aws_s3_bucket.storage.id}"
  policy = "${data.aws_iam_policy_document.anonymous_read_bucket_policy.json}"
}
```
seems like it might work (terraform successfully builds a plan and the changes look reasonable), but I don't know if that's correct and in good order.

hectcastro: `var.bucket_name` was something that came along for the ride from the example I pulled. I think that it can safely be replaced with `arn:aws:s3:::${lower("${var.environment}")}-pfb-storage-${var.aws_region}/*` here.

Associating it with the bucket would be something like:

```hcl
resource "aws_s3_bucket" "storage" {
  bucket = "${lower("${var.environment}")}-pfb-storage-${var.aws_region}"
  policy = "${data.aws_iam_policy_document.read_only_bucket_policy.json}"

  ...
}
```

First time seeing `aws_s3_bucket_policy`, actually. I'd lean toward rending the policy JSON directly into the `policy` parameter of `aws_s3_bucket`.

KlaasH: Ok, set it up with the policy defined in `iam.tf` with the bucket name composed as above and attached in the bucket definition.

plan now says:
```
~ aws_s3_bucket.storage
    acl:                           "private" => "public-read"
    cors_rule.#:                   "0" => "1"
    cors_rule.0.allowed_headers.#: "0" => "1"
    cors_rule.0.allowed_headers.0: "" => "Authorization"
    cors_rule.0.allowed_methods.#: "0" => "1"
    cors_rule.0.allowed_methods.0: "" => "GET"
    cors_rule.0.allowed_origins.#: "0" => "1"
    cors_rule.0.allowed_origins.0: "" => "*"
    cors_rule.0.expose_headers.#:  "0" => "1"
    cors_rule.0.expose_headers.0:  "" => "ETag"
    cors_rule.0.max_age_seconds:   "" => "3000"
    policy:                        "" => "{\n  \"Version\": \"2012-10-17\",\n  \"Id\": \"S3StorageAnonymousRea
dPolicy\",\n  \"Statement\": [\n    {\n      \"Sid\": \"S3ReadOnly\",\n      \"Effect\": \"Allow\",\n      \"A
ction\": \"s3:GetObject\",\n      \"Resource\": \"arn:aws:s3:::staging-pfb-storage-us-east-1/*\",\n      \"Pri
ncipal\": \"*\"\n    }\n  ]\n}"
```

deployment/terraform/storage.tf
```diff
@@ -17,7 +17,16 @@ resource "aws_s3_bucket" "static" {
 
 resource "aws_s3_bucket" "storage" {
   bucket = "${lower("${var.environment}")}-pfb-storage-${var.aws_region}"
-  acl    = "private"
+  acl    = "public-read"
+  policy = "${data.aws_iam_policy_document.anonymous_read_storage_bucket_policy.json}"
+
+  cors_rule {
+    allowed_origins = ["*"]
+    allowed_methods = ["GET"]
```
CloudNiner: Might want to allow HEAD here as well. I'm not sure its strictly necessary, but it is classified as a read operation.

flibbertigibbet:
Closes #349 (not #353).

Issue #343, opened 2017-04-24
Multiple analysis jobs on a single AWS EC2 instance cannot be run concurrently


CloudNiner:
All jobs attempt to bind to the host postgresql port 5432. An attempt to fix this by changing `listen_addresses` and only connect locally via unix socket was made here:

https://github.com/azavea/pfb-network-connectivity/compare/feature/awf/analysis-remove-host

But this resulted in the same error when multiple jobs were run.

As a quick fix to get the BigJump neighborhoods run, the production Batch cluster autoscaling group was reverted via terraform to `i3.large` instances which can only run a single job at a time, instead of `i3.2xlarge` instances which can run four jobs.

CloudWatch log from a failed job displaying the error:
https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=/aws/batch/job;stream=production-pfb-analysis-run-jo--17--f1346347/80286034-51c1-49c5-a5c7-4bc0724890f2/4f78993e-747f-4b28-879f-32f4851a5c60

Relevant section from the above log extracted:
```
18:44:55
waiting for server to start....LOG: could not bind IPv4 socket: Address already in use

18:44:55
HINT: Is another postmaster already running on port 5432? If not, wait a few seconds and retry.

18:44:55
WARNING: could not create listen socket for "localhost"

18:44:55
FATAL: could not create any TCP/IP sockets

18:44:55
LOG: database system is shut down

18:44:56
stopped waiting

18:44:56
pg_ctl: could not start server

18:44:56
Examine the log output.

18:44:57
waiting for database, try 2

18:45:00
waiting for database, try 3

18:45:03
waiting for database, try 4

18:45:06
waiting for database, try 5

18:45:09
waiting for database, try 6

18:45:12
waiting for database, try 7

18:45:15
waiting for database, try 8

18:45:18
waiting for database, try 9

18:45:21
waiting for database, try 10

18:45:24
waiting for database, try 11

18:45:27
waiting for database, try 12

18:45:30
waiting for database, try 13

18:45:33
waiting for database, try 14

18:45:36
waiting for database, try 15

18:45:39
waiting for database, try 16

18:45:42
waiting for database, try 17

18:45:45
waiting for database, try 18

18:45:48
waiting for database, try 19

18:45:51
waiting for database, try 20

18:45:54
waiting for database, try 21

18:45:57
waiting for database, try 22

18:46:00
waiting for database, try 23

18:46:03
waiting for database, try 24

[...]

PR #355, opened 2017-04-26
Split Batch task definitions by environment

CloudNiner:
## Overview

Due to #343 we need to update the production stack to only run a single analysis job at a time on the larger instance type, `i3.2xlarge`. As a result, the aws batch job definition json files were split by environment so that all non-deployment specific configuration remains in the json definition files.

### Notes

Removes the `--environment` flag of update_job_definition.py

./scripts/infra now also prints the newly created job definition name:revision to the console

The new production settings are untested. We will be running a small Batch later today to test before running the remaining 100/300 boundary batches.

## Testing Instructions

I deployed this branch to staging and successfully ran a Germantown job: https://staging.pfb.azavea.com/#/admin/analysis-jobs/a4d33bba-7db9-4936-9b52-1bab8be95dff/


deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
```diff
@@ -0,0 +1,45 @@
+{
+    "jobDefinitionName": "production-pfb-analysis-run-job",
+    "type": "container",
+    "parameters": {},
+    "containerProperties": {
+        "image": "{image}",
+        "vcpus": 8,
```
CloudNiner: We could very safely drop this to 7 vcpus so that one CPU and enough mem remains available for instances to concurrently process tiling jobs as well. Thoughts?

deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
```diff
@@ -0,0 +1,45 @@
+{
+    "jobDefinitionName": "production-pfb-analysis-run-job",
```
CloudNiner: The production version bumps CPU/mem, as well as the postgresql settings in the environment section below, following the comments for each variable that are in setup_database.sh. I tended to err on the conservative side.

CloudNiner:
Once this is merged and deployed to production, we'll be ready to start the first round of batch jobs.

deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
KlaasH: Yeah, I think that's a good idea.  Aside from the tiling job using that one CPU more efficiently, this would also reduce the possibility of a single tiling job tying up a whole instance (which I think is a possibility).

CloudNiner: Great, I'll update to 7vpcus then.

PR #356, opened 2017-04-26
Feature/kak/return simplified bounds#353

### LINKED ISSUE #353: Add endpoint for simplified neighborhood bounds

flibbertigibbet:
## Overview

Add endpoint for simplified neighborhood bounds, for use in front end maps.

### Demo

With 22 neighborhoods before simplification, endpoint returned 301 KB in 3.6s.
After simplification, endpoint returned 40.3 KB in 660ms.

### Notes

Error handling added that falls back on no simplification in case broken geometries result. In testing, none of the boundaries failed to simplify.

## Testing Instructions

 * Run migrations (`./scripts/django-manage migrate`)
 * Endpoint is at http://localhost:9200/api/neighborhoods_bounds_geojson/

Closes #353


flibbertigibbet:
I'd considered adding a fallback simplification level, but since the existing one worked fine for the test files at hand I decided it might be over-complicating things.

Regarding the empty boundary after simplification, I think that may have been fixed in GEOS December, in closing [this issue](https://trac.osgeo.org/geos/ticket/741). I'll see if it would be easy for us to upgrade GEOS.

flibbertigibbet:
Apparently GEOS is pulled as a dependency of `gdal-bin`. Container reports from `dpkg -s libgeos-c1`
version 3.4.2-6.

flibbertigibbet:
Adding a fallback to 0.0001 simplification works for LA, and results in 1,414 instead of the original 7,949 `num_points` reported.

flibbertigibbet:
Okay, I think the simplification fallbacks just pushed should sort it out.

Issue #359, opened 2017-04-27
clip_osm.sql throws error because :nb_boundary_buffer is unset on AWS


CloudNiner:
Looks like we never pass in NB_MAX_TRIP_DISTANCE via the json job config or Django. Which is then used to indirectly set the default for NB_BOUNDARY_BUFFER

PR #363, opened 2017-04-27
Feature/awf/analysis mount tmp to ephemeral

### LINKED ISSUE #358: Batch Jobs failed due to docker volume running out of disk space
This is likely because we perform a bunch of I/O operations in directories created by `mktemp -d` which currently mount to the docker volume and not to `/tmp`.

CloudNiner:
## Overview

Updates the analysis job configuration to mount `/tmp` to the NVME scratch space drives on AWS.

Leaving /tmp within the containers leads to storage issues since larger jobs use many GB of files which rapidly eats into the available space on an instance's docker drive pool.

Instead we mount /tmp to a volume attached to the host scratch space, which has plenty of room for our tmp files. 

### Demo

Note /tmp file paths in screenshot, now contain job id and subdirectories:
screen shot 2017-04-27 at 13 35 46

screen shot 2017-04-27 at 13 43 27

Fix for #359:
screen shot 2017-04-27 at 13 44 18

Staging disk usage not exploding when running LA:
screen shot 2017-04-27 at 14 03 15

### Notes

I originally only mounted `/tmp` leaving all other file paths in the analysis jobs the same. But it seems dangerous to have potentially multiple jobs mounting and using the same available space. So, I updated analysis to be able to take a `NB_TEMPDIR` option which specifies the temp directory to use for all transient file operations. This allows the Django application to properly namespace the directories mounted to the job containers. The alternative would be to modify the volume/mountpoint paths when we call Batch's SubmitJob, but there is no option to do that, those settings can only be modified in JobDefinition.

## Testing Instructions

If desired, can test changes locally by setting `NB_TEMPDIR="/tmp/${PFB_JOB_ID}"` in `./scripts/run-local-analysis` and then passing in `NB_TEMPDIR` by adding an extra env line in the docker run command: `-e NB_TEMPDIR=${NB_TEMPDIR}` before triggering a local job. You should see the tmp dir created and the printed file paths change appropriately. 

Alternatively, staging LA job has now progressed past where the production one failed due to filling up the root volume (see above screenshot).

Closes #358 #359


deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
```diff
@@ -15,8 +15,16 @@
...
             }
+        }, {
+            "name": "tmp",
+            "host": {
+                "sourcePath": "/media/nvme0n1/tmp"
+            }
         }],
         "environment": [{
+            "name": "NB_MAX_TRIP_DISTANCE",
+            "value": "2680"
```
KlaasH: This seems like the wrong place for this.  I think it should be set inside `import_osm.sh`, i.e. right before [the line setting the default for NB_BOUNDARY_BUFFER to equal NB_MAX_TRIP_DISTANCE](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/import/import_osm.sh#L15) we should have a line setting the default value for NB_MAX_TRIP_DISTANCE (like [in import_neighborhood.sh](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/import/import_neighborhood.sh#L5))
```
NB_MAX_TRIP_DISTANCE="${NB_MAX_TRIP_DISTANCE:-2680}"
```

CloudNiner: After looking at this again, I disagree. Both NB_MAX_TRIP_DISTANCE and NB_BOUNDARY_BUFFER are used in multiple files, and the value of those variables should be consistent in all files. 

What if instead I moved the declarations of these two variables to run_analysis.sh and `export`ed them so they're available everywhere else, removing their definitions from each of the individual scripts they're used in?

KlaasH: Yeah, that would work as well.  But the setup I'm proposing wouldn't cause inconsistencies, I don't think.  
```
NB_MAX_TRIP_DISTANCE="${NB_MAX_TRIP_DISTANCE:-2680}"
NB_BOUNDARY_BUFFER="${NB_BOUNDARY_BUFFER:-$NB_MAX_TRIP_DISTANCE}"
```
would mean "set `NB_MAX_TRIP_DISTANCE` to the value in the environment, but fall back to 2680, then set `NB_BOUNDARY_BUFFER` to the value in the environment but fall back to whatever `NB_MAX_TRIP_DISTANCE` is set to."  So if either is already set in the environment that value will be used.

However, having lots of different component scripts do their own setup like this probably doesn't make sense.  It seems like its purpose would be to allow them to operate independently, but that's not something we need and it probably wouldn't work out that well anyway given all the development that has taken place treating it as a whole.

So yeah, defining variables further out and `export`ing them sounds fine to me, it just doesn't fit the existing pattern as well.

CloudNiner: Ah, yes, wasn't implying your solution wouldn't work, because it would, and is definitely faster to implement. Was just suggesting that while I'm dabbling here its worth the extra few minutes to define these values farther up to make explicit an assumption we're currently making (that these values are the same across the analysis)

Issue #360, opened 2017-04-27
Static tile generation for larger boundaries takes a very long time


CloudNiner:
Many hours for larger boundaries

Investigate ways to utilize multiple cores/parallelization?

Split job so that each tile layer is generated in separate job. Can be done without adding an additional job definition.

KlaasH:
Resolved by #364.  If we want to restore zoomlevel 18 or make other changes, that can be a new issue.

PR #364, opened 2017-04-28
Parallelize tilemaker

KlaasH:
## Overview & Notes

Makes tilemaker much more parallel, using GNU `parallel`.

In addition to running one command per zoomlevel, it also divides the bounding box into segments and runs one command per segment.  The default I went with is 4.  Any tile that falls at all within a boundary gets generated, so more partitions would result in more tiles getting generated twice.  So I only used the partitioning at higher zoomlevels (cutoff of 13 chosen somewhat arbitrarily).

I also increased the resources limits in the job definition, since it should be able to make use of them now.

I didn't split ways and blocks into separate tasks.  It would be pretty easy to do so, but a little attention would have to be paid to how to keep the first one that finishes from updating the status to COMPLETE prematurely.

I'm out for the morning, so feel free to modify and merge this.  I'll be online in the afternoon.

## Testing Instructions

Build the tilemaker container and run a job (you can get the tiling command for your most analyzed-but-not-tiled job by running `AnalysisJobStatusUpdate.objects.filter(status=AnalysisJob.Status.EXPORTED).latest('timestamp').job.generate_tiles()` in shell_plus)
It should use more CPU and finish faster than before, but produce identical results.

Connects #360.

### DIFF
--- a/src/tilemaker/entrypoint.sh
+++ b/src/tilemaker/entrypoint.sh
@@ -6,7 +6,7 @@ set -e
 source /opt/pfb/tilemaker/scripts/utils.sh
 
 export TL_MIN_ZOOM="${TL_MIN_ZOOM:-8}"
-export TL_MAX_ZOOM="${TL_MAX_ZOOM:-18}"
+export TL_MAX_ZOOM="${TL_MAX_ZOOM:-17}"
 export TL_NUM_PARTS="${TL_NUM_PARTS:-4}"
 export TIME="\nTIMING: %C\nTIMING:\t%E elapsed %Kkb mem\n"
 
--- a/src/tilemaker/entrypoint.sh
+++ b/src/tilemaker/entrypoint.sh
@@ -7,6 +7,8 @@ source /opt/pfb/tilemaker/scripts/utils.sh
 
 export TL_MIN_ZOOM="${TL_MIN_ZOOM:-8}"
 export TL_MAX_ZOOM="${TL_MAX_ZOOM:-18}"
+export TL_NUM_PARTS="${TL_NUM_PARTS:-4}"
+export TIME="\nTIMING: %C\nTIMING:\t%E elapsed %Kkb mem\n"
 
 if [ -z "${PFB_JOB_ID}" ]; then
     echo "Error: PFB_JOB_ID is required"
--- a/src/tilemaker/scripts/run_tilemaker.sh
+++ b/src/tilemaker/scripts/run_tilemaker.sh
@@ -33,21 +33,67 @@ function generate_tiles_from_shapefile() {
 
     # Get bounds. ogrinfo can handle big files, but doesn't provide tons of output flexibility, so
     # parse out the answer from the blob of text it returns.
-    TL_BOUNDS=$(ogrinfo -so -al "/data/${TL_SHAPEFILE_NAME}.shp" \
+    export TL_BOUNDS=$(ogrinfo -so -al "/data/${TL_SHAPEFILE_NAME}.shp" \
         | grep Extent | sed -E 's/.*\((.*), (.*)\) - \((.*), (.*)\).*/\1 \2 \3 \4/')
 
     # Make tiles and upload them to S3.
-    # 'tl' only has totally-quiet or very-verbose (every tile) output options, so this filters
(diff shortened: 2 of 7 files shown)


src/tilemaker/scripts/run_tilemaker.sh
```diff
@@ -33,21 +33,67 @@ function generate_tiles_from_shapefile() {
...
-        "s3://${AWS_STORAGE_BUCKET_NAME}/${PFB_S3_TILES_PATH}/${TL_SHAPEFILE_NAME}/{z}/{x}/{y}.png" \
-        | awk 'NR == 1 || NR % 1000 == 0'
+    #
+    # The work is done in the two functions, which are spawned in many (relatively) bit-sized
+    # pieces with `parallel`
+
+    # Define these once to save space below. They need to be exported because they'll be used
+    # in the functions.
+    export TL_INPUT="mapnik:///opt/pfb/tilemaker/styles/${TL_SHAPEFILE_NAME}_style.xml"
+    export TL_OUTPUT="s3://${AWS_STORAGE_BUCKET_NAME}/${PFB_S3_TILES_PATH}/${TL_SHAPEFILE_NAME}/{z}/{x}/{y}.png"
```
CloudNiner: If we have multiple tilemaker jobs and each is running in parallel, I believe we'll easily hit the S3 PUT limits described here:

http://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html

We should probably open an support case with AWS to make them aware of potential burst usage over the next few weeks?

src/tilemaker/entrypoint.sh
```diff
@@ -7,6 +7,8 @@ source /opt/pfb/tilemaker/scripts/utils.sh
 
 export TL_MIN_ZOOM="${TL_MIN_ZOOM:-8}"
 export TL_MAX_ZOOM="${TL_MAX_ZOOM:-18}"
+export TL_NUM_PARTS="${TL_NUM_PARTS:-4}"
```
KlaasH: Yeah, probably so.  Then we can see what the timings look like for the 17 layer and think about whether it's practical to add another 4x that to do 18.

KlaasH: So it runs one `tl` command per partition per zoomlevel (for 13 and above. below that I judged there aren't enough tiles for it to be worth partitioning), so with max-zoom 17 that's 20 total jobs.  The higher-zoomlevel ones will still be significantly bigger than the lower-zoomlevel ones, though.  We could make the partition count scale with zoomlevel, either linearly to mitigate the difference somewhat or exponentially to make all the jobs about the same size.  Might make sense to partition in both directions in that case, too.  I just did one way for simplicity.

KlaasH: Or we could just set `vcpus` to 4 so that having different-sized jobs wouldn't cause idle CPU time because they'd always come in sets of 4 roughly-equal jobs.

CloudNiner: Yeah I think one-way partitioning to keep things simple at this point is fine. Seems like 4vcpus is a good middle ground?

src/tilemaker/scripts/run_tilemaker.sh
KlaasH: Yeah, I'm not sure how fast each thread will be making tiles in the production environment, but I would hope it's at least a few per second, and if we could end up with 20 jobs running 8 threads each, that's 160 threads, so anything more than 2/second would put us over the 300/s PUT/LIST/DELETE limit.  So a support case seems like a good idea.

KlaasH: Ok, I picked a tiled job arbitrarily and got Memphis, which saved its 70,000 tiles over the course of 78 minutes.  So that's 15/second from one thread.  (Also, so many tiles.  17 is seeming like a good choice unless/until we definitely need 18.)  So 15 4-threaded jobs running would produce about 900 PUTs per second.  If we ended up running a re-tiling job and had 20 instances with two jobs each, that could mean 2,400/s.  Interesting.

CloudNiner: Thanks for running through that, we should definitely open up the support case to be safe, I'll do that today or tomorrow.

PR #365, opened 2017-04-28
Cleanup Django S3 url signing

CloudNiner:
## Overview

A bit more fallout from running batches. In this case, it appears pre-singed URLs are failing if they sit in the AWS Batch queue for too long. The presigned url is set to expire after two weeks, but since the requests are generated on an EC2 instance, they contain an `x-amz-security-token` which contains the EC2 instance temporary credential used to sign the request, and this fails, presumably because the EC2 credentials are rotated by the time the signed url is used. In any case, in #354 we made the storage bucket public so this just switches to use unsigned urls for all objects that live in the storages bucket.

@KlaasH you might want to take a look as well, since this requires running a command to update your dev storage bucket permissions.

### Demo

screen shot 2017-04-28 at 09 48 02

## Testing Instructions

If you don't already have your local storage bucket created, in the VM run:
```
aws s3api create-bucket --bucket "${DEV_USER}-pfb-storage-us-east-1"
```
Then, to ensure the public read policy is attached to your bucket, run:
```
aws s3api put-bucket-policy --bucket "${DEV_USER}-pfb-storage-us-east-1" --policy "{\"Statement\":[{\"Effect\":\"Allow\",\"Principal\":\"*\",\"Action\":\"s3:GetObject\",\"Resource\":\"arn:aws:s3:::${DEV_USER}-pfb-storage-us-east-1/*\"}]}"
```

Afterwards, navigate to a Job Results page in the admin UI and inspect the resource URLs. They should work as before and not contain any signing information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ Run `./scripts/setup` to install project dependencies and prepare the developmen
 vagrant ssh
 ```
 
+Once in the VM, with your AWS credentials configured, run the following commands to configure your development S3 buckets:
+```
+aws s3api create-bucket --bucket "${DEV_USER}-pfb-storage-us-east-1"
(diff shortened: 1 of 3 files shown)


CloudNiner:
Closes #367

Issue #345, opened 2017-04-24
Change base map


KlaasH:
While looking at the tile styling, @designmatty and I observed that the somewhat-dark roads on Stamen Toner Lite made it harder to see the segment stress layer.  He suggested [Positron](https://leaflet-extras.github.io/leaflet-providers/preview/#filter=CartoDB.Positron) instead.  Not sure if that was a "let's use this" or "let's try this to see if it looks better" suggestion, but in any case it makes sense to switch to something with outline rather than solid roads.

designmatty:
I imagine Positron would be our best bet since it's a very subdued basemap and would push the focus on the stress layer. I have no issues with using it right away.

PR #368, opened 2017-04-28
Feature/kak/places list map#249

### LINKED ISSUE #249: Implement map on neighborhood list page

flibbertigibbet:
## Overview

Implement map view on public places list page and load layer with neighborhood points. Also added a map constants file for shared use by the map directives in setting up base layers. Moved existing places detail map under `detail` module, as it can't be readily shared with the list page map.

### Demo

With satellite layer selected:
image

### Notes

Also fixed an issue where the new `geom_simple` GeoJSON was loading in the default neighborhood API endpoint (that endpoint should have no geometries in it).

#251 should add overlay centered over map for a places comparison overview.

## Testing Instructions

 * Visit places page at http://localhost:9301/?#/places/
 * Should get points layer of places, like on home page
 * Should have a layer switcher with a secondary satellite base layer option

Closes #249

### DIFF
--- a/src/angularjs/src/app/components/map/map.constants.js
+++ b/src/angularjs/src/app/components/map/map.constants.js
@@ -21,7 +21,7 @@
         }
     },
     conusBounds: [[24.396308, -124.848974], [49.384358, -66.885444]],
-    conusMaxZoom: 18
+    conusMaxZoom: 17
   };
 
   angular
--- a/src/angularjs/src/app/components/map/map.constants.js
+++ b/src/angularjs/src/app/components/map/map.constants.js
@@ -3,6 +3,12 @@
 
   var config = {
     baseLayers: {
+        'Positron': {
+            url: 'https://cartodb-basemaps-{s}.global.ssl.fastly.net/light_all/{z}/{x}/{y}.png',
+            attribution: [
+                '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>, ',
+                '&copy; <a href="https://carto.com/attribution">CARTO</a>'].join('')
+        },
         'Stamen' : {
             url: 'https://stamen-tiles.a.ssl.fastly.net/toner-lite/{z}/{x}/{y}.png',
             attribution: ['Map tiles by <a href="http://stamen.com">Stamen Design</a>, ',
--- a/src/angularjs/src/app/components/map/map.constants.js
+++ b/src/angularjs/src/app/components/map/map.constants.js
@@ -1,0 +1,31 @@
+(function() {
+  'use strict';
+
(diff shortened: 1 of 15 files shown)


src/angularjs/src/app/components/map/map.constants.js
```diff
@@ -0,0 +1,31 @@
...
+            url: 'https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
+            attribution: [
+                '&copy; <a href="http://www.esri.com/">Esri</a> ',
+                'Source: Esri, DigitalGlobe, GeoEye, Earthstar Geographics, CNES/Airbus DS, USDA, USGS, ',
+                'AEX, Getmapping, Aerogrid, IGN, IGP, swisstopo, and the GIS User Community'
+            ].join('')
+        }
+    },
+    conusBounds: [[24.396308, -124.848974], [49.384358, -66.885444]],
+    conusMaxZoom: 18
```
CloudNiner: Can you drop this to 17 since that's the highest zoom we'll be generating via tiles?

See https://github.com/azavea/pfb-network-connectivity/pull/364#discussion_r113988746

src/angularjs/src/app/components/map/map.constants.js
```diff
@@ -0,0 +1,31 @@
+(function() {
+  'use strict';
+
+  var config = {
+    baseLayers: {
+        'Stamen' : {
```
CloudNiner: It should be trivial, can we address #345 here as well by switching to positron?

src/angularjs/src/app/places/list/places-list-map-directive.js
```diff
@@ -0,0 +1,80 @@
+(function() {
+
+    /* @ngInject */
+    function PlacesListMapController(MapConfig, Neighborhood) {
```
CloudNiner: Bummer. In looking at this and `neighborhood-map.directive.js` it seems like they're just different enough that using the same map directive that pulls the neighborhoods geojson would take a bit more work.

flibbertigibbet: I expect they'll diverge more as we add features. Was thinking it might be worth revisiting to see about a refactor once we have more of idea of what else we'll be doing with these (interactivity and popup building, in particular.)

CloudNiner: 👍

src/angularjs/src/app/components/map/map.constants.js
flibbertigibbet: Sure thing.

PR #376, opened 2017-05-01
Feature/kak/places list compare#251

### LINKED ISSUE #251: Implement compare feature on neighborhood list page

flibbertigibbet:
## Overview

Implement comparison feature on places list page.

### Demo

image

### Notes

Dropdown list items are currently unstyled.

Also added a footer to the comparison page cards to link to the place detail, since that was in the wireframes.

Refactored some of the comparison place tracking to preserve the offsets of selections, so the cards don't shift around on going back and forth between the place list and comparison pages after removing a selection.

## Testing Instructions

 * Go to [places list](http://localhost:9301/#/places////)
 * Should be able to add and remove places for comparison
 * Shouldn't be able to add the same place twice
 * Should be able to get to comparison page with current selections using last option in compare drop-down
 * Should be able to move between compare and list page with selections preserved

Closes #251

### DIFF
--- a/src/angularjs/src/app/index.route.js
+++ b/src/angularjs/src/app/index.route.js
@@ -20,7 +20,7 @@
                 template: '<ui-view/>'
             })
             .state('places.list', {
-                url: '',
+                url: ':place1/:place2/:place3/',
                 controller: 'PlaceListController',
                 controllerAs: 'placeList',
                 templateUrl: 'app/places/list/place-list.html'
@@ -32,7 +32,7 @@
                 templateUrl: 'app/places/detail/places-detail.html'
             })
             .state('places.compare', {
-                url: ':compare/:place1/:place2/:place3/',
+                url: '/compare/:place1/:place2/:place3/',
                 controller: 'CompareController',
                 controllerAs: 'compare',
                 templateUrl: 'app/places/compare/compare.html'
--- a/src/angularjs/src/app/places/list/place-list.html
+++ b/src/angularjs/src/app/places/list/place-list.html
@@ -37,31 +37,26 @@
 
                 <!-- Compare dropdown -->
                 <div class="form-group">
-                    <label for="3">Compare</label>
-                    <div class="dropdown">
-                      <button class="btn btn-default dropdown-toggle btn-block" type="button" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
-                        3 Selected
-                        <span class="caret"></span>
-                      </button>
-
-                      <!-- We are using bootstraps dropdown js -->
-                      <ul class="dropdown-menu compare-dropdown" aria-labelledby="dropdownMenu1">
-
-                        <!-- Only up to 3 to compare. Duplicate li and children for each item -->
-                        <li>
-                            <div class="compare-title">Los Angeles, California</div>
-                            <div class="network-score small">90</div>
+                    <label for="compare-dropdown">Compare</label>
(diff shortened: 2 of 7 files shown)


src/angularjs/src/app/index.route.js
```diff
@@ -20,7 +20,7 @@
                 template: '<ui-view/>'
             })
             .state('places.list', {
-                url: '',
+                url: ':place1/:place2/:place3/',
```
KlaasH: Would this be necessary if we could add compare cities from the compare page?  I see how it's doing something useful, but it seems awkward to me.  I.e. it's user-visible and I think to a user it would not be clear why the places list should have such a long, complicated URL.

flibbertigibbet: Not sure what you mean by adding from the compare page; like another search drop-down? From the prototype, it seemed to me the intent was to go back and forth between the list and compare pages with some set of selections.

Another option would be to handle the selections internally with a service, but then the selections would be lost on page refresh and direct linking wouldn't work (not sure if we care about either of those cases). If they haven't selected anything yet, the URL just has a couple of extra slashes on the end.

src/angularjs/src/app/places/list/place-list.html
```diff
@@ -37,31 +37,26 @@ <h2 class="sidebar-title">City/Town Search</h2>
...
-                        </li>
-                        <li>
-                            <div class="compare-title">Los Angeles, California</div>
-                            <div class="network-score small">90</div>
+                    <label for="compare-dropdown">Compare</label>
+                    <select ng-model="placeList.placeToRemoveFromComparison" name="compare-dropdown"
+                        ng-disabled="!placeList.comparePlacesCount()" class="form-control"
+                        id="role" type="text">
+                    <option selected disabled value="">{{placeList.comparePlacesCount()}} Selected</option>
+                    <option ng-repeat="neighborhood in placeList.comparePlaces | filter:{uuid: '!!'}"
```
flibbertigibbet: Right, it gets lost; that's why I was going to make a separate issue for the styling. Prototype used plain Bootstrap, but we have Angular components here that apparently have their own ideas about styling. Good find on the `compare-dropdown` class.

flibbertigibbet:
@KlaasH you've reported a number of issues here that aren't in the behavior I'm seeing. Perhaps you need to run `./scripts/update`. I've rebased against develop, if that helps. After rebuild myself, all seems well.

KlaasH:
Did a full `scripts/setup` and I'm still seeing the issues above.  The error on trying to "remove" might be more specific than what I described, but I definitely still get it with "add, add, remove first, remove second".

flibbertigibbet:
This should be ready for another look now.

* Fixed issue with removing for comparison via drop-down (order-specific to reproduce)
* Updated link back to place list from top of compare card (match behavior of button at page top)
* Add ternary state for compare buttons in list cards; toggles to be a remove button after selection
* Disable link to compare page until there are 2+ selections. Dropdown itself still disabled if no selections.
* Remove place from compare page without reloading other cards' content
* Bonus: swapped out `ng-options` with Angular Bootstrap dropdown control, which fixes styling issues (supports templating)

Punting on the idea of a separate dropdown/search interface in the compare page, as there's nothing for that in the prototype. Not clear what the compare place card 'search' buttons are supposed to do, and the link back serves the purpose, though it seems to imply user will be selecting for that specific column.

flibbertigibbet:
Needed to call `invalidateSize` on the map for the bounds to fit properly. Curiously, a timeout wasn't enough for the map to figure out for itself its new size.

PR #391, opened 2017-05-03
Add login and logout links to public site navbar

### LINKED ISSUE #312: Add login/logout links to navbar

flibbertigibbet:
## Overview

Change logo linking to homepage on the left of the navbar to a dropdown menu, in order to add a link to log in. Moved link to home under dropdown.

### Notes

@designmatty  styling of dropdown text appears to be broken by the brand class styles. Commenting out the `brand` class styling in `_navbar.scss` makes the drop-down text visible again. Text is visible on admin portion of site, but not on the public site.

## Testing Instructions

 * Logo on left of navbar should now be a dropdown menu
 * Should be able to log out via bottom item in dropdown
 * Should be able to navigate to home page via top item in dropdown

Closes #312.

### DIFF
--- a/src/angularjs/src/app/components/navbar/navbar.html
+++ b/src/angularjs/src/app/components/navbar/navbar.html
@@ -2,10 +2,25 @@
   <div class="container">
     <!-- Logo area -->
     <div class="navbar-left">
-      <a ui-sref="home" class="brand">
-        <img ng-if="navbar.hideName" src="assets/images/placesforbikes-icon.png" alt="Places for bikes">
-        <img ng-if="!navbar.hideName" src="assets/images/placesforbikes-logo-white.png" alt="Places for bikes">
-      </a>
+      <span uib-dropdown class="dropdown"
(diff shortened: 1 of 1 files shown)


designmatty:
I would recommend against putting login and logout as a dropdown to the logo. The link would be much better suited in the navbar to the right.

flibbertigibbet:
Normally we do put login/out on the right, but we have a link out to the parent site there now. Another concern is that most users will have no access to the administrative portion of the site that's behind the login, so we'd like to bury the link a bit. Any suggestions on a good way to handle these things @designmatty?

designmatty:
If the goal is to bury the link a bit, I would say put it in the footer. The footer is at the bottom and less likely to be reached by every visitor. Its non-intrusive and doesn't interrupt the user intuition of clicking what the logo is meant to do.

Something like this would work:
screen shot 2017-05-03 at 3 35 10 pm

flibbertigibbet:
Okay, thanks. Will update PR to go with the footer link.

flibbertigibbet:
Superseded by #392

PR #398, opened 2017-05-05
Allow read-only access to neighborhoods and analysis jobs

### LINKED ISSUE #382: Allow unauthenticated requests to AnalysisJob/Neighborhood GET endpoints
So that the public site works without requiring a login

flibbertigibbet:
## Overview

Allow read-only access to neighborhoods and analysis jobs, to support use of the public site without login. 

## Testing Instructions

 * Log out of UI and API
 * Should be able to get results from http://localhost:9200/api/neighborhoods/
 * Should be able to get results from http://localhost:9200/api/analysis_jobs/
 * Should not be able to perform POST operations on those endpoints
 * Should not be able to access other, protected endpoints, i.e. http://localhost:9200/api/organizations/

Closes #382.

### DIFF
--- a/src/django/pfb_analysis/views.py
+++ b/src/django/pfb_analysis/views.py
@@ -8,6 +8,7 @@ from django.utils.text import slugify
 from rest_framework import status
 from rest_framework.decorators import detail_route
 from rest_framework.filters import DjangoFilterBackend, OrderingFilter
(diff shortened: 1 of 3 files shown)


CloudNiner:
Merging so included in the deployment

PR #401, opened 2017-05-08
Allow unauthenticated requests

### LINKED ISSUE #388: Ensure public site works for un-authenticated requests
Do not attempt to send auth if it does not exist.

flibbertigibbet:
## Overview

Allow unauthenticated requests to API endpoints used by public site.

### Notes

Fixes up changes in #398 that overlapped changes in #394. Makes read-only GeoJSON endpoints publicly accessible.

## Testing Instructions

 * Public UI should work when logged out

Closes #388

### DIFF
--- a/src/django/pfb_analysis/views.py
+++ b/src/django/pfb_analysis/views.py
@@ -11,7 +11,7 @@ from rest_framework import status
 from rest_framework.decorators import detail_route
 from rest_framework.exceptions import NotFound
 from rest_framework.filters import DjangoFilterBackend, OrderingFilter
-from rest_framework.permissions import DjangoModelPermissionsOrAnonReadOnly
+from rest_framework.permissions import (AllowAny, DjangoModelPermissionsOrAnonReadOnly)
 from rest_framework.views import APIView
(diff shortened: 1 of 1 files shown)


src/django/pfb_analysis/views.py
```diff
@@ -96,6 +96,7 @@ class NeighborhoodBoundsGeoJsonViewList(APIView):
 
     pagination_class = None
     filter_class = None
+    permission_classes = (AllowAny,)
```
CloudNiner: It's probably more correct to use existing `DjangoModelPermissionsOrAnonReadOnly` for these endpoints, so that if we did add write operations to these endpoints we wouldn't accidentally give the world access. But I don't see that happening, so if its not a direct drop and replacement I'm not so worried about it.

flibbertigibbet: `DjangoModelPermissionsOrAnonReadOnly` only works on views with a `queryset` defined. These are read-only endpoints just for rendering GeoJSON; the POST requests all go to the primary `Neighborhood` endpoint.

CloudNiner: 👍

Issue #402, opened 2017-05-09
Unable to create AnalysisJobs or Neighborhoods via Browseable API or Admin UI


CloudNiner:
Likely fallout from #398 or #401? The user making the requests is an Administrator.

Reproduce by:
- navigating to http://localhost:9200/api/analysis_jobs/ -- the "Create new AnalysisJob" form at the bottom of the page should be missing

or

- navigating to http://localhost:9301/#/admin/analysis-jobs/create/ and attempt to create a new job. It should fail with a red error message notification in the top right after clicking "Submit Analysis Job"

CloudNiner:
Closed by #403

PR #403, opened 2017-05-09
Fixup admin users unable to create AnalysisJob/Neighborhood

CloudNiner:
## Overview

Restores lost ability to create AnalysisJob/Neighborhood objects as an admin user. 

### Notes

Permissions are implemented via custom Permission classes,
and do not interface with the Django model permissions feature.

So IsAuthenticatedOrReadOnly is sufficient here.

## Testing Instructions

Ensure a public user can still access GET for analysisjob, neighborhood and the geojson endpoints.

Ensure a logged in viewer user can view the same GET endpoints, but no user/org info or write to analysisjob/neighborhood

Ensure an admin user can write to everything.

Connects #402

### DIFF
--- a/src/django/pfb_analysis/views.py
+++ b/src/django/pfb_analysis/views.py
@@ -11,7 +11,7 @@ from rest_framework import status
 from rest_framework.decorators import detail_route
 from rest_framework.exceptions import NotFound
 from rest_framework.filters import DjangoFilterBackend, OrderingFilter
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
By the way, #402 may need to be closed directly, or summary here changed from 'connects' to 'closes'.

CloudNiner:
True about RestrictedCreate, no finer-grained access control is required at this point aside from what's being implemented for OrgAdmin, so this should be ok for now.

PR #404, opened 2017-05-09
Neighborhood visibility flag

### LINKED ISSUE #381: Add public/private flag field to Neighborhood model
Any neighborhood with the "private" flag set will not be visible to unauthenticated users (to be implemented in a separate task)

Consider choices field over boolean to allow for additional options in future

CloudNiner:
## Overview

Adds visibility field to Neighborhood model, so neighborhoods can be filtered to only show certain neighborhoods to the public.

NOTE: Requires rebase against #403 to work properly, so left the WIP flag for now, as if you test this branch you won't actually be able to create neighborhoods via API or admin UI 

### Demo

screen shot 2017-05-09 at 11 37 34
screen shot 2017-05-09 at 11 37 26

## Testing Instructions

- Run migrations. Ensure all existing neighborhoods have default 'public' state. 
- Navigate to http://localhost:9301/#/admin/neighborhoods/create/ and create both a new public and private neighborhood
- Inspect neighborhoods at http://localhost:9200/api/neighborhoods/ to ensure visibility is correct

Closes #381


CloudNiner:
Visibility added to neighborhood list in Admin UI.

Issue #406, opened 2017-05-09
Clean up UI error handling messages


flibbertigibbet:
Angular app displays `toastr` messages sometimes on promise failure, but not always, and not always in a format consistent with the response. Clean up and add UI error handling.

PR #407, opened 2017-05-09
Feature/kak/org admin group#389

### LINKED ISSUE #389: Add middle OrgAdmin user level
Currently we support two user levels:
- Administrator: RW on all models
- Viewer: R on AnalysisJob/Neighborhood

We need to add a third user level, to provide to PFB so they can create their own users. This OrgAdmin level should have the following permissions:
- R on AnalysisJob/Neighborhood
- RW on Users in the OrgAdmin's organization, but can only create users with an equal or lower permission level (so only create OrgAdmins or Viewers)

Ensure that the existing Administrator/Viewer roles work as expected as part of this task.

### LINKED ISSUE #405: Remove uploader and editor roles
Will not be implemented. Only user roles are organization admin, (full) admin, and viewer. Remove references from DRF app and UI.

Users with uploader or editor roles should be reassigned to viewer role (effective status now).

flibbertigibbet:
## Overview

Add user role for organization administrators. Can create users within their org and edit non-admin users in org; otherwise have viewer permissions.

### Demo

In UI, org admin has access to manage users but not organizations. 'Edit' button hidden by full admins from org admins, as they don't have access to modify full admins.

image

### Notes

Also removes editor and uploader roles (closes #405.) Fixes up error message display in the UI for user management, but it looks like UI error handling generally could use some cleanup and additions; analysis job creation, for example, does not display correctly. Created #406 for additional UI error handling cleanup.

## Testing Instructions

 * Run django migrations
 * Any existing uploaders or editors should now have viewer role
 * Check browsing admin angular app UI as org admin user
 * Org admin user should be able to create and manage non-admin users in their org
 * Org admin user should not be able to create a full admin or make existing user a full admin
 * Org admin users should otherwise have viewer access
 * Uploader and editor roles should not longer be displayed in drop-down options

Closes #389. Closes #405.


src/angularjs/src/app/users/detail/users-detail.controller.js
```diff
@@ -100,9 +99,18 @@
...
+                    displayError(error);
                 });
             }
         }
+
+        // Helper to display error popup on DRF request failure
+        function displayError(error) {
+            if (error.data && error.data.detail) {
+                toastr.error(error.data.detail, {timeOut: 5000});
+            }
```
CloudNiner: For better error visibility, an else block that calls `$log.error(error)` might be useful

src/angularjs/src/app/users/detail/users-detail.html
```diff
@@ -45,7 +45,7 @@ <h1 ng-if="user.editing">Edit User</h1>
             <div class="form-group">
```
CloudNiner: Now that the logic here to disable this control is getting a bit complicated, would it be clearer to just `ng-if` this block into existence if `user.isAdminUser || user.isOrgAdminUser || user.loggedInEmail === user.user.email`?

CloudNiner:
Whoops. I spoke too soon. Was able to trigger a crash by attempting to create an Administrator user using an OrgAdmin user via the admin UI user create page:
```
django_1     |   File "/usr/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 236, in is_valid
django_1     |     self._validated_data = self.run_validation(self.initial_data)
django_1     |   File "/usr/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 427, in run_validation
django_1     |     value = self.to_internal_value(data)
django_1     |   File "/usr/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 459, in to_internal_value
django_1     |     validated_value = validate_method(validated_value)
django_1     |   File "/usr/src/users/serializers.py", line 77, in validate_role
django_1     |     if value != self.instance.role:
django_1     | AttributeError: 'NoneType' object has no attribute 'role'
```

CloudNiner:
Ah, also watch out for a merge migration in the rebase against develop, another migration was added in #404

flibbertigibbet:
Migration here is on a separate app (`users`), so shouldn't present a conflict. Going to rebase before merge, in any case.

src/angularjs/src/app/users/detail/users-detail.html
flibbertigibbet: We do want to show it, disabled, so the role is visible but not editable, so probably shouldn't hide it with an `ng-if`. Considering it's necessary to be an admin or org admin to get to the user management page at all, I'll just go ahead and remove the first two tests.

PR #410, opened 2017-05-10
Add score labels/descriptions to public UI

### LINKED ISSUE #378: Add score labels/descriptions to UI neighborhood detail view
Score row currently only displays the score name, instead it should display the score label and the score description in a tooltip to the right.

Depends on #325

CloudNiner:
## Overview

Adds score labels and descriptions to the places detail view

### Demo

screen shot 2017-05-10 at 10 04 14
screen shot 2017-05-10 at 10 13 08

### Notes

The description tooltips won't currently display because the fontello.css file isn't being built into the application bundle. I spent 15 mins attempting to fix it but no luck, needs a bit more digging to get properly bundled. The description text is added to the DOM, so if you manually give the tooltip div a width via dev tools you can hover the blank div and view the tooltips.

## Testing Instructions

- `./scripts/django-manage loaddata analysis-score-metadata`
- Navigate to places detail page
- Labels should replace the score names, and descriptions should exist in the DOM 
- http://localhost:9200/api/score_metadata/ should exist and return data without logging in

Closes #378

### DIFF
--- a/src/angularjs/src/app/components/places.service.js
+++ b/src/angularjs/src/app/components/places.service.js
@@ -51,8 +51,9 @@
                     // sort alphabetically by metric name so they will line up by row
                     place.scores = _(results.overall_scores).map(function(obj, key) {
                         return {
-                            metric: key.replace(/_/g, ' '),
-                            score: obj.score_normalized
+                            metric: key,
+                            score: obj.score,
+                            score_normalized: obj.score_normalized
                         };
                     }).sortBy(function(result) { return result.metric; }).value();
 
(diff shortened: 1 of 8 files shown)


CloudNiner:
If I'm looking at the right thing, it seems like those popups only display high/low stress connectivity which aren't in the list of labels/descriptions we can return. That can be a separate issue, I'm not sure what that popup is going to look like in its final state quite yet.

flibbertigibbet:
Yes, those. They simply display whatever comes back in their properties, underscores replaced in the label. Okay, those don't map to metrics.

PR #414, opened 2017-05-10
Feature/kak/job status filter speed#400

### LINKED ISSUE #373: API endpoints very slow on production
With a large number (211) of neighborhoods:
screen shot 2017-05-01 at 08 10 27

### LINKED ISSUE #400: Filtering and ordering of analysis jobs slow in API
On production, querying `/api/analysis_jobs/?format=json&latest=True&limit=8&ordering=-overall_score&status=COMPLETE` takes over 9 seconds, but querying `/api/analysis_jobs/?format=json&limit=8` takes only ~350ms.

Fixing the slow filtering and ordering of analysis jobs should address #373.

flibbertigibbet:
## Overview

Reduce the number and complexity of queries and operations performed for commonly used filter to get last completed or last run analysis job for a neighborhood. Closes #400, which should also close #373.

### Notes

Denormalizes job status, start time (from when import began), and final run time. Reduced queries for latest analysis jobs from 25 to four, two of which are for neighborhoods. Neighborhood queries could possibly be also reduced to one, but added `select_related` to filter query had no effect, likely due to the related queryset defined for the custom `PrimaryKeyReferenceRelatedField` on the serializer.

## Testing Instructions

To enable database query debugging, add the following clause under 'loggers' in the settings:
```python
'django.db.backends': {
    'handlers': ['console'],
    'level': 'DEBUG',
},
```

 * Run migrations
 * Should have added fields set after migrations (i.e. `last_job` set on Neighborhood objects)
 * Query http://localhost:9200/api/analysis_jobs/?format=json&latest=True&limit=8&ordering=-overall_score&status=COMPLETE
 * Query results should output same results as previously

Connects #373 Closes #400


flibbertigibbet:
Pushed fix for migration.

flibbertigibbet:
On further examination, it looks like there are separate `Neighborhood` queries being performed for each job returned, so a serializer refactor may also be necessary.

CloudNiner:
Ok. So wait to review until you've had a chance to dig into the Neighborhood queries?

flibbertigibbet:
Fix forthcoming.

flibbertigibbet:
Okay, have pushed a modification to the serializer that allows for pre-fetched neighborhoods to be used. There is now no separate query for neighborhoods, but are cached when the jobs are retrieved.

For some reason querying in the DRF pretty view (without the `format=json` query parameter) results in an extra organization queries being performed repeatedly that don't seem to be addressed with an extra pre-fetch. However, this doesn't happen when the JSON format is explicitly requested, so probably okay to leave as-is.

PR #424, opened 2017-05-11
Feature/awf/django filter neighborhood vis#387

### LINKED ISSUE #387: AnalysisJob/Neighborhood GET endpoints should only return results for PUBLIC neighborhoods
Depends on #381, #382

CloudNiner:
## Overview

Updates the AnalysisJob and Neighborhood endpoints to only return objects with visibility=PUBLIC when a request is made with no auth credentials.

### Demo

Authenticated (5 results):
screen shot 2017-05-11 at 09 09 00

No authentication (4 results):
screen shot 2017-05-11 at 09 09 12

Same behavior for the AnalysisJobs endpoint

### Notes

Also fixes a crash that would occur when trying to create a Neighborhood via the admin UI, caused by `ValidationError: last_job cannot be blank`, which it can be because its a nullable field and is read-only.

## Testing Instructions

 * Run migrations
 * Create a private Neighborhood via the admin UI
 * Create a new job or two for the new private Neighborhood via the admin UI (not necessary to run them)
 * Log in/log out of the API while viewing:
  - http://localhost:9200/api/neighborhoods/
  - http://localhost:9200/api/analysis_jobs/
* The job counts should change when you log in/out and private jobs should not be visible while logged out

Closes #387


src/django/pfb_analysis/serializers.py
```diff
@@ -61,7 +61,7 @@ class Meta:
         exclude = ('created_at', 'modified_at', 'created_by', 'modified_by', 'geom', 'geom_simple',
                    'geom_pt',)
         read_only_fields = ('uuid', 'createdAt', 'modifiedAt', 'createdBy', 'modifiedBy',
-                            'organization', 'name',)
+                            'organization', 'last_job', 'name',)
```
flibbertigibbet: Good catch.

PR #433, opened 2017-05-11
Display scores using metadata on compare page

### LINKED ISSUE #428: Add metric labels/descriptions to compare detail page
Follow up from #378

flibbertigibbet:
## Overview

Add score labels from new metadata endpoint to comparison page metrics.

### Demo

image

## Testing Instructions

 * Check compare page
 * Metrics should have pretty labels and still line up across columns

Closes #428.

### DIFF
--- a/src/angularjs/src/app/places/compare/compare.controller.js
+++ b/src/angularjs/src/app/places/compare/compare.controller.js
@@ -10,13 +10,15 @@
     'use strict';
 
     /** @ngInject */
-    function CompareController($stateParams, Neighborhood, AnalysisJob, Places, $log, $q, $state) {
+    function CompareController($stateParams, AnalysisJob, Neighborhood, Places, ScoreMetadata,
+                               $log, $q, $state) {
         var ctl = this;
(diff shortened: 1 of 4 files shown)


src/angularjs/src/app/places/compare/compare.controller.js
```diff
@@ -28,9 +30,23 @@
                 return Places.getPlace(uuid);
             });
 
+            // first request score metadata
+            promises.unshift(ScoreMetadata.query().$promise);
```
CloudNiner: Now that we're using this in multiple places, can you add the `cache: true` parameter for the `ScoreMetadata.query` declaration on the `$resource`?

See: https://docs.angularjs.org/api/ngResource/service/$resource

Search for cache on the page at that link, its one of the options to the `actions` argument.

flibbertigibbet: We should probably be caching most resources by default, and only override for requests expecting to pick up a change.

PR #434, opened 2017-05-11
Search/filtering revamp

### LINKED ISSUE #419: Places list page makes a request which pulls all neighborhoods
https://bna.peopleforbikes.org/#/places//// requests:

https://bna.peopleforbikes.org/api/neighborhoods/?limit=all

Which takes 4+seconds on production. We can likely refactor this to only pull the neighborhoods we need, as this list view should be paginated.

### LINKED ISSUE #420: User interaction during place search is slow and janky on places list view
See screencapture:

pfb_place_search

### LINKED ISSUE #422: Status search on admin AnalysisJob view reloads full list automatically after filtering
pfb-jobs-status-search

CloudNiner:
## Overview

This PR completely refactors/revamps the AnalysisJob search feature for the admin and public sites.

Removes the typeahead and allows generic text filtering on the AnalysisJob results, sourced from the `Neighborhood.name` and `Neighborhood.state_abbrev` fields.

### Demo

Not really demoable via screenshot.

### Notes

I briefly toyed with using postgres vector search or trigrams, but figured it wasn't worth the extra complexity required to prepare the database. The list of neighborhoods isn't expected to grow above a few hundred rows, and the generic `search` param would allow for the backend implementation to be swapped later if necessary.

## Testing Instructions

Do some text searches on your completed jobs in the public places list view OR, if you only have a single germantown neighborhood there like I do, make some manual requests to http://localhost:9200/api/analysis_jobs/?search=value (so that the latest=True&status=complete filtering is not applied)

Do some filtering on the admin UI view for AnalysisJobs.

Both should be faster and feel much more natural.

I'll also do some testing of this once pushed to production to make sure it still works when the data source is larger.

Closes #419, Closes #420, Closes #422

### DIFF
--- a/src/angularjs/src/app/analysis-jobs/list/analysis-jobs-list.controller.js
+++ b/src/angularjs/src/app/analysis-jobs/list/analysis-jobs-list.controller.js
@@ -41,10 +41,10 @@
         }
 
         function filterJobs(filters) {
-            var params = _.merge({}, defaultParams, filters);
+            var params = _.merge({}, defaultParams);
 
             if (filters.neighborhood) {
-                params.neighborhood = filters.neighborhood;
+                params.search = filters.neighborhood;
             }
             if (filters.status) {
                 params.status = filters.status;
--- a/src/angularjs/src/app/components/filters/analysis-job-filter.html
+++ b/src/angularjs/src/app/components/filters/analysis-job-filter.html
@@ -2,23 +2,29 @@
     <th>ID</th>
     <th>
         <div class="dropdown">
-        <input name="neighborhood-filter" ng-model="ctl.statusFilter" title="Filter by Status"
-            placeholder="Status" type="text" class="form-control"
-            uib-typeahead="status for status in ctl.statuses |
-                           filter: $viewValue |
-                           limitTo: 8">
+        <input name="neighborhood-filter"
+               title="Filter by Status"
+               placeholder="Status"
+               ng-model="ctl.statusFilter"
+               ng-model-options="{ debounce: 400 }"
+               ng-change="ctl.onStatusFilterChanged()"
+               type="text"
+               uib-typeahead="status for status in ctl.statuses | filter: $viewValue | limitTo: 8"
+               typeahead-wait-ms="400"
+               typeahead-on-select="ctl.onStatusFilterChanged($item)"
+               class="form-control">
         </div>
     </th>
     <th>
         <div class="dropdown neighborhood">
-            <input title="Filter by Neighborhood" placeholder="Neighborhood"
-                ng-model="ctl.neighborhoodFilter"
-                type="text" class="form-control"
-                uib-typeahead="neighborhood as (neighborhood.label + ', ' +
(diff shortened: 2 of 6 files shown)


src/angularjs/src/app/analysis-jobs/list/analysis-jobs-list.controller.js
```diff
@@ -41,10 +41,10 @@
         }
 
         function filterJobs(filters) {
-            var params = _.merge({}, defaultParams, filters);
```
CloudNiner: Filters was only ever set via the watcher, it was removed since we manually set both properties we get from the filters object below

src/angularjs/src/app/components/filters/analysis-job-filter.html
```diff
@@ -2,23 +2,29 @@
...
-            placeholder="Status" type="text" class="form-control"
-            uib-typeahead="status for status in ctl.statuses |
-                           filter: $viewValue |
-                           limitTo: 8">
+        <input name="neighborhood-filter"
+               title="Filter by Status"
+               placeholder="Status"
+               ng-model="ctl.statusFilter"
+               ng-model-options="{ debounce: 400 }"
+               ng-change="ctl.onStatusFilterChanged()"
```
CloudNiner: Switching to `ng-change` and `typeahead-on-select` allows us to explicitly control when we set the `filters` property in the controller. This fixes a nasty side effect where the old implementation would trigger a call to `/api/analysis_jobs` in the parent controller every time a character was typed in either filter field. Now the call only explicitly happens when a meaningful change occurs.

src/angularjs/src/app/components/filters/analysis-job-filter.html
```diff
@@ -2,23 +2,29 @@
...
-        <input name="neighborhood-filter" ng-model="ctl.statusFilter" title="Filter by Status"
-            placeholder="Status" type="text" class="form-control"
-            uib-typeahead="status for status in ctl.statuses |
-                           filter: $viewValue |
-                           limitTo: 8">
+        <input name="neighborhood-filter"
+               title="Filter by Status"
+               placeholder="Status"
+               ng-model="ctl.statusFilter"
+               ng-model-options="{ debounce: 400 }"
```
CloudNiner: Apparently we need to debounce both via `ng-model-options` and `typeahead-wait-ms` for the debounce to work as expected. Only setting one or the other leads to the debounce not taking effect.

src/angularjs/src/app/places/list/place-list.html
```diff
@@ -75,7 +74,7 @@ <h2 class="sidebar-title">Place Search</h2>
 
         <section>
             <!-- Duplicate .result div for each result. -->
-            <div class="result" ng-repeat="neighborhood in placeList.places">
+            <div class="result" ng-repeat="neighborhood in placeList.places track by neighborhood.uuid">
```
CloudNiner: This fixes a bug where the list would re-render every item in the ng-repeat when the list of places was changed. Now it properly only re-renders the diff.

CloudNiner:
Ah true, let me see if I can fix that up quick.

PR #435, opened 2017-05-11
Add tiles and make layers exclusive on detail map

### LINKED ISSUE #426: Show generated tiles on detail map
See PR #354 for URL pattern.

### LINKED ISSUE #429: Destinations layers should be radio buttons
Only one can be toggled on at any given time

KlaasH:
## Overview

Makes two main changes to the place detail map:
- Adds the generated tile layers ("ways" and "census blocks")
- Makes the destination layers "pick one" (radio) rather than checkboxes

### Demo

image

### Notes

This is using the [leaflet-groupedlayercontrol](https://github.com/ismyrnow/leaflet-groupedlayercontrol) plugin.  That gets us the ability to make groups mutually exclusive.  The thing it doesn't have that the regular `L.Control.layers` does is sorting.  The regular picker supports two groups (base maps and overlays), so it's possible to get the separation we want by putting our overlays in as the base maps and our destinations in as overlays.  But we can't get the destinations to be radio buttons.  Probably that's more important than sorting, and we could possibly get them to sort by some other means if we need to.

Anyway, if you want to see the `L.Control.layers` version you can check out the second-to-last commit on this branch.  It looks like this:
image

Speaking of base maps, this moves the actual Positron base map out of the picker altogether, which makes sense since we only offer one.

## Testing Instructions

Rebuild your `angularjs` container to get the new bower package installed, then view a neighborhood that has a fully completed (including tiling) job.  You should still get the boundary initially but now be able to switch to the ways or census blocks tiles, and you should be able to pick only one destinations layer at a time.

Resolves #426, resolves #429.

### DIFF
--- a/src/angularjs/bower.json
+++ b/src/angularjs/bower.json
@@ -17,6 +17,7 @@
     "angular-ui-router": "0.4.2",
     "angular-toastr": "2.1.1",
     "leaflet": "1.0.3",
+    "leaflet-groupedlayercontrol": "0.6.0",
     "moment": "2.17.1",
     "animate.css": "3.5.2",
     "angular-bootstrap": "1.3.3",
--- a/src/angularjs/src/app/analysis-jobs/detail/analysis-jobs-detail.html
+++ b/src/angularjs/src/app/analysis-jobs/detail/analysis-jobs-detail.html
@@ -73,9 +73,9 @@
                   <td>Ways</td>
                   <td><a ng-href="{{ analysisJobDetail.results.ways_url }}">ESRI Shapefile</a></td>
               </tr>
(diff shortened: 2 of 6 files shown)


CloudNiner:
Hmmm, re only offering one basemap: @jcahail I think I remember something about offering a satellite basemap from the functional spec. Is that still a requirement? In which case the groupedlayercontrol is a superior choice since we could re-add positron+satellite

jcahail:
Correct- "option to view street map or satellite image map background" is listed as a key feature in the functional spec.

flibbertigibbet:
Satellite layer is a non-default option for the list view map.

KlaasH:
Added satellite basemap option:
image

KlaasH:
The random order was bothering me, so I did a sort.  Done fiddling now.

PR #469, opened 2017-05-17
Display bike scores in requested sort order

### LINKED ISSUE #458: Reorder scores on Compare page
[ ] Change “Overall score” to “BNA Score” and move to top of the list
[ ] Move "Total population" to second on the list
[ ] Move "People" to third on the list

CloudNiner:
## Overview

Displays analysis scores in requested sort order, according to #458. 

### Demo

### Notes

More annoying than expected. Ordering is a view layer concern here, so it didn't make sense to extend the API model with something like a priority field. Instead we declare the order in client code and always sort by the declared order before returning the array of AnalysisScoreMetadata objects.

Any metadata object returned that doesn't have a sort order defined in the UI code will sort after all of the defined ones.

## Testing Instructions

View compare and neighborhood detail views. Scores on page should reflect order described in #458. 

Closes #458


src/angularjs/src/app/places/compare/compare.controller.js
```diff
@@ -31,22 +31,23 @@
...
                 ctl.places = _.drop(results);
+                _.forEach(ctl.places, function (place) {
+                    place.scores = _.reduce(place.scores, function (result, score) {
+                        result[score.metric] = {
+                            score: score.score,
+                            score_normalized: score.score_normalized
+                        }
+                        return result;
+                    }, {});
+                })
```
KlaasH: Two questions on this and its companion below:
1. It's converting `place.scores` to an object keyed by metric rather than a list of objects, right?  If I'm reading that right, then
```
place.scores = _.keyBy(place.scores, 'metric');
```
would accomplish the same thing.  (The same except that it would also leave the `metric` key in the object, which is a bit redundant but harmless.)
2. If this is how we want the data, couldn't we just do it in [`places.service.js`](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/angularjs/src/app/components/places.service.js#L53)?  I believe these are the only two places where `Places.getPlace` is used.

CloudNiner: Whoops missed keyBy. I'll switch to that.

Yeah, I'll look at putting this change in the service.

CloudNiner:
After taking a break from this, putting a duplicate config in the JS was dumb. That might be justified for an app using an external API, but that's not the case here. It should be straightforward to move the ordering to the django model.

CloudNiner:
Updated and cleaned up. Probably worth another look.

PR #475, opened 2017-05-21
Draw intersection stress

spencerrecneps:
## Overview

This addresses one of the top priorities voiced by PFB in the mapping. Adds a table named neighborhood_ways_carto, which can be used for more fine-grained cartographic representation of the stress. Basically, it splits each line segment into three parts: one representing the way, and one on either end representing the intersection. This new table would need to be referenced in place of neighborhood_ways in generating tiles for the traffic stress layer.

It appears the tiles are all generated from the shapefile outputs, so it looks like we would need to be exporting this new table to shapefile as well. The field names to use in the cartographic styles are slightly different. The fields would be FT_STRESS and TF_STRESS.

### DIFF
--- a/src/analysis/features/ways_carto.sql
+++ b/src/analysis/features/ways_carto.sql
@@ -1,0 +1,77 @@
+----------------------------------------
+-- Creates a table of segmented geometries
+-- based on neighborhood_ways. Used for
+-- showing segment and intersection stress
+-- in mapping.
+--
+-- Variables:
+--      :nb_output_srid -> SRID of the analysis
+--      :min_length -> Minimum length, below which lines are split based on percentages
+--      :int_length -> Length of line to reserve for intersection stress
+----------------------------------------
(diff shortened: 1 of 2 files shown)


CloudNiner:
@spencerrecneps we just saw this PR, and its been hanging out untouched for almost two months. So a few questions:
- Is this still relevant?
- Is the feature complete?

If yes to both, can you provide a bit more detail on the columns that should be visualized on the ways tile layer? We currently select `FT_SEG_STR` and `TF_SEG_STR`. Presumably we would now have to also visualize `FT_STRESS` and `TF_STRESS` with the same styles as the segments?

spencerrecneps:
At this late date I think we can shelve this for now. I've explained to PFB the difficulties in visualizing the intersection stress in the online version and showed them how this effect can be accomplished in a desktop GIS. That's sufficient for now.

CloudNiner:
Ok thanks for the update. I'm going to close this PR for now. It can be reopened and reviewed at a later date if necessary.

PR #486, opened 2017-05-23
styles the nationwide card on homepage

### LINKED ISSUE #451: Add styling to "303 Places Nationwide" box on homepage map
PFB thinks the white box blends into the map too much.

image

designmatty:
## Overview

Adds a bit more styling to the card over the nationwide map.

### Demo
screen shot 2017-05-23 at 11 42 06 am

Fixes #451

### DIFF
--- a/src/angularjs/src/styles/components/_card.scss
+++ b/src/angularjs/src/styles/components/_card.scss
@@ -19,6 +19,30 @@
   }
 }
 
+.pfb-stripe {
+  &:before,
+  &:after {
+    content: '';
+    height: 6px;
+    transform: skew(45deg);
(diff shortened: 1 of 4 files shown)


src/angularjs/src/styles/components/_card.scss
```diff
@@ -19,6 +19,30 @@
   }
 }
 
+.pfb-stripe {
```
CloudNiner: Can this class be re-used for the cards in the "how is this score calculated" section below the map?

designmatty: yeah it can be.

PR #487, opened 2017-05-23
Changes default lane width to 5

spencerrecneps:
## Overview

Changes default bike lane width to 5 next to parking, which Fixes a problem with calculating the stress on roads with bike lanes. Must be committed before re-running Boulder and Fort Collins.


spencerrecneps:
@CloudNiner you should be OK to rerun Boulder and Fort Collins after this change. Let me know if you have questions.

PR #494, opened 2017-05-31
Add detail compare button

### LINKED ISSUE #489: Add "Compare" button to Details page
Compare button should direct user to Compare page.

N.b., The compare button was in the original wireframes, but was replaced with the Download button in a marked up wireframe from the client. Client would like the compare button restored and placed next to the Download button.

image

flibbertigibbet:
## Overview

Add button to header of place detail page that links to compare page.
Current place gets put in first slot of comparison page.

### Demo

image

### Notes

@designmatty might want to look at this. Default button styling is grey text until hover, which looks somewhat low-contrast on the grey header bar.

Closes #489.

### DIFF
--- a/src/angularjs/src/app/places/detail/places-detail.html
+++ b/src/angularjs/src/app/places/detail/places-detail.html
@@ -32,12 +32,10 @@
                             <div class="tooltip" data-title="On average, this is how well the bike network gets people to the places they want to go."><i class="icon-info-circled"></i></div>
                         </label>
                     </div>
-                    <div class="column">
+                    <div class="column text-right">
                         <button class="btn btn-default" type="button"
                             ui-sref="places.compare({place1: placeDetail.place.uuid})">Compare
                         </button>
-                    </div>
-                    <div class="column text-right">
                         <div class="dropdown" uib-dropdown is-open="status.isopen"
                             ng-if="placeDetail.downloads">
                           <button id="single-button" type="button" class="btn btn-default btn-s" uib-dropdown-toggle ng-disabled="disabled" id="downloadButton">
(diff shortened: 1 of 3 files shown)


flibbertigibbet:
Okay, will hold off on merge.

src/angularjs/src/app/places/detail/places-detail.html
```diff
@@ -32,20 +32,25 @@ <h4 class="sidebar-title">{{placeDetail.place.state_abbrev}}</h4>
...
-                        Download <span class="caret"></span>
-                      </button>
-                      <ul class="dropdown-menu" uib-dropdown-menu role="menu"
-                        aria-labelledby="downloadButton">
-                        <li ng-repeat="item in placeDetail.downloads">
-                            <a role="menuitem" href="{{item.url}}"
-                                class="compare-remove">{{item.label}}</a>
-                        </li>
-                      </ul>
+                        <div class="dropdown" uib-dropdown is-open="status.isopen"
```
designmatty: The compare button should sit inside of the same `.column` where the download button is located. The middle `.column` should be deleted.

Within the scss file `_location.scss` starting on line 2 add the styling:
```
.dropdown {
  display: inline;
}
```

As a minor styling detail I would recommend the following change in the scss file `_button.scss`, line 74: 
`color: $shade-dark;` to `color: #000;`

After these changes everything should be good to go.

flibbertigibbet:
Made the style changes above. Without the added column, though, the button stacks over the drop-down. Is that what we want @designmatty ?

image

designmatty:
It looks correct for me on osx chrome.

screen shot 2017-06-05 at 9 26 55 am

designmatty:
I would say go ahead and merge it in and I can fix that issue this sprint.

PR #495, opened 2017-05-31
Feature/kak/indent subscores#457

### LINKED ISSUE #457: Indent subscore categories on Compare page
Indent the subscores for the network categories (e.g., “Dentists” through “social services” are part of the Core Services score--indent those)

flibbertigibbet:
## Overview

Modify display of scores lists on comparison page and places detail page:
 - sort metrics into category
 - alphabetize by category
 - list 'total' score for category, if it exists, first
 - if 'total' exists, indent other metrics (sub-scores) listed below it

### Demo

Top of places detail:
image

Bottom of comparison page list:

image

### Notes

One category is an empty string, and so sorts first (population and the 'people' score.)

Identifying the category total metric relies on finding 'Total', preceded by a space, in the metric label; the space is required because one of the metrics is called 'Total Population', which does not represent a summary metric for its category.

Closes #457

### DIFF
--- a/src/angularjs/src/app/components/scoremetadata.service.js
+++ b/src/angularjs/src/app/components/scoremetadata.service.js
@@ -8,42 +8,27 @@
 (function() {
     'use strict';
 
-    /**
-     Sort metrics in metadata by category, with total first in the category.
-     Add subscoreClass property to metrics that should be indented under the total.
-     */
-    function orderCategories(response) {
-        var metadata = [];
+    // Add subscoreClass property to metrics that should be indented under the total.
+    function labelSubcategories(response) {
 
         // remove overall score and group into categories
-        var groupedMetadata = _.chain(response).reject(function(m) {
+        var metadata = _.reject(response, function(m) {
             return m.name === 'overall_score';
-        }).groupBy('category').value();
+        });
 
-        // sort by category name (note population category is blank string and sorts to top)
-        var categories = _.keys(groupedMetadata).sort();
-
-        // Build metadata array, sorted by category.
-        // Category total metric comes first within a category, if it exists.
-        // If it does exist, the other metrics within its category have a special property added,
-        // 'subscoreClass', which can be used to format sub-scores under the total.
-        _.each(categories, function(category) {
-            var metrics = groupedMetadata[category];
-            var totalMetric = _.remove(metrics, function(metric) {
-                return metric.label && metric.label.indexOf(' Total') > -1;
-            });
-
-            if (totalMetric && totalMetric.length) {
-                totalMetric = totalMetric[0]; // remove returns an array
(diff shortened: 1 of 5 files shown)


src/angularjs/src/app/components/scoremetadata.service.js
```diff
@@ -8,6 +8,47 @@
 (function() {
     'use strict';
 
+    /**
+     Sort metrics in metadata by category, with total first in the category.
+     Add subscoreClass property to metrics that should be indented under the total.
+     */
+    function orderCategories(response) {
```
CloudNiner: This is ok, and results in the same sort order as explicitly requested by PFB, but it is also brittle in that if we have to add an additional score down the road, or if PFB requests a different ordering, we're no longer respecting the explicit sort order provided by the API that's encapsulated by [AnalysisScoreMetadata.priority](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/django/pfb_analysis/models.py#L631).

A potential alternative to the logic in this method would be to update the AnalysisScoreMetadata.category values in [the fixture](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/django/pfb_analysis/fixtures/analysis-score-metadata.json) to be null for any top level category, and then set the subscore CSS class for any metric that has a non-null category.

flibbertigibbet: Ah, I didn't know there was explicit ordering applied to the metadata response from the API.

Relying on the ordering of the response to be stable will also mean assuming that the ordering matches the structure we need (grouped by category, with summary metric first). This task has those implicit ordering assumptions.

Dropping the category for summary scores would potentially break any other use of categories.

If we can rely on a stable sort order from the API that matches those assumptions, then this client-side sub-category identification logic can be simplified somewhat to just iterate through once.

CloudNiner: Yes, the two goals are definitely in contention. You cannot have arbitrary ordering at the same time as indentation for subcategories, as that implies at least some constraints on order. I can't find the comment, but I did mention this before. I think for now we can do as you suggest and "rely on a stable sort order from the API that matches [the assumption that the order is constrained by the subcategories all appearing directly after their associated top-level category]". 

In any case, we'd need to revisit this with any significant revision to the sort order, and I'm not worried about that happening ATM. 

I don't believe there are any other uses of category in the public site.

flibbertigibbet:
Updated to rely on API response order. Had to delete local metadata before re-running `update` script in order to pick up the changes in d8cfc9ca3e1dd75293e2fe792cce48c7233da585.

CloudNiner:
Merging so this gets to staging before RRP.

CloudNiner:
Oh wait, nvmnd can't because its on your fork.

PR #496, opened 2017-05-31
Add map attribution link to OSM

### LINKED ISSUE #485: Add link to improve OSM in map
We would like to have a link on the web map for people to click if they want to make improvements to OSM. The use case is that someone sees that a bike lane is missing and wants to add it, so they can click the link and learn how to make OSM edits.

I've seen something like this on Mapbox maps:
image

If this can easily be added to the attribution it would be great to have. If adding to the attribution is more than a minimal effort, another suitable alternative would be to add a link and text that says "See something missing from your bike network? Learn how to make updates _here_." The link should go to http://learnosm.org/en/
Ideally, this would be added on the neighborhood detail page above or underneath the map. Alternatively, this could be added at the bottom of the About page under the "Our process is described in more detail, including a full technical summary, here."

flibbertigibbet:
## Overview

Add a link to OSM informational page from map attribution for the OSM-based base layer options (but not the satellite map.) Displays on all non-thumbnail maps, as the thumbnail maps do not display the attribution.

### Demo

image

Closes #485.

### DIFF
--- a/src/angularjs/src/app/components/map/map.constants.js
+++ b/src/angularjs/src/app/components/map/map.constants.js
@@ -7,14 +7,16 @@
             url: 'https://cartodb-basemaps-{s}.global.ssl.fastly.net/light_all/{z}/{x}/{y}.png',
             attribution: [
                 '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>, ',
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
:+1:

PR #500, opened 2017-06-07
Update header text on home page.

spencerrecneps:
## Overview

Changes header text on home page based on request from PFB.


CloudNiner:
Presumably this PR is also still relevant correct?

CloudNiner:
@spencerrecneps bump. See above ^^

spencerrecneps:
Looks to me like this text has already been changed on the main website, so I think this can be ignored. Wasn't this one of the "emergency" changes that you had to make shortly after publishing the website?

CloudNiner:
So it is. Closing, already applied in a97a1366542396a963c3f03ccb04ea2042037db5

PR #512, opened 2017-06-22
Add README section with notes for non-windows users

spencerrecneps:
## Overview

Adds a section to the README with notes for non-Windows users. I tried running on Ubuntu and noticed that an NFS daemon is required, which is not included by default.


CloudNiner:
Thanks!

PR #516, opened 2017-07-03
Add tile styling project files, and a README

### LINKED ISSUE #506: Add tile style projects to repo
I used Mapbox Studio Classic to compose and tweak the styles for the ways and census blocks layers.  It would be a pain to work backwards from the exported XML styles, so those project files should go into the repo.

KlaasH:
Adds the working files necessary to load and modify the tile styles with
Mapbox Studio Classic, and a README file to explain what's going on.

### Demo

image

### Notes

The process is a little squirrelly--starting with hand-editing the project files to correct hard-coded absolute file paths and ending with hand-copying the translated styles from the project XML to the style XML we use to do the actual tile generation.  There are probably better ways to go about this, but this does the job.

## Testing Instructions

Follow the README instructions to install Mapbox Studio Classic, modify the file paths, and open the project.  It should show up looking roughly like the screenshot.  Changing a style and saving should result in a diff in the files in `combined_styles.tm2`.

Resolves #506.


src/tilemaker/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+For ease of editing and previewing, the styles are defined and maintained in a
+Mapbox Studio Classic project from which the XML style definitions are extracted by hand.
+
+Changes to the working files should be committed so that they're kept in sync with the actual
+tile styles being applied.
+
+### To open and edit the styles
+
+- Download the latest release of Mapbox Studio Classic from https://mapbox.s3.amazonaws.com/mapbox-studio/index.html (the [Mapbox download page](https://www.mapbox.com/mapbox-studio-classic/) links to an older, seemingly-broken version) and install it (for Linux,
```
CloudNiner: Just to note, I didn't have any issues using the latest release for MacOS on the Mapbox download page, the About dialog indicates it is `v0.3.8`.

src/tilemaker/README.md
```diff
@@ -0,0 +1,49 @@
...
+that just means unzipping it and running the `atom` command from the base directory).
+- Modify the hard-coded absolute paths in the following files to point to the actual locations
```
CloudNiner: I didn't have any issues doing this because I knew what to look for, but it may be helpful to include a sample file path that the user would be expected to change so they know what to look for. Then again maybe not, since the path currently there (`/home/local/...`) will probably accidentally change over time because the file is checked in. I think we can expect any developer looking at this to be able to figure out what to change based on the description provided. Maybe just change the wording to explicitly state "all", e.g. "Modify all hard-coded absolute paths in the following files..."?

src/tilemaker/README.md
CloudNiner: The latest version of Mapbox Studio Classic required that I have a mapbox login to use it, which we don't ever state in the readme. Might be worth mentioning as a prerequisite somewhere.

src/tilemaker/README.md
```diff
@@ -0,0 +1,49 @@
...
+on your machine.
+  ```
+  working_files/combined_data.tm2source/data.yml
+  working_files/combined_data.tm2source/data.xml
+  working_files/combined_styles.tm2/project.yml
+  working_files/combined_styles.tm2/project.xml
+  ```
+- In Mapbox Studio Classic, click "Styles & Sources" > "Browse" and select the
```
CloudNiner: I was presented with this splash page on initial login to mapbox studio:

screen shot 2017-07-05 at 09 26 05

Where I initially had to select "Blank Source" in the right-most column before getting to the editor view that I _think_ your directions are referencing:

screen shot 2017-07-05 at 09 27 41

The initial blue splash view cannot be returned to once you've started working with a project unless you logout and login to the application again. I think an additional screenshot or two showing these pages would be helpful here to orient the README user.

src/tilemaker/README.md
KlaasH: Ah, ok.  Will adjust to make this more specific.  v0.3.8 tries to start but immediately crashes on my machine.

KlaasH: Good point.  It remembers you, so I only saw that the first time and forgot about it by the time I was writing down the steps.

PR #520, opened 2017-07-10
Add legend for Bike Infrastructure

### LINKED ISSUE #505: Add legend for bike infrastructure tile layer
Depends on #504 

Use whatever stylings were determined as part of that issue

CloudNiner:
## Overview

Adds legend for the new Bike Infrastructure layer

### Demo

screen shot 2017-07-12 at 12 46 04

### Notes

For now, this is a trivial change, but it doesn't implement the white dashed line for "Track" paths. If PFB wants that styling, we'll need to address that separately.

## Testing Instructions

- If you haven't already run a tile generation step to generate tiles for the bike infrastructure layer, do so.
- Toggle the bike infrastructure layer on via the neighborhood detail view

Closes #505, Connects #521

### DIFF
--- a/src/angularjs/src/app/components/map/map.constants.js
+++ b/src/angularjs/src/app/components/map/map.constants.js
@@ -39,7 +39,7 @@
         },
         bike_infrastructure: {
             position: 'bottomright',
-            colors: ['#00aeef', '#005e83', '#005e83', '#006b3e'],
+            colors: ['#8c54de', '#5072f5', '#44a3a6', '#15bf50'],
             labels: ['Lane', 'Buffered Lane', 'Track', 'Off-Street Paths'],
             title: 'Bike Infrastructure'
         },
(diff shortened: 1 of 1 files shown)


CloudNiner:
@designmatty if we have to add the white dashed line over the blue background of the "Track" item in the legend, how would you go about doing so? We currently generate these legends based on a configuration object, so any changes would require some modification of the underlying library, unless we can do it all within the CSS `background` property

designmatty:
We could use an svg/png as the background image using the `background` property. But I think we'd have to have a specific class or id to target that specific item.

Do we always have the same number of items in the legend? and are they always in the same order? If yes, and classes or id's aren't viable, then we could use css `nth-child` to target the specific item.

CloudNiner:
Great, I'd expect we can do the SVG/PNG strategy, as we just set the background property for each `<i>` element to whatever is in the config in the diff on this PR. I'll followup again when we hear exactly what PFB wants.

CloudNiner:
@designmatty we're safe. PFB only wants solid colors for the features on this layer, so the existing code works fine. Updated the branch with the final colors from #521

PR #535, opened 2017-07-17
Feature/awf/bike infra fix track

CloudNiner:
## Overview

Fixes up the drawing order of the bike infrastructure lines. There were rendering issues with the different layers, specifically that the "tracks" would render as a thick green line instead of a thick teal line. 

#### Before

screen shot 2017-07-17 at 09 37 50

#### After

The thick green line on the right side of the screenshot above becomes teal below:

screen shot 2017-07-17 at 09 37 17

Verification that the crosswalk filtering is correct (its difficult to tell on the FoCo shapefile):

screen shot 2017-07-17 at 09 36 26
to
screen shot 2017-07-17 at 09 36 14

### Notes

This is now super annoying, as the styles can be easily broken by shifting the order around of various definitions in the CartoCSS, but I'm not sure how to improve on this right now. All ears on ways to improve this.

I also updated the FoCo shapefile in the tilemaker 'working_files' directory, so that it has the `XWALK` column.

## Testing Instructions

 * Ensure styles look correct in Mapbox Studio
 * Run Germantown job to ensure that the XWALK filtering still works correctly.


src/tilemaker/working_files/bike_infrastructure.tm2/project.yml
```diff
@@ -9,9 +9,9 @@ bounds:
   - 180
   - 85.0511
 center: 
-  - -111.0677
-  - 40.5554
-  - 12
+  - -111.1784
+  - 40.0162
```
KlaasH: This is making the project open for me with everything outside the viewport to the top.

CloudNiner: Hrmm, ok. I'll load it up and click around until its fixed.

Issue #541, opened 2017-08-01
Remove MN from the analysis results


jcahail:
Remove the detail page (https://bna.peopleforbikes.org/#/places/d03ec36c-74f2-4627-b26c-cddeed5a7e50/) and remove from search page.

CloudNiner:
After deploying #544 to production, I manually ran:
```python
minneapolis = Neighborhood.objects.get(name__icontains="minneapolis")
minneapolis.visibility = Neighborhood.Visibility.HIDDEN
minneapolis.save()
```
in the production console to remove Minneapolis from the results:

![screen shot 2017-08-08 at 11 15 49](https://user-images.githubusercontent.com/1818302/29079391-152d4fba-7c2b-11e7-8496-ed81229b6506.png)

CloudNiner:
@jcahail @spencerrecneps the other MN neighborhoods are still visible on the site, it looks like Jennifer specifically only requested that Minneapolis be hidden. If we want to hide all the Minnesota neighborhoods we can do that too. Otherwise, this issue is closed.

CloudNiner:
This task is complete. Confirmed with Jennifer that they only want Minneapolis hidden from view.

PR #544, opened 2017-08-02
Add Neighborhood.Visibility.HIDDEN option

CloudNiner:
Additionally hide HIDDEN neighborhoods from geojson

## Overview

Adds an additional visibility option, that hides a neighborhood from all public views on the site.

This allows us to set the MN neighborhoods/jobs to HIDDEN on production after this PR is deployed with the final site changes, and make them reappear later if PFB desires for any reason.

### Demo

Not so easy to demo, see testing section below.

### Notes

_Copied from commit message:_

Ideally, we should default to only returning PUBLIC.
Then if the request is authenticated, also return PRIVATE
neighborhoods. This is more complicated, though, and I
think I've exhausted my "make an incremental improvement
by spending a bit more time" budget by adding the HIDDEN
visibility option rather than just deleting all MN
jobs from the production site to address #541.

The private neighborhoods appearing on the home page
"is your place on the map?" map was already an issue here,
so we're not making things worse.

If this is deemed unnecessary or there are complications from adding HIDDEN that arise during code review, we should be ok to simply close this PR and manually delete the MN jobs from production.

## Testing Instructions

- Open a django shell console
- Get a ref to a neighborhood that has a job with `status=COMPLETE`
- Modify the neighborhood's visibility to `PRIVATE`, then refresh the site.
  - If you're signed out, the neighborhood should no longer be visible, except on the "is your place on the map?" map.
  - If you're signed in, the neighborhood should remain visible
- Modify the neighborhood's visibility to `HIDDEN`, then refresh the site.
  - If you're signed out, the neighborhood should no longer be visible, except on the "is your place on the map?" map.
  - If you're signed in, the neighborhood should no longer be visible, except on the "is your place on the map?" map.

Connects #541


src/angularjs/src/app/home/neighborhood-map.directive.js
```diff
@@ -19,7 +19,12 @@
         ctl.onMapReady = function (map) {
             ctl.map = map;
             Neighborhood.geojson().$promise.then(function (data) {
-                ctl.count = data && data.features ? data.features.length : '0';
+                if (data && data.features) {
```
CloudNiner: This tweak ensures that the JS on the site doesn't throw an exception when an empty feature collection with `features=null` is returned by the neighborhoods_geojson endpoint.

CloudNiner:
Agree, I don't like it so much either, but it felt out of scope for what I was trying to accomplish here...there were a few more things it would have been nice to clean up.

There is not an issue for better visibility/status filtering, as noted in the original PR comments, neighborhoods with visibility=PRIVATE will also show on the public home page map. I will create a separate issue. Excluding HIDDEN is desirable in all places, and was at least a first step towards this goal. The actual task should probably explore a DRYer way to accomplish the geojson filtering.

KlaasH:
I think I misread the bit about what's currently happening with the home page map.  What you described is indeed what's happening.

Issue #555, opened 2018-02-16
Build fails with NoRegionError


CloudNiner:
The `update_job_definitions.py` file fails with the following exception:
```
Creating network "awsbatch_default" with the default driver
Traceback (most recent call last):
  File "update_job_definition.py", line 50, in <module>
    main()
  File "update_job_definition.py", line 38, in main
    client = boto3.client('batch')
  File "/usr/local/lib/python3.6/site-packages/boto3/__init__.py", line 83, in client
    return _get_default_session().client(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/usr/local/lib/python3.6/site-packages/botocore/session.py", line 836, in create_client
    client_config=config, api_version=api_version)
  File "/usr/local/lib/python3.6/site-packages/botocore/client.py", line 71, in create_client
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/usr/local/lib/python3.6/site-packages/botocore/client.py", line 281, in _get_client_args
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/usr/local/lib/python3.6/site-packages/botocore/args.py", line 45, in get_client_args
    endpoint_url, is_secure, scoped_config)
  File "/usr/local/lib/python3.6/site-packages/botocore/args.py", line 111, in compute_client_args
    service_name, region_name, endpoint_url, is_secure)
  File "/usr/local/lib/python3.6/site-packages/botocore/client.py", line 354, in resolve
    service_name, region_name)
  File "/usr/local/lib/python3.6/site-packages/botocore/regions.py", line 122, in construct_endpoint
    partition, service_name, region_name)
  File "/usr/local/lib/python3.6/site-packages/botocore/regions.py", line 135, in _endpoint_for_partition
    raise NoRegionError()
botocore.exceptions.NoRegionError: You must specify a region.
```

CloudNiner:
It's unclear how this was working before, since we haven't changed boto versions in the docker container that runs the `update_job_definitions.py` file. 

[...]

PR #557, opened 2018-02-27
Remove unused and incorrect AWS_REGION variable

CloudNiner:
## Overview

In all environments, we fully configure the AWS
credentials via profile settings. So we don't need to
make the region available separately as an ENV variable.

Note that this variable was also incorrect. The correct
name for the ENV variable to set the region is
AWS_DEFAULT_PROFILE. See:
http://boto3.readthedocs.io/en/latest/guide/configuration.html#environment-variables

Given the above, it seemed best to just remove this ENV var without replacing it with `AWS_DEFAULT_PROFILE`.

## Testing Instructions

Follow the instructions in [deployment/README](./deployment/README.md#terraform) to deploy your current commit to staging. If your `pfb` AWS profile is configured correctly, with the region set in `.aws/config` and your access + secret keys in `.aws/credentials` then the deployment should still succeed, since the param removed here is extraneous.

Connects #555

### DIFF
--- a/deployment/aws-batch/docker-compose.yml
+++ b/deployment/aws-batch/docker-compose.yml
@@ -6,7 +6,6 @@ services:
         context: .
         dockerfile: Dockerfile
       environment:
-        - AWS_REGION=us-east-1
         - AWS_PROFILE=pfb
       volumes:
         - $HOME/.aws:/root/.aws:ro


CloudNiner:
@tnation14 note that this is low priority given that the build failures are addressed by [this note](https://github.com/azavea/pfb-network-connectivity/issues/555#issuecomment-368924020) and develop is building again.

PR #558, opened 2018-03-19
Upgrade Django to 1.11

### LINKED ISSUE #553: Upgrade to Django 1.11

KlaasH:
## Overview

Bumps Django to the next minor version, 1.11, and updates Python dependencies.

### Demo

Looks the same.  Also, I destroyed my instance so I don't have many neighborhoods...
image

### Notes

Re Django:
- The Azavea Django base container comes in two flavors for 1.11, "-slim" and "-alpine". I picked "-slim" because it comes with `gcc` installed, which is necessary for installing fiona.
- As the comment indicates, the Django and psycopg2 versions in the requirements file are used by other containers that install Django to push status updates to the Django container via management commands. This updates those versions to keep them in sync with what's installed in the base Django container.

Re dependencies:
- For `django-amazon-ses`, moving to version 1.0.0 would require some changes, so I upgraded to the last version before that (0.3.2).
- For everything else in this commit, I checked changelogs and didn't see any concerns, so they're at the latest versions.
- Django Rest Framework required some small adjustment--some functionality that was previously part of DRF was moved into `django-filters`.

## Testing Instructions

- Run `scripts/update`.  It should fully rebuild the Django container, since the base image changed, and partially rebuild the `analysis` and `tilemaker` containers, since they pull in the requirements file.
- Everything should work as before.  This project doesn't have much in the way of automated testing.  I clicked around the user-facing and admin-facing sites and ran an analysis (by making the neighborhood and job in the admin-facing site then running the commands from the container log).

Resolves #553.

### DIFF
--- a/src/django/Dockerfile
+++ b/src/django/Dockerfile
@@ -1,7 +1,7 @@
 # Note: the Django and psycopg2 versions are repeated in requirements.txt for the benefit
 # of the analysis container. The base container and requirements file versions should be kept
 # in sync.
-FROM quay.io/azavea/django:1.10.3
+FROM quay.io/azavea/django:1.11-python2.7-slim
 
 MAINTAINER Azavea
 
--- a/src/django/pfb_analysis/filters.py
+++ b/src/django/pfb_analysis/filters.py
@@ -2,7 +2,7 @@ import logging
 
 from django.db.models import Q
 
-from rest_framework import filters
+from django_filters import rest_framework as filters
 import django_filters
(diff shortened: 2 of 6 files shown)


KlaasH:
I had a look at staging after this got pushed there (which required bumping the postgres version in the tfvars file to 9.6.6, since the RDS instance had been automatically upgraded by a few minor versions since the last staging deployment).  Everything looks to be working fine.

Issue #452, opened 2017-05-16
Austin issue


jcahail:
Austin shows up twice, with the same odd boundary.
![image](https://cloud.githubusercontent.com/assets/24727081/26110550/b02e781a-3a20-11e7-9916-566bcf2c6dfd.png)

Is this a map issue or a boundary shapefile issue?

CloudNiner:
Duplicate neighborhood fixed as part of #453 

As for the odd boundary, the answer is...its complicated.

The Austin, TX MultiPolygon contains two polygons, one of which is very small and disjoint:
![screen shot 2017-05-17 at 15 45 56](https://cloud.githubusercontent.com/assets/1818302/26172920/0b34bb78-3b18-11e7-8163-26502a6d79c5.png)

When calling simplify without preserving topology, in all cases we currently handle in our simplification method, the result is a simplification of the small polygon shown above, rather than the larger actual boundary. 

I manually ran the following on production to fix the badly simplified geometry:
```
n = Neighborhood.objects.get(name='austin-tx')
n.geom_simple =  n.geom.simplify(SIMPLIFICATION_TOLERANCE_MORE, preserve_topology=True)
n.save()
```
This resulted in a MultiPolygon with ~1100 points, down from ~9000 in the original geom, and is a manageable, but not ideal size for now.

I spoke with @KlaasH and we have a few concerns about making modifications to the existing `pfb_analysis.models.simplify_geom`, we should followup with @flibbertigibbet when she returns, we don't want to make major changes to this method because we could easily mess up more simplified geometries than we fix. We think that updating the method to use `preserve_topology` will always work, but we worry about making the simplified geometries generally larger than they need to be.

In any case, here's the updated simplified geom on the site:
![screen shot 2017-05-17 at 15 54 30](https://cloud.githubusercontent.com/assets/1818302/26173248/2ae323e6-3b19-11e7-82ab-94defbcdf86b.png)
![screen shot 2017-05-17 at 15 54 37](https://cloud.githubusercontent.com/assets/1818302/26173247/2adf09e6-3b19-11e7-9a41-3b5ef671e838.png)

KlaasH:
Ann Arbor has this problem as well:
![image](https://cloud.githubusercontent.com/assets/6598836/26261142/10fc1d62-3c9e-11e7-8917-b48084877aad.png)

Possibly we should run a query to find all geometries with fewer than some minimum number of points.

[...]

PR #562, opened 2018-04-10
Refactor simplify_geom method to preserve topology if simplification issues

### LINKED ISSUE #488: Improve geometry simplification in pfb_analysis.models.simplify_geom
In #452 we ran one-off scripts against the existing neighborhoods in production to clean up Neighborhood simple geometries broken by the existing simplify_geom algorithm. 

The algorithm in simplify_geom should be updated once we have insight into whether it makes sense to just always use preserve_topology with the lower simplification tolerance. Is it worth keeping the existing logic around? Does preserve_topology lead to geojson geometries that are too large to serve/display?

CloudNiner:
## Overview

_Hopefully_ improves the failure rate, and is specifically intended to avoid geometries exhibiting the issue noted in #452.

It's difficult to locally test all of the geometries that are on production to ensure there aren't any signficant regressions, happy to hear any ideas you might have for that.

### Demo

Here is the Ann Arbor, MI shapefile uploaded to my local box with the changes in the PR applied. The geojson was retrieved via:

http://localhost:9200/api/neighborhoods_bounds_geojson/b578299c-effb-4bce-8a0a-ed3f6d04dceb/?format=json

Which is the same endpoint that the client app uses to show boundaries on the map.

Putting the result into geojson.io gives us:

screen shot 2018-04-10 at 11 34 46 am

Where the simple geometry is ~500 points down from the original ~1800.

### Notes

I tested this on 5-6 different known boundaries that we've uploaded in the past. It appears robust at a glance to empty, valid and non-polygon geometries, but I'm sure there's some failure case we're not anticipating that it doesn't handle well. 

I added a few simple tests that broadly verify that simplification was successful for a boundary that used to work with the old simplify_geom method, and a boundary that didn't work but now does that exhibits the preserve_topology issue. Happy to expand there a bit if we think that's appropriate.

I'm unsure 

## Testing Instructions

- `./scripts/server`
- Via the admin interface at http://localhost:9301/#/admin/neighborhoods/ upload a variety of new neighborhoods via shapefile. Ensure they all upload and simplify correctly by verifying the json output of `http://localhost:9200/api/neighborhoods_bounds_geojson/:neighborhoodId/?format=json` in a viewer such as geojson.io. Good choices that exhibit the issue in #488 are:
  - https://s3.amazonaws.com/production-pfb-storage-us-east-1/neighborhood_boundaries/ann-arbor-mi/ann-arbor-mi.zip
  - https://s3.amazonaws.com/production-pfb-storage-us-east-1/neighborhood_boundaries/austin-tx/austin-tx.zip
  - https://s3.amazonaws.com/production-pfb-storage-us-east-1/neighborhood_boundaries/denver/denver.zip
- The new tests should also pass.

Closes #488


src/django/pfb_analysis/models.py
```diff
@@ -47,24 +48,45 @@ def get_neighborhood_file_upload_path(instance, filename):
...
+        if is_simple_polygon_valid(simple, geom):
+            logger.debug('Neighborhood.simplify_geom used ' +
+                         'geom.simplify(SIMPLIFICATION_TOLERANCE_LESS)')
+            return simple
+    except Exception:
+        pass
+    # If both simplifications fail, fallback to preserve topology, which should always succeed
+    logger.debug('Neighborhood.simplify_geom used ' +
+                 'geom.simplify(SIMPLIFICATION_TOLERANCE_MORE, preserve_topology=True)')
+    return geom.simplify(SIMPLIFICATION_TOLERANCE_MORE, preserve_topology=True)
```
KlaasH: Is it intentional that this returns the result of `geom.simplify` directly whereas the other two return `MultiPolygon([geom.simplify()])`?  It causes a problem in cases where it gets to this point and the simplified geometry is just a `Polygon`.

CloudNiner: No, this was definitely an oversight. I'll update this.

CloudNiner:
Brilliant! That worked really well. I ran your code after applying the fix in the latest commit and got zero simplification errors and the same spread for each particular case. I added an extra print statement within your `set_boundary` method to return the area ratio and the point counts and they all looked sensible.

PR #564, opened 2018-04-18
Cache OSM extracts for analysis

### LINKED ISSUE #536: Remove usage of the OSM Overpass API
Our large batch runs aren't in the spirit of the good faith limits, and if we trigger jobs in batches, we get throttled, so we know we're not being good stewards.

So, what alternatives do we have?

I only see one strong contender, which is to use [AWS Lambda + Scheduled Events](http://docs.aws.amazon.com/lambda/latest/dg/with-scheduled-events.html) or a similar service to copy the state OSM extracts posted at http://download.geofabrik.de/north-america.html to an S3 bucket in the PFB AWS account on a regular schedule, that keeps them current as the extracts are updated upstream. Then, the analysis job can have some logic added to it that chooses the proper extract based on the state that each analysis is run in.

This project doesn't have the infrastructure, time, or monetary resources to attempt a more ambitious solution such as maintaining custom infrastructure, or downloading the entire world on a regular basis and cutting it ourselves.

Disadvantages to the proposed solution include a dependence on the external site in question, which would be trouble if they stopped releasing the exports. In which case we'd be back to where we started.

KlaasH:
## Overview

Adds a script to the analysis to pull OSM state extracts from [Geofabrik](http://download.geofabrik.de/north-america.html) and cache them on S3 for a week.

Features and benefits:
- Using Geofabrik extracts avoids the problems we've had with exceeding both the size and rate limits of the [Overpass API](http://overpass-api.de/).
- Caching on S3 avoids the potential problem of getting throttled by Geofabrik (not to mention wasting time) for downloading the same file for all of the analysis jobs being run in a state.
- The script uses a fairly simple lockfile system, writing the lockfile to the same bucket where the files are kept, to prevent two jobs started around the same time from both finding the file they need missing from the bucket and both downloading it.
- Cache invalidation is handled by a lifecycle rule added to the "directory" in the bucket.  It's set for 7 days, which means it could conceivably keep files around after newer versions are available on Geofabrik, but since in fact we do analysis runs much less often than that, the effect should be that the files are downloaded once per batch of jobs and are gone by the time another batch gets run.
- The script is written in python, since
   - we already have the environment set up for it, since the container installs the Django requirements
   - we already have one python script as part of the analysis ([`detect_utm_zone.py`](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/scripts/detect_utm_zone.py))
   - it's pretty self-contained--receiving 3 arguments and returning the path to the OSM extract file
   - it's complex enough that I thought I would be able to write it more easily and better in Python

### Demo

image

Here's the console output from running two copies of the script at the same time for California:
```
$ ./download_osm_extract.py -v /tmp/osm_extracts CA khoekema-pfb-storage-us-east-1 & \
  ./download_osm_extract.py -v /tmp/osm_extracts CA khoekema-pfb-storage-us-east-1 &
[1] 10200
[2] 10201
download_osm_extract.py 2018-04-18 13:23:18 DEBUG    Checking for OSM extract lockfile for ca
download_osm_extract.py 2018-04-18 13:23:18 DEBUG    Checking for OSM extract lockfile for ca
download_osm_extract.py 2018-04-18 13:23:19 DEBUG    Looking for OSM extract on S3 for ca
download_osm_extract.py 2018-04-18 13:23:19 DEBUG    Looking for OSM extract on S3 for ca
download_osm_extract.py 2018-04-18 13:23:19 DEBUG    Trying to acquire download lock for ca
download_osm_extract.py 2018-04-18 13:23:19 DEBUG    Trying to acquire download lock for ca
download_osm_extract.py 2018-04-18 13:23:49 INFO     Downloading OSM extract from Geofabrik for ca
download_osm_extract.py 2018-04-18 13:23:50 INFO     Another job got the download lock for ca
download_osm_extract.py 2018-04-18 13:23:50 DEBUG    Checking for OSM extract lockfile for ca
download_osm_extract.py 2018-04-18 13:23:50 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:24:51 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:25:51 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:26:51 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:27:52 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:28:52 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:29:11 INFO     Uploading OSM extract ca.osm.pbf to S3
download_osm_extract.py 2018-04-18 13:29:52 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:30:53 DEBUG    Lockfile exists for ca, waiting
download_osm_extract.py 2018-04-18 13:31:58 DEBUG    Lockfile exists for ca, waiting
/tmp/osm_extracts/ca.osm.pbf
download_osm_extract.py 2018-04-18 13:32:58 DEBUG    Looking for OSM extract on S3 for ca
download_osm_extract.py 2018-04-18 13:37:16 INFO     Downloaded OSM extract from S3 for ca
/tmp/osm_extracts/ca.osm.pbf
[1]-  Done   ./download_osm_extract.py -v /tmp/osm_extracts CA khoekema-pfb-storage-us-east-1
[2]+  Done   ./download_osm_extract.py -v /tmp/osm_extracts CA khoekema-pfb-storage-us-east-1
```

### Notes

- If something crashes and leaves a lockfile in the bucket, every other job for that state will crash as well.  This might be good or it might be bad.  To me it seems more good, since it avoids spending lots of time/resources processing jobs that might be doomed to fail anyway.  Obviously if finishing as much as possible regardless of errors is the desired approach, this wouldn't serve as well.
- For how often and how long to poll the lockfile, I picked every minute for 20 minutes.  A longer interval would mean jobs with smallish files would be waiting needlessly long.  I tried California on my machine at home and it took about 8 minutes to download and 3 minutes to upload.  It would likely be faster on Batch instances, but it makes sense to set a timeout that we don't expect will be hit unless something has actually gone wrong.
- This downloads the protobuf files and, conveniently, we don't need to extract them because `osmconvert`, which is the first thing that gets called on the OSM file (to clip it to the buffered boundary) supports `.osm.pbf` files automagically.

## Testing Instructions

- Create and run an analysis job (don't supply an OSM extract URL).  It should call the new script in its turn and use the resulting downloaded file for the OSM import.  Everything else should work as before.
- To test the concurrency behavior, it's easier to call the script directly, which you can do either by making a virtualenv on your host machine or using the shell you get at the end of running an analysis job.  In either case, running e.g.:
```
./download_osm_extract.py -v /tmp WV USERNAME-pfb-storage-us-east-1
```
should download the West Virginia extract from Geofabrik and write it to that bucket, and print the path.  Running the same thing again should result in it downloading the cached file from S3.
- To see the lockfile behavior, try:
```
./download_osm_extract.py -v /tmp/osm_extracts RI USERNAME-pfb-storage-us-east-1 & \
./download_osm_extract.py -v /tmp/osm_extracts RI USERNAME-pfb-storage-us-east-1 &
```
I.e. run the same command twice at the same time.  They'll both start up and check for the lockfile, then they'll both write the lockfile, then one will see that it won the race and start downloading the extract while the other will see that it lost the race and start polling for the lockfile, downloading from S3 when it eventually disappears.

- Start the script for a new state then cancel it while it's downloading from Geofabrik.  That will leave a stale lockfile in the bucket.  Run the same command and you should see it poll the lockfile for 20 minutes then raise an exception.

- I'm not sure how to test the addition of the lifecycle rule to terraform.

Resolves #536.


KlaasH:
Whoops, the planning outline says "Catch any job errors and always delete the lockfile if the job created it".  That's not what I did.  I got it in my head that "default to inaction/wait for human intervention" was better.  Should be pretty easy to switch, though.

deployment/terraform/storage.tf
```diff
@@ -32,4 +32,13 @@ resource "aws_s3_bucket" "storage" {
     Environment = "${var.environment}"
     Project     = "${var.project}"
   }
+
+  lifecycle_rule {
```
CloudNiner: This resource should be parameterized with the `environment`. I'm not sure what terraform will do when we do a prod deploy and it detects that this resource exists, but the production state never modified it. Also, it's better to keep our stack resources separate, such that staging tests don't interrupt any ongoing prod analyses.

Otherwise this config appears correct. I ran a staging deploy to the point of `./scripts/infra plan` and got:
```
[Standard task definition changes omitted]

~ aws_s3_bucket.storage
    lifecycle_rule.#:                                                    "0" => "1"
    lifecycle_rule.0.enabled:                                            "" => "true"
    lifecycle_rule.0.expiration.#:                                       "0" => "1"
    lifecycle_rule.0.expiration.3023609085.date:                         "" => ""
    lifecycle_rule.0.expiration.3023609085.days:                         "" => "7"
    lifecycle_rule.0.expiration.3023609085.expired_object_delete_marker: "" => ""
    lifecycle_rule.0.id:                                                 "" => "osm_extracts"
    lifecycle_rule.0.prefix:                                             "" => "/osm-data-cache"


Plan: 3 to add, 3 to change, 3 to destroy.
```

CloudNiner: Oops, disregard, misunderstood that this was a global rule and not already scoped within the storage bucket.

KlaasH:
Just for the record:
- Added the migration (I never manage to keep that little gotcha in my memory from one time to the next...)
- Added a `try: ... finally: ...` to delete the lockfile regardless of what happens with the download.
- As we discussed briefly, the case of an error while trying to acquire the lock is a tiny bit complicated, but I added a `try ... except ...` that deletes the lockfile before re-raising, so that should take care of that (fairly slim) potential source of stale locks.
- Also following on our discussion, I had noted that it checks for a lockfile before trying to download. I did it that way because that puts the download-from-S3 step in exactly one place, but it also adds risk in the case of lockfile weirdness.  And the download is in a function anyway, so it's very simple to try it up front.  So I changed it to do that.

CloudNiner:
👍

PR #581, opened 2018-06-27
Fixup local analysis

### LINKED ISSUE #574: Update README.RUNNING to work when AWS credentials are unconfigured
See: https://github.com/azavea/pfb-network-connectivity/blob/develop/README.RUNNING.md

PFB attempted to follow this set of instructions and had issues due to a null JOB_ID and no configured AWS credentials. 

@spencerrecneps, you made some modifications to support the local-only workflow described in this README. Could you point us to the changes you made on your fork or elsewhere to kickstart this effort?

**Done Criteria:** Allow a local-only job to be run via the instructions in README.RUNNING without having to start the dev web server or configure AWS credentials.

CloudNiner:
## Overview

This PR updates the `./scripts/run-local-analysis` script and the associated inner workings of the BNA analysis scripts to work entirely locally, with no connection to AWS or the development Django application. I was able to complete a Germantown network analysis locally without AWS credentials configured or the Django stack running.

### Notes

~~This PR does not yet incorporate any of the additional arguments provided by Spencer's replacement python script in https://github.com/azavea/pfb-network-connectivity/compare/develop...spencerrecneps:feature/better-local-analysis. It only takes the path of least resistance to a working local analysis within the current VM/Docker setup.~~ Edit: It now respects the `NB_POSTGRESQL_*` variables and supports targeting a remote DB.

## Testing Instructions

Follow the setup instructions in README.md, without configuring any AWS credentials. Then follow the instructions to run a job in README.LOCAL-ANALYSIS.md. The job should run successfully with no AWS credentials or PFB_JOB_ID configured.

If you already have a dev VM setup, and have AWS credentials configured, you can run a local analysis of the Germantown neighborhood in Philly with the following command:
```
AWS_PROFILE=doesnotexist NB_OUTPUT_DIR=/data/output/germantown-local \
    ./scripts/run-local-analysis \
    'https://andrew-pfb-storage-us-east-1.s3.amazonaws.com/neighborhood_boundaries/root/PA/germantown-philadelphia.zip' PA 42
```
In the above command, we must manually set AWS_PROFILE to an invalid profile that doesn't exist, to avoid using the correctly configured one.

Closes #574


CloudNiner:
@spencerrecneps I also just pushed a commit that theoretically means the `./scripts/run-local-analysis` respects the `NB_POSTGRESQL_*` variables that are used everywhere. So you could point to a remote instance with:
```
NB_OUTPUT_DIR=/data/output/germantown-local \
NB_POSTGRESQL_DB=remote_db_name \
NB_POSTGRESQL_HOST=remote_hostname \
NB_POSTGRESQL_USER=remote_user \
NB_POSTGRESQL_PASSWORD=remote_password \
    ./scripts/run-local-analysis \
    'https://andrew-pfb-storage-us-east-1.s3.amazona
ws.com/neighborhood_boundaries/root/PA/germantown-philadelphia.zip' PA 42
```

Can you test that and let me know if it works, copy/pasting any error messages? If it doesn't work I'll see about provisioning a remote DB, but that's not trivial for me to do. If you're unable to test, let me know as well.

CloudNiner:
@KlaasH I was able to successfully test against a remote DB with commit 584a754 by doing the following within the VM:
```
docker-compose build analysis
# Build a custom postgres container wiht all deps installed, but none of the Django, AWS or custom entrypoint stuff
docker build -f src/analysis/Dockerfile.db-alt -t pfb-db-alt ./src/analysis/
# Start new external postgres DB
docker run --name pfb-db-alt \
    -e POSTGRES_PASSWORD=pfb_alt -e POSTGRES_DB=pfb_alt -e POSTGRES_USER=pfb_alt \
    -d pfb-db-alt postgres -c 'listen_addresses=*'
# Install psql to test
sudo apt-get install postgresql-client
# Get DB container IPAddress
docker inspect <containerid gotten with \`docker ps\`> | grep -i ipaddress
PGPASSWORD=pfb_alt psql -h <db container ip address> -d pfb_alt -U pfb_alt
# If you connect successfully with the above, SUCCESS!
# Now run a local analysis
NB_OUTPUT_DIR=/data/output/germantown-remote-db \
NB_POSTGRESQL_DB=pfb_alt \
NB_POSTGRESQL_HOST=remote-postgres \
NB_POSTGRESQL_USER=pfb_alt \
NB_POSTGRESQL_PASSWORD=pfb_alt \
    ./scripts/run-local-analysis \
    'https://andrew-pfb-storage-us-east-1.s3.amazona
ws.com/neighborhood_boundaries/root/PA/germantown-philadelphia.zip' \
    PA 42 \
    '--rm --link pfb-db-alt:remote-postgres'
```
The `docker run` within `./scripts/run-local-analysis` will still start the normal postgresql db. To prove you're connected to the remote one, you can kill the one started automatically by docker with a `docker stop <containerid>` gotten from `docker ps`, or you can connect to the remote db with psql as above and verify the tables are being created/used via SQL queries.

CloudNiner:
I didn't get to test it, but here's the start of a diff that allows the analysis container entrypoint script to accept `PFB_CONNECTIVITY_ONLY` and `PFB_IMPORT_ONLY` to control which parts of the analysis are run. These args would still need to be tested and added to the docker run call. They're also only really useful in the case of a remote db connection, since the internal docker database doesn't persist between runs.

Note that its also necessary for the user to set NB_OUTPUT_SRID if PFB_CONNECTIVITY_ONLY is set, since the default output SRID is derived from the shapefile. If we don't have that, we can't get the default srid easily, without querying into the DB.

While unfortunate, it's debatable whether its worth including these new vars here in the run-local-analysis script for the most likely use case (user runs a small to medium boundary using the default docker workflow), because there's some caveats on their use and they would need to be disabled unless running against a remote instance.

CloudNiner:
I should also note that I chose not to rewrite `./scripts/run-local-analysis` in python because most of the changes were internal to the analysis scripts to get support for remote DBs and cleanup the AWS/Django interactions. It was basically just arg changes to the outer calling script, so it didn't appear there was a whole lot to gain by rewriting everything in Python. The env vars as args is a bit odd if you're not used to it, but its documented and should be clear enough to get started.

Hopefully the state fips helper flag is addressed by the link to the table of FIPS codes added to the documentation.

KlaasH:
Re testing with remote DB:
Changing the command in `run-local-analysis` from `docker-compose run` to `docker run` means it doesn't get the `volumes:` specified in the `docker-compose.yml`, so there's no linked `/data` directory inside the container.  There's a `mkdir -p` command that creates it, so the analysis runs fine because it can create and write to that directory inside the container, but it doesn't make it out onto the host filesystem.

So I changed back to using `docker-compose run`.

I got a "remote DB" test working similarly to yours, but I built and ran the alternate DB on my host machine, with a separate docker-compose config file and with port 5432 exposed on the host, then I used `NB_POSTGRESQL_HOST=10.0.2.2` to access it from the analysis container (which has access to the VM, which in turn uses that virtual IP to access services on the host).

I actually got it working first with the base Azavea PostGIS container which meant it was missing a few plugins, but that was enough to confirm that it was using the overrides and not trying to look for the default database (I also just commented out the `links:` section of the `analysis` config in `docker-compose.yml` to prevent it from starting any other containers).  When I tried to swap in your modified Dockerfile I had problems connecting to it, but I decided to call it good.

I used the same command as in your example, but without the pass-through arguments for `docker` at the end.


Other than that, I:
- Added a bit of provisioning to upgrade SSL-related packages, since my attempt to provision from scratch failed due to old-Python/SSL issues.
- Removed the bit about a default value for `PFB_JOB_ID` from the usage message of `run-local-analysis`. It had already been removed from the code and the README.
- Switched the example shapefile in the remote-database example to the copy in `test-pfb-inputs` bucket.
- Added a timestamped directory to the output path for jobs with no `PFB_JOB_ID`. That was formerly the default value for the job ID, which had the advantage of creating namespaced output directories and preventing accidental clobbering or mingling of results. It makes sense to only have a `PFB_JOB_ID` when there's an actual job, but that doesn't mean we can't still do output namespacing for local analysis runs.

I added my commits onto this branch.  I reckon I'll go ahead and merge it once the build is green.

PR #585, opened 2018-10-05
Water removal and residential speed defaults

beckymasond:
## Overview

The changes include added functionality to:
1) Remove census blocks that contain no land area, which will improve the BNA maps by only visualizing blocks that have a BNA score. Block removal is based on a csv listing all 2010 Decennial census blocks that have zero land area.
2) Set default speed limits for residential streets based on the state, to account for states where the legislated residential speed limit is greater than 25 mph.
3) Set default speed limits for residential streets based on the city (takes precedence over state), to account for cities where the legislated residential speed limit is different than the BNA default of 25 mph or different than the state default (e.g. Georgia has a 30 mph residential limit, but the city of Dunwoody, GA has a 25 mph residential limit).

### Demo

Original Dunwoody, GA heat map:

image

Dunwoody, GA heat map with water removed and state default residential speed implemented (30 mph):

image

Dunwoody, GA heat map with water removed, state default, and city default residential speed implemented (25 mph):

image

### Notes

* All external files (three csv files) are stored as public files on S3. 
* The city defaults require an additional input term, the city FIPS code, as an identifier.
* As of 10/5, running the analysis locally is failing due to https://github.com/azavea/ansible-pip/issues/9

## Testing Instructions

* Run local analysis for Dunwoody, GA with PFB_CITY_FIPS = 1324768
* Run local analysis for Dunwoody, GA without PFB_CITY_FIPS
* Compare results to https://bna.peopleforbikes.org/#/places/d8dfa19c-8048-46ad-ab01-80760fb21b27/
* See images of expected results above


src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
...
+   -c "\copy city_speed FROM ${CITY_SPEED_DOWNLOAD} delimiter ',' csv header"
+
+# Set default residential speed for city
+CITY_DEFAULT=$( psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+      -t -c "SELECT city_speed.speed FROM city_speed WHERE city_speed.fips_code_city = ${PFB_CITY_FIPS}" )
+# Check if no value for city default, if so set to NULL
+echo "The city residential default speed is ${CITY_DEFAULT}."
+if [[ -z "$CITY_DEFAULT" ]];
+then
+  echo "No default residential speed in city"
```
flibbertigibbet: The indentation should be four spaces, not two, for consistency with the rest of the bash scripts.

src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
...
+            city varchar,
+            state char(2),
+            fips_code_city integer,
+            speed smallint
+        );"
+
+# Import city residential speeds file
+CITY_SPEED_FILENAME="city_fips_speed"
+CITY_SPEED_DOWNLOAD="/data/${CITY_SPEED_FILENAME}.csv"
+wget -nv -O "${CITY_SPEED_DOWNLOAD}" "https://s3.amazonaws.com/pfb-public-documents/${CITY_SPEED_FILENAME}.csv"
```
flibbertigibbet: Similar to [the check for a local Census blocks file](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/import/import_neighborhood.sh#L80), checking first here for a local copy of the city speeds file will allow us to add or override city speeds for testing. As the current file only has a few entries, we will probably want to do so.

src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
...
+psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+   -c "\copy city_speed FROM ${CITY_SPEED_DOWNLOAD} delimiter ',' csv header"
+
+# Set default residential speed for city
+CITY_DEFAULT=$( psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+      -t -c "SELECT city_speed.speed FROM city_speed WHERE city_speed.fips_code_city = ${PFB_CITY_FIPS}" )
```
flibbertigibbet: The FIPS parameter will need to be quoted here to treat it as a string.

src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
 # Delete downloaded temp OSM data
 rm -rf "${OSM_TEMPDIR}"
 
+# Create table for state residential speeds
+echo 'START: Importing State Default Speed Table'
+psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+     -c "CREATE TABLE IF NOT EXISTS \"state_speed\" (
+            state char(2),
+            fips_code_state smallint,
```
flibbertigibbet: The FIPS code here should be stored as a `char` type.

src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
...
+      -t -c "SELECT state_speed.speed FROM state_speed WHERE state_speed.fips_code_state = ${PFB_STATE_FIPS}" )
+echo "DONE: Importing state default residential speed"
+
+# Create table for city residential speeds
+echo 'START: Importing City Default Speed Table'
+psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+     -c "CREATE TABLE IF NOT EXISTS \"city_speed\" (
+            city varchar,
+            state char(2),
+            fips_code_city integer,
```
flibbertigibbet: The FIPS code here should be stored as a `char` type. Like zip codes, they are identifiers with significant leading zeroes, rather than numbers we would do math with, and they are formed by concatenation.

src/analysis/import/import_osm.sh
```diff
@@ -145,6 +147,56 @@ osm2pgsql \
...
+            fips_code_state smallint,
+            speed smallint
+        );"
+
+# Import state residential speeds file
+STATE_SPEED_FILENAME="state_fips_speed"
+STATE_SPEED_DOWNLOAD="/data/${STATE_SPEED_FILENAME}.csv"
+wget -nv -O "${STATE_SPEED_DOWNLOAD}" "https://s3.amazonaws.com/pfb-public-documents/${STATE_SPEED_FILENAME}.csv"
+psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+   -c "\copy state_speed FROM ${STATE_SPEED_DOWNLOAD} delimiter ',' csv header"
```
flibbertigibbet: Nitpick: there's three spaces here; this should visually line up with the `-h` above it.

PR #586, opened 2018-10-17
Removing pedestrian paths and fixing one-way evaluation

beckymasond:
## Overview

These changes:
(1) Remove paths from the data set if they are tagged "bicycle=no". These are primarily pedestrian paths. By removing them from the data set, they are not scored and are not visualized.
(2) Corrects errors in the 'lanes.sql' script. Due to syntax errors, lanes for one-way streets were being calculated by taking the total lane count and dividing it by two. This led to lanes being counted in both directions for one-way streets and incorrect stress designations as a result.

### Demo

(1) Before: Pedestrian paths included in a section of Boulder, CO
boulder_pedpath

After: Pedestrian paths removed
boulder_nopedpath

(2)
Before: One-way segments with error in logic, resulting in the circled area having a mix of high and low stress segments. These segments of Folsom St. in Boulder (between Pearl St. and Spruce St.) are each one-way and have two or three lanes of traffic adjacent to a bike lane with a speed limit of 30 mph.
folsom_oneway_bef

After: One-way segments with corrected logic, resulting in the highlighted area having only high stress segments.
folsom_oneway_aft

## Testing Instructions

* Run analysis for Boulder, CO with changes
* Compare BNA for Boulder on public website with results
* Outcome: Walking trails are removed
* Outcome: Folsom Street north of Pearl St. changes from low and high stress segments to only high stress segments

### DIFF
--- a/src/analysis/features/lanes.sql
+++ b/src/analysis/features/lanes.sql
@@ -15,7 +15,7 @@ SET     ft_lanes =
                                     ),
                                     1       -- only one dimension
                                 )
-                    WHEN osm."turn:lanes" IS NOT NULL AND one_way = 'ft'
+                    WHEN osm."turn:lanes" IS NOT NULL AND osm."oneway" IN ('1', 'yes')
                         THEN    array_length(
                                     regexp_split_to_array(
                                         osm."turn:lanes",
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
Closing in favor of #682 to appease CI tests.

Issue #591, opened 2018-11-15
Copy over Tilegarden tiler code and get it serving tiles locally


KlaasH:
Depends on #587 (not to start, but at least to finish)

Elements:
- Docker-compose setup to run `claudia-local-api` server and link to database
- Tile config MML/styles
- Begin to integrate into `scripts/` (`update`, `server`, etc)

PR #599, opened 2018-11-23
Feature/kak/store analysis geometries#587

### LINKED ISSUE #587: Load analysis geometries into AnalysisJob model
This task is required to support admin user uploaded analysis results.

- [ ] Add appropriate geom fields to the AnalysisJob model
- [ ] Write migration that iterates existing AnalysisJobs and loads the geometries from the shapefile urls we've already stored

Consider that in follow up tasks we'll need to perform the same shapefile -> geom field task when we load in uploaded user jobs.

flibbertigibbet:
## Overview

Adds two geometry fields to the `AnalysisJob` model for neighborhood ways and Census block results. Also adds a migration to read back the stored results shapefiles from S3 for existing analysis jobs, extract the geometries, and write them to the new fields as singular, multi-type geometries.

### Demo

Results for Germantown.

Neighborhood ways after re-import:

image

As generated by the tiler, for comparison:

image

Census blocks after re-import:

image

Census blocks on the map. Clipped or with transparent styling on the edges?

image

### Notes

I believe this does what #587 calls for, but not what #591 needs for tiling results from the database.

For tiling results, I think we should instead add separate models for each of the results shapefiles that store one object for each geometry within the shapefile (i.e., store each line string or polygon separately, instead of combining them into a single multi-type geometry), and also read back the properties from the shapefile to each object, which will be necessary to style them individually based on the results, as that cannot be accomplished with a single multi-type geometry.

I noticed there is also a bike infrastructure tiling style for the current tiler job, so I'm mentioning that here in case there's a third results shapefile that should be getting ingested as well.

Another note: multipolygons formed from the constituent set of polygons and sub-multipolygons for the Census blocks on import are either not valid due to self-intersection, or if made valid either with a zero buffer or the `MakeValid` function, results in dissolving all internal bounds. So, the method here went with forming invalid, self-intersecting multipolygons, which still can be read and look okay when viewed in QGIS.

Although this has a migration for reading back previous job results, it does not add a means for storing the results geometries at the time of a new job run, as that was not called for in #587.

Initially attempted using `fiona` for the migration import, as that is used elsewhere in the project for similar imports, but ran into issues building the multi-type geometries reliably, so switched to Django's in-built GDAL wrapper for imports, [LayerMapping](https://docs.djangoproject.com/en/2.1/ref/contrib/gis/layermapping/), which will likely serve well for direct use importing each sub-geometry and its properties to new models for each results type, as it is built for that purpose; doing so will probably be a simpler operation that building the multi-type geometries as done here.

## Testing Instructions

 * Before checking out this PR, run one or more analysis and tiler jobs locally
 * Check out this PR and run migrations
 * Should succeed in saving geometries to the `AnalysisJob`
 * To inspect, export the geometries from the job by dumping them as geojson
 * Resulting geojson shapes should look like those generated by the current tiler

Closes #587


flibbertigibbet:
Closing in favor of #600, which is this branch with a couple of extra commits.

Made a new branch and PR to keep the writeup here, which does not apply to the new PR.

PR #600, opened 2018-11-26
Feature/kak/store analysis results#587

### LINKED ISSUE #587: Load analysis geometries into AnalysisJob model

flibbertigibbet:
## Overview

Supersedes #599.

Adds two models for the two exported shapefiles, to store each geometry in the shapefile as a separate object, along with its properties necessary for map tiling.

### Demo

Inspecting created objects at the console:
```
>>> NeighborhoodWaysResults.objects.first().geom.json
'{"type": "LineString", "coordinates": [[-106.336278600001, 42.8248898000048], [-106.335824300001, 42.8249606000048], [-106.335458400001, 42.8250697000048], [-106.335032200001, 42.8252319000048], [-106.334606, 42.8254471000048]]}'
>>> CensusBlocksResults.objects.first().geom.json
'{"type": "MultiPolygon", "coordinates": [[[[-106.35391, 42.7742500000048], [-106.35367, 42.7746600000048], [-106.35348, 42.7749700000048], [-106.35334, 42.7751500000048], [-106.35306, 42.7750300000048], [-106.35305, 42.7749500000048], [-106.35318, 42.7747900000048], [-106.35328, 42.7746100000048], [-106.35336, 42.7744800000048], [-106.35336, 42.7743700000048], [-106.35341, 42.7743], [-106.35352, 42.7741900000048], [-106.35359, 42.7740600000048], [-106.35362, 42.7739500000048], [-106.35362, 42.7737900000048], [-106.35379, 42.7738], [-106.3541, 42.7738800000048], [-106.354082, 42.7739130000048], [-106.35391, 42.7742500000048]]]]}'
>>> CensusBlocksResults.objects.first().job      
<AnalysisJob: <AnalysisJob: COMPLETE Casper>>
>>> NeighborhoodWaysResults.objects.first().job      
<AnalysisJob: <AnalysisJob: COMPLETE Casper>>
```

### Notes

`LayerMapping` ([source](https://github.com/django/django/blob/master/django/contrib/gis/utils/layermapping.py)) does not provide a means for setting model fields that do not come from the shapefile, or to get at the objects created from the shapefile, but we need to set the related job on the objects to be able to identify them later. To deal with this, the data migration assumes any created objects without a job set already came from the last imported shapefile, and set the job to the current job. This will work for the migration, as we can expect it to run serially and for no other operations to happen on the newly added models at the same time, but we will need to find some other way to set the `job` in whatever later task handles adding these objects at the time the shapefile-creating analysis job is run. The easiest solution for future analysis jobs may be to simply add a field to the exported shapefile for the job UUID, which could then be used as part of the field mapping.

## Testing Instructions

 * Run migrations
 * Go to Django console
 * Expected properties should exist on the created objects
 * All fields used by the current tiler stylings should exist on the new models

Closes #587


src/django/pfb_analysis/migrations/0032_reimport_results.py
```diff
@@ -0,0 +1,95 @@
...
+    with zipfile.ZipFile(local_zipfile, 'r') as zip_handle:
+        zip_handle.extractall(tmpdir)
+    return tmpdir
+
+
+def s3_job_url(job, filename):
+    return 'results/{uuid}/{filename}'.format(uuid=job.uuid, filename=filename)
+
+
+def add_results_geoms(apps, schema_editor):
```
CloudNiner: This core of this data migration is really a reusable piece of code that "imports the geometries for a single analysis job" for each analysis job in the database. Like many other data migrations we've written in the past few months, ideally I'd like to see this migration written as a management command with the core functionality that does the geom import for a single analysis job abstracted to either an AnalysisJob model method (`import_geometries_from_shapefile()`?) or a utility method. 

This provides reusability with very little extra effort (writing the wrapper management command). For instance it would be easier to add support for the job id issue you mentioned in notes without leaving this code orphaned. It would also allow us to easily add management command options for updating individual jobs or a range of jobs (instead of the default of all jobs) in case of failure when we run this code on production.

What do you think?

src/django/pfb_analysis/management/commands/import_results_shapefiles.py
```diff
@@ -0,0 +1,100 @@
...
+        if ways_tmpdir:
+            shutil.rmtree(ways_tmpdir, ignore_errors=True)
+
+
+class Command(BaseCommand):
+    help = "Import back results and geometries from exported shapefiles for an analysis job"
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('job_id')
```
CloudNiner: Could we tweak this so that by default with no args, this command runs for all AnalysisJob objects in the DB? And if the `--job-id` is provided it will run only for the provided id?

flibbertigibbet:
Updated.

src/django/pfb_analysis/management/commands/import_results_shapefiles.py
```diff
@@ -0,0 +1,110 @@
...
+
+class Command(BaseCommand):
+    help = """Import back results and geometries from exported shapefiles for an analysis job.
+
+    If job UUID not specified, will run for all analysis jobs.
+    """
+
+    def add_arguments(self, parser):
+        # Positional arguments
+        parser.add_argument('job_id', nargs='?')
```
CloudNiner: Nice use of nargs.

PR #607, opened 2018-11-29
Add Job ID to shapefile export

### LINKED ISSUE #601: Store job UUID on results shapefiles
Add the `AnalysisJob` UUID as a field to the results shapefiles for neighborhood ways and Census blocks, so that it may be read out using `LayerMapping` on re-import.

See #600 for more discussion.

CloudNiner:
## Overview

This PR Includes job id in the shapefile import management
command and runs the command during analysis export.

This will slow down analysis jobs but not signficantly.

Jobs should run far enough apart that it doesn't lead
to excessive DB load during a large analysis run (but
this is untested).

While inefficient because we need to re-download the
shapefile we just uploaded, the shapefile sizes aren't
huge and it's cleaner than attempting to include
custom logic in bash to directly copy data from the
analysis DB to the app DB.

#### Django job

screen shot 2018-11-29 at 9 44 01 am

#### Local job

screen shot 2018-11-29 at 9 42 03 am

## Testing Instructions

Checkout branch then rebuild analysis VM with `docker-compose build analysis`.

Trigger a new job via the PFB web admin UI, then run the job via the command printed to the docker console. It should succeed, and you should have NeighborhoodWaysResults and CensusBlocksResults objects with the Job ID you just ran in your django DB. This can be checked by running `./scripts/django-manage shell_plus` then:
```
CensusBlocksResults.objects.filter(job='<job_uuid>').count()
NeighborhoodWaysResults.objects.filter(job='<job_uuid>').count()
```

You should also test that local analysis jobs disconnected from the Django site still work correctly. The JOB_ID field will still be added to the output shapefiles, but it will be blank for all rows.

An example run:
```
NB_OUTPUT_DIR=/data/outputs/test-run-local-job-id ./scripts/run-local-analysis 'https://andrew-pfb
-storage-us-east-1.s3.amazonaws.com/neighborhood_boundaries/root/PA/germantown.zip' PA 42
```

Closes #601

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,8 @@ services:
     build:
       context: ./src/nginx
       dockerfile: Dockerfile
+    depends_on:
+      - django
     ports:
       - "9200:443"
     links:
--- a/src/analysis/scripts/export_connectivity.sh
+++ b/src/analysis/scripts/export_connectivity.sh
@@ -50,6 +50,20 @@ function ec_export_table_shp() {
     OUTPUT_DIR="$1"
     EXPORT_TABLENAME="$2"
 
+    psql -h "${NB_POSTGRESQL_HOST}" \
+         -U "${NB_POSTGRESQL_USER}" \
+         -p "${NB_POSTGRESQL_PORT}" \
+         -d "${NB_POSTGRESQL_DB}" \
+         -c "ALTER TABLE ${EXPORT_TABLENAME} ADD COLUMN IF NOT EXISTS job_id uuid;"
+    if [ -n "${PFB_JOB_ID}" ];
+    then
+        psql -h "${NB_POSTGRESQL_HOST}" \
+            -U "${NB_POSTGRESQL_USER}" \
+            -p "${NB_POSTGRESQL_PORT}" \
+            -d "${NB_POSTGRESQL_DB}" \
+            -c "UPDATE ${EXPORT_TABLENAME} SET job_id = '${PFB_JOB_ID}';"
+    fi
+
     FILENAME="${OUTPUT_DIR}/${EXPORT_TABLENAME}.shp"
     pgsql2shp -h "${NB_POSTGRESQL_HOST}" \
               -u "${NB_POSTGRESQL_USER}" \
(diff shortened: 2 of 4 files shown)


flibbertigibbet:
We should probably add some sort of guard for attempting import on old jobs that don't have the `JOB_ID` set. The resulting error is:

```
[2018-11-29 16:47:04 +0000] [48] [ERROR] Error importing results shapefiles for job: e354ec29-2924-4aaf-a6b7-d77fd9cbcfc4
Traceback (most recent call last):
  File "/usr/src/pfb_analysis/management/commands/import_results_shapefiles.py", line 67, in add_results_geoms
    CENSUS_BLOCK_LAYER_MAPPING)
  File "/usr/local/lib/python2.7/site-packages/django/contrib/gis/utils/layermapping.py", line 107, in __init__
    self.check_layer()
  File "/usr/local/lib/python2.7/site-packages/django/contrib/gis/utils/layermapping.py", line 236, in check_layer
    idx = check_ogr_fld(ogr_field)
  File "/usr/local/lib/python2.7/site-packages/django/contrib/gis/utils/layermapping.py", line 187, in check_ogr_fld
    raise LayerMapError('Given mapping OGR field "%s" not found in OGR Layer.' % ogr_map_fld)
LayerMapError: Given mapping OGR field "JOB_ID" not found in OGR Layer.
```

flibbertigibbet:
After removing the `update` calls to set the job from the import script, the job does get set from the layer mapping:
```
>>> NeighborhoodWaysResults.objects.filter(job__uuid='b9e5d945-8d4a-4ebd-b522-9db23fd1c01e').count()
6144
>>> CensusBlocksResults.objects.filter(job__uuid='b9e5d945-8d4a-4ebd-b522-9db23fd1c01e').count()    
1740
```
And it's present here on the neighborhood ways shapefile:

pfb_jobid_shapefile_feature

CloudNiner:
Updated under the wire.

I tweaked the code to handle both cases, first trying with job and then falling back to it removed from the layer config, since at least in the interim we have to handle both cases. The former for new jobs and the latter when we update all of them existing production jobs.

CloudNiner:
Reminder to me to rebase to remove WIP from commit log.

CloudNiner:
👍 good catch I'll add that before merging.

CloudNiner:
Also updated the shapefile import code to delete existing objects that match the job ID before importing new ones, otherwise we end up with duplicates if we run this for a given job ID more than once.

Issue #609, opened 2018-11-30
Add filtering by job ID to Tilegarden endpoint


KlaasH:
The Tilegarden endpoint is producing tiles from the `NeighborhoodWaysResults` and `CensusBlocksResults` tables, but it's not filtering by job ID, so if there are multiple jobs for a given location (including different runs for the same neighborhood over time) it will get all the geometries and superimpose them.

So we need to filter by job ID.  This isn't actually something that Tilegarden currently supports--it supports a `layers` param to select among pre-defined layers, but not a param to filter the actual database query.

I think it should be a path parameter, since it's required (and also because of https://github.com/azavea/tilegarden/issues/126#issuecomment-441776311) and we'll need to dynamically modify the table definition in the config to apply it.  I.e. make the table into a query like in https://github.com/azavea/tilegarden/blob/develop/src/tiler/src/config/point-example.mml#L20 and modify it at runtime to supply the job ID value.

Issue #595, opened 2018-11-15
Deploy Tilegarden tiler to staging


KlaasH:
Flesh out any configuration needed for remote, multi-environment deployment that wasn't necessary for local development.

jeancochrane:
Just had a nice meeting with @KlaasH to get some more context for this issue. Rounding up the next steps for us:

- Figure out the best way to manage templating environment variables in [`./scripts/template-vars.js`](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/tilegarden/scripts/template-vars.js)
- Refactor compilation/deployment commands in [`package.json`](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/tilegarden/package.json)
- Stand up staging/production deployments
- Hook up Tilegarden deployments to existing PFB CI

Issue #594, opened 2018-11-15
Configure CloudFront proxy for tile endpoint


KlaasH:
API gateway can't serve tiles directly, so we need to configure a CloudFront proxy to add an `Accept: image/png` header.  See https://github.com/azavea/tilegarden/issues/19

Issue #593, opened 2018-11-15
Set up remote development Tilegarden endpoint using Staging database


KlaasH:
We need to be able to test the tiler running in the actual AWS environment but without having to merge to develop and deploy staging.  So we need to be able to run a remote development Tilegarden endpoint, pointed at the Staging database.  (It might still be limiting to not be able to test database schema changes remotely in development, but that would be much harder to accomplish.)

hectcastro:
In Raster Foundry, we use the `test/` branch prefix to trigger automatic deployments to staging. That allows us to more easily test things in staging without having to merge into `develop`. Afterwards, re-triggering a `develop` build wipes the `test/` branch contents away.

KlaasH:
Looking at this again, that "build `test/` branches" strategy sounds pretty good.  Would that just be a matter of changing [the Jenkinsfile](https://github.com/azavea/pfb-network-connectivity/blob/0.8.1/Jenkinsfile#L32) and adding `test*` to the "Include" filter under "Filter by name (with wildcards)" in the repo configuration on Jenkins?

hectcastro:
Yeah, roughly https://github.com/raster-foundry/raster-foundry/commit/bb787d09a2a5d1fad0d3094b071df013d499721b and the Jenkins org (you can't tweak settings for individual repos) configuration change to inculde `test/*`.

PR #610, opened 2018-11-30
Add tilegarden

### LINKED ISSUE #591: Copy over Tilegarden tiler code and get it serving tiles locally

KlaasH:
## Overview

Copies over enough of [Tilegarden](https://github.com/azavea/tilegarden) to get a tiler service running in development and serving census blocks, ways, and bike infrastructure tiles (though not filtered by job).

It doesn't copy over any of the configuration code (i.e. [`.env.template`](https://github.com/azavea/tilegarden/blob/develop/.env.template) and the [terraform config](https://github.com/azavea/tilegarden/tree/develop/src/terraform)), since the plan is to integrate at least some of that into the existing deployment infrastructure for this project.  So it's development-only at the moment.  There are a few remnants (e.g. `deploy` and related commands in the `project.json` file) that are in what was copied but won't currently work.

Still to do:
- Add filtering by job (#609)
- Modify the tile URLs returned by the API (#592)
- Get it deploying (i.e. adapt and integrate Tilegarden's deployment process into the scripts and config for this project) (#593, #594, #595)

### Demo

image

### Notes

See the commit messages for some additional notes and details.

## Testing Instructions

If you haven't run the migration from PR #600, run that first.  If you don't have any completed analysis jobs at all, you'll need to run one to have some data to look at.

- Run `scripts/setup` or (in the VM) `scripts/update` to build, then `scripts/server`
- Go to the front-end app (http://localhost:9301), click through to a neighborhood, and show one of the layers listed under "Overlays" in the layer picker.
- Open up the debug console and pick the "Network" tab, then move or zoom the map to get it to request some new tiles.
- Copy the URL of one of the tile requests from your development S3 tile source, then replace everything up to the first number (i.e. the Z value in the tile path) with http://localhost:9400/tile/, e.g. http://localhost:9400/tile/14/4770/6200.png.
- That request will fail, telling you you need to supply a config parameter.  Add `?config=ways`, `?config=census_blocks`, or `?config=bike_infrastructure` to see some actual tiles.

Also, `scripts/test` should run Tilegarden's tests in addition to the Django tests, and should pass.

Resolves #591.


src/tilegarden/package.json
```diff
@@ -29,7 +29,7 @@
     "destroy": "claudia destroy",
     "dev": "nodemon -e js,mss,json,mml,mss --ignore bin/ --ignore '*.temp.mml' --exec yarn local",
     "lint": "eslint src",
-    "local": "yarn transpile && node --inspect=0.0.0.0:9229 -- node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module bin/api | bunyan -o short",
+    "local": "yarn transpile && node --inspect=0.0.0.0:9401 -- node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module bin/api | bunyan -o short",
```
CloudNiner: To avoid unnecessary changes to the tilegarden diff, might want to just leave this one as is and make the docker port mapping `9401:9229`.

src/tilegarden/src/config/map-config.mml
```diff
@@ -12,10 +12,10 @@ Layer:
   geometry: polygon
   srs: "+proj=lcc +lat_1=40.96666666666667 +lat_2=39.93333333333333 +lat_0=39.33333333333334 +lon_0=-77.75 +x_0=600000 +y_0=0 +ellps=GRS80 +datum=NAD83 +to_meter=0.3048006096012192 +no_defs"
   Datasource:
-    host: ${TILEGARDEN_HOST}
-    dbname: ${TILEGARDEN_DB}
-    user: ${TILEGARDEN_USER}
-    password: ${TILEGARDEN_PASSWORD}
+    host: ${PFB_DB_HOST}
```
CloudNiner: This is a far reaching change that impacts upstream changes when we can just map the environment variable in the docker compose config. Are we sure we want to do this now?

The env munging change seems separate, and ok.

KlaasH: The config files where this is used are already project-specific.  I think adding generality to how configuration is assembled (especially not having to repeat the whole database config for every layer) is probably something that should happen in Tilegarden, but as it stands right now changing these variables doesn't amount to changing anything very general.  So I think the benefit of not having two copies of these variables to define and keep in sync in PFB is enough that it's worth switching.

KlaasH:
> There's quite a lot changed already in here wrt upstream tilegarden. Are we keeping a log of those changes somewhere for future reference?

Good point.  I just made https://github.com/azavea/tilegarden/issues/129 as a place to keep an annotated list of what changes we're making and what they mean for future development on Tilegarden proper.

I also switched the `inspect` port back (I was going to rebase it in, but the port switch came before the removal of transpilation, and it didn't seem worth the trouble to resolve the conflicts).

src/tilegarden/src/config/map-config.mml
CloudNiner: Makes sense.

KlaasH:
> Looks like a missing eslint configuration in the tilegarden source directory?

Whoops, had it locally but had neglected to add it.  Since I was rebasing that in anyway, I went ahead and resolved the conflicts to make it look like the `inspect` port was never changed...

Issue #612, opened 2018-12-04
Imported census block results geometries include the buffer area as all zero


KlaasH:
The tiles generated from the imported census block scores all have a big red boundary around the neighborhood being analyzed.  E.g.:

![image](https://user-images.githubusercontent.com/6598836/49449801-b00d5e80-f7a9-11e8-8e93-a6fa8d3b7e7c.png)

The same neighborhood on the production site shows:
![image](https://user-images.githubusercontent.com/6598836/49449830-c2879800-f7a9-11e8-8ddc-45b0240f651a.png)

I initially thought that must be something in the shapefile that we were formerly throwing away when tiling, but I don't see anything in [the `tilemaker` script](https://github.com/azavea/pfb-network-connectivity/blob/0.7.0/src/tilemaker/scripts/run_tilemaker.sh#L58-L62) that would have that effect.  So now I'm wondering if there's a distinction between 0 and "no score" in the shapefiles that "just worked" for tilemaker but isn't working how we want for the geometry import.

The "ways" results also extend outside the boundary to the edge of the buffer, but that's fine because the scores are still correct in the buffer area (I compared Germantown to Philadelphia and they were the same).

flibbertigibbet:
Lower Manhattan, without all-water Census blocks removed:

![image](https://user-images.githubusercontent.com/960264/49461632-1a31fd80-f7c2-11e8-8f8e-f94fd2ebb650.png)


Lower Manhattan, with water Census blocks removed (#585):

![image](https://user-images.githubusercontent.com/960264/49461675-40f03400-f7c2-11e8-9ebd-b32f27db80b6.png)

flibbertigibbet:
On further investigation, the issue seems to be that the new map styles erroneously include zeroes in the same bin as the smallest non-zero values, where the old style smallest bin excluded zeroes.

KlaasH:
I read that Esri link as saying that shapefiles _do_ support nulls and talking about how to handle that when converting to Esri's "coverage" format.

But yeah, `LayerMapping` is not observing that distinction, nor does there seem to be any way to tell it to do so.

flibbertigibbet:
[Here](http://resources.esri.com/help/9.3/ArcGISDesktop/com/Gp_ToolRef/geoprocessing_tool_reference/geoprocessing_considerations_for_shapefile_output.htm) is some further documentation from ESRI on how nulls are not supported in Shapefiles.

(see "When should I not use a shapefile?")

[...]

PR #613, opened 2018-12-04
Enable Tilegarden tiler and add job_id filtering

### LINKED ISSUE #592: Enable using non-S3 tile endpoints
Adjust `AnalysisJob.tile_urls` to be able to be configurable to return either the current S3 URLs or different endpoints (i.e. Tilegarden URLs).

### LINKED ISSUE #609: Add filtering by job ID to Tilegarden endpoint

KlaasH:
## Overview

Switches to using the Tilegarden tiler in development.  Specifically:
- Adds a `job_id` path parameter to the Tilegarden route
- Moves the `config` parameter out of the querystring and into the path
- Changes `AnalysisJob.tile_urls` to return Tilegarden endpoints as tile URLs rather than the S3 ones

### Demo

image

### Notes

- There was already a syntax for applying arbitrary filters, but it's [pretty complicated](https://github.com/azavea/tilegarden#client-specified-filtering) and relies on querystring parameters, which could [make it hard to add S3 caching](https://github.com/azavea/tilegarden/issues/126#issuecomment-441776311).
- I considered removing the `layers` filtering, since we're not using it, but 1) that would have made these changes particularly hard to reconcile with upstream Tilegarden and 2) it (or a slightly modified version of it) could be something we want to add in the future, e.g. for overlaying multiple geometry layers.
- The code for extracting path parameters uses a list of known parameters that it should ignore and tries to turn any other path parameters into query filters.  This seems a bit weird and brittle, but it also means that multiple paths with different path parameters could easily share the same function.  Passing a list of expected parameters to the function might be more sensible, but it would mean some double-entry (since they're already defined in the route's path pattern).
- Moving `config` from the querystring to the path wasn't called for on the issue, but it was quite easy to do and gets us to the point of the tile URLs being all path, no querystring.
- I wrote on issue #592 that `AnalysisJob.tile_urls` should "be configurable to return either the current S3 URLs or different endpoints", but I don't remember why I said that and it doesn't make much sense to me know.  This just switches to Tilegarden URLs and removes the old ones.
- The red area outside the analysis boundary that I had thought was an overlay of scores from a different analysis turned out to be something else, about which I made issue #612.

## Testing Instructions

- As on PR #610, if you haven't run the migration from PR #600, run that first.  If you don't have any completed analysis jobs at all, you'll need to run one to have some data to look at.
- Run `scripts/setup` or (in the VM) `scripts/update` to build, then `scripts/server`
- Go to the front-end app (http://localhost:9301) and click through to a neighborhood.
- All the tile layers should render.

## Checklist

- [ ] Update https://github.com/azavea/tilegarden/issues/129 with relevant changes (to do after merging when the commits are finalized)

Resolves #592 
Resolves #609

### DIFF
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -557,8 +557,8 @@ class AnalysisJob(PFBModel):
     def tile_urls(self):
         return [{
             'name': layer,
-            'url': self._s3_url_for_result_resource('tiles/neighborhood_{}'.format(layer) +
-                                                    '/{z}/{x}/{y}.png')
+            'url': (settings.TILEGARDEN_ROOT + '/tile/{z}/{x}/{y}.png' +
+                    '?config={layer}'.format(layer=layer))
         } for layer in ['ways', 'census_blocks', 'bike_infrastructure']]
 
     @property
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -557,7 +557,8 @@ class AnalysisJob(PFBModel):
     def tile_urls(self):
         return [{
             'name': layer,
-            'url': (settings.TILEGARDEN_ROOT + '/tile/{z}/{x}/{y}.png' +
+            'url': (settings.TILEGARDEN_ROOT + '/tile/{job_id}/'.format(job_id=self.uuid) +
+                    '{z}/{x}/{y}.png' +
                     '?config={layer}'.format(layer=layer))
         } for layer in ['ways', 'census_blocks', 'bike_infrastructure']]
 
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -557,9 +557,12 @@ class AnalysisJob(PFBModel):
     def tile_urls(self):
         return [{
             'name': layer,
-            'url': (settings.TILEGARDEN_ROOT + '/tile/{job_id}/'.format(job_id=self.uuid) +
-                    '{z}/{x}/{y}.png' +
-                    '?config={layer}'.format(layer=layer))
+            'url': '{root}/tile/{job_id}/{layer}/{tile_template}'.format(
+                root=settings.TILEGARDEN_ROOT,
+                job_id=self.uuid,
+                layer=layer,
+                tile_template='{z}/{x}/{y}.png',
+            )
         } for layer in ['ways', 'census_blocks', 'bike_infrastructure']]
 
     @property
--- a/src/tilegarden/src/util/xml-tools.js
+++ b/src/tilegarden/src/util/xml-tools.js
@@ -1,0 +1,25 @@
+/* Helper functions to encapsulate converting to and from XML.
+ *
+ * There's not much here, but it does get reused, and having it centralized should make it
+ * easier to change libraries or settings if necessary.
+ */
+
+const { promisify } = require('util')
+const xml2js = require('xml2js')
+
+// Make an async-friendly version of the parser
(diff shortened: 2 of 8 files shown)


flibbertigibbet:
I just scheduled a CI rebuild, since it seemed to have run into an issue pulling a docker image.

src/django/pfb_analysis/models.py
```diff
@@ -557,8 +557,11 @@ def destinations_urls(self):
     def tile_urls(self):
         return [{
             'name': layer,
-            'url': self._s3_url_for_result_resource('tiles/neighborhood_{}'.format(layer) +
-                                                    '/{z}/{x}/{y}.png')
+            'url': (
+                settings.TILEGARDEN_ROOT +
```
flibbertigibbet: This should go in the string formatter with the other parameters

src/django/pfb_analysis/models.py
```diff
@@ -557,8 +557,11 @@ def destinations_urls(self):
...
+                '/tile/{job_id}/{layer}/'.format(job_id=self.uuid, layer=layer) +
+                '{z}/{x}/{y}.png'
```
flibbertigibbet: Is there a reason this can't go in the string above? Can the brackets not be escaped?

src/tilegarden/src/util/xml-tools.js
```diff
@@ -0,0 +1,25 @@
+/* Helper functions to encapsulate converting to and from XML.
+ *
+ * There's not much here, but it does get reused, and having it centralized should make it
```
flibbertigibbet: :+1:

src/tilegarden/src/util/xml-tools.js
```diff
@@ -0,0 +1,25 @@
...
+ * easier to change libraries or settings if necessary.
+ */
+
+const { promisify } = require('util')
+const xml2js = require('xml2js')
+
+// Make an async-friendly version of the parser
+const parsePromise = promisify(xml2js.parseString)
```
flibbertigibbet: That's neat, I haven't used `promisify` before.

src/django/pfb_analysis/models.py
KlaasH: What about
```python
'url': '{root}/tile/{job_id}/{layer}/{tile_template}'.format(
    root=settings.TILEGARDEN_ROOT,
    job_id=self.uuid,
    layer=layer,
    tile_template='{z}/{x}/{y}.png',
)
```
?

I find the escaped brackets very unsightly
```python
'url': '{root}/tile/{job_id}/{layer}/\{z\}/\{x\}/\{y\}.png'.format(
    root_url=settings.TILEGARDEN_ROOT, job_id=self.uuid, layer=layer)
```

flibbertigibbet: That does look nice

Issue #614, opened 2018-12-05
Support running international jobs


flibbertigibbet:
As a result of #604, not all regions will be in the US and so running analysis jobs will need to support running without US Census data.

Connects #604

PR #615, opened 2018-12-05
Feature/kak/add neighborhood country#604

### LINKED ISSUE #604: Add a country field to Neighborhood model
In order to support the upcoming feature that allows uploads of non-US AnalysisJobs.

Turns out there's a [django-countries](https://github.com/SmileyChris/django-countries/) project, it might be worth a look.

flibbertigibbet:
## Overview

Adds:
 - country field to the `Neighborhood` model and serializers
 - cross-field validation to the `Neighborhood` creation serializer to ensure the US state field is set for the US, and not set outside the US
 - check on `AnalysisJob` creation that job is for the US, to gracefully handle until #614 done
 - filtering by country

### Demo

Two serialized neighborhoods, one in US, one not:

```json
{
    "uuid": "9be694dd-3537-4922-9dcf-74334f57d268",
    "createdAt": "2018-12-05T19:00:02.016253Z",
    "modifiedAt": "2018-12-05T19:00:02.016290Z",
    "name": "boulder",
    "label": "Boulder",
    "organization": "4ff19226-d0a8-4c09-a08f-a6a56ce0e111",
    "country": "US",
    "state_abbrev": "CO",
    "boundary_file": "https://kat-pfb-storage-us-east-1.s3.amazonaws.com/neighborhood_boundaries/root/CO/boulder.zip",
    "visibility": "public",
    "last_job": null
},
{
    "uuid": "7ec58fed-983a-4e2d-8308-55c667c0c491",
    "createdAt": "2018-12-05T19:15:09.388036Z",
    "modifiedAt": "2018-12-05T19:30:43.872267Z",
    "name": "narnia",
    "label": "Narnia",
    "organization": "4ff19226-d0a8-4c09-a08f-a6a56ce0e111",
    "country": "SC",
    "state_abbrev": "",
    "boundary_file": "https://kat-pfb-storage-us-east-1.s3.amazonaws.com/neighborhood_boundaries/root/narnia.zip",
    "visibility": "public",
    "last_job": "71f377ad-7ff7-4a22-a588-1f4f28e21ad3"
}
```

Neighborhood API creation form on the back end. Country in form defaults to US, and fields are ordered to put country and state together:

image

Error on attempting to create a neighborhood in the US without setting the state:

pfb_missing_state_error

Error on attempting to create a neighborhood outside the US, but with a US state set:

pfb_neighborhood_no_state_error

### Notes

Should we have reason to support looking up and storing administrative regions outside the US, we will probably want to switch from using the `localflavor` and `django-countries` apps currently in use for country and state lookup, and instead use [django-cities](https://github.com/coderholic/django-cities). Setting it up appears to be non-trivial, and involves importing data from `geonames`.

The front end should be unaffected by this, as it only displays neighborhoods with completed analysis jobs, and all of those will still be in the US.

## Testing Instructions

 * Run migrations
 * State should be set to US on any existing neighborhoods by the migration
 * Try creating neighborhoods inside and outside the US
 * Neighborhoods should serialize as expected
 * An analysis job created for a non-US place should immediately go to `ERROR` status
 * Filters should work as expected, like http://localhost:9200/api/neighborhoods/?country=US

Closes #604


src/django/pfb_analysis/models.py
```diff
@@ -272,7 +276,7 @@ def _set_geom_from_boundary_file(self, overwrite=False):
                 shutil.rmtree(tmpdir, ignore_errors=True)
 
     class Meta:
-        unique_together = ('name', 'state_abbrev', 'organization',)
+        unique_together = ('name', 'country', 'state_abbrev', 'organization',)
```
KlaasH: I noticed that the `boundary_file` path on the new neighborhoods outside the US looked pretty simple, and I confirmed that if I make two neighborhoods with the same Label in two different countries, it works fine but the boundary file for the second clobbers the first one on S3.

Turns out these same `unique_together` fields are used in [get_neighborhood_file_upload_path](https://github.com/azavea/pfb-network-connectivity/blob/0.7.0/src/django/pfb_analysis/models.py#L43-L48) to generate that path.  So we should add country into the mix.  (And maybe add comments in both places to alert future-us to this nuance.)

flibbertigibbet: I assume we'll need to maintain backwards compatibility with previously uploaded files as well.

flibbertigibbet:
Updated to add country to file path, using three-letter country abbreviation instead of two-letter state abbreviation in the path to support backwards compatibility with exiting uploads without country in the path while avoiding ambiguity in two-letter abbreviations. So, `CA` refers to California:

image

...and `CAN` refers to Canada.

image

src/django/pfb_analysis/models.py
KlaasH: I'm not sure we do, actually.  I think that's run only when the field is saved and the value is stored (in the database, the `boundary_file` field is just a `VARCHAR[1024]`) for use in lookups.

But the `obj.state_abbrev or obj.country.alpha3` way works fine, and will keep the bucket cleaner since changing the path for existing neighborhoods would mean that updating their geometry would leave a stale copy at the old path.

I do think a comment here noting that changing it might affect upload path generation and a note in the `get_neighborhood_file_upload_path` docstring that any terms it includes need to be either unique in themselves (e.g. job_id) or included in the `unique_together` would be helpful for posterity.

flibbertigibbet: Comments updated again.

flibbertigibbet:
Bump for re-review

Issue #625, opened 2018-12-12
Wire up Staging to use Tilegarden deployment


jeancochrane:
Once the Tilegarden staging instance is successfully serving tiles, wire up the app to use it to serve tiles on staging.

Depends on #594, #595, and #624.

jeancochrane:
This will hopefully be as simple as adjusting `settings.TILEGARDEN_ROOT` to use an environment variable pointing to the CloudFront CDN for Tilegarden, but there may be other gotchas hiding in the config.

jeancochrane:
As part of this issue, make sure to document the final deployment process for Tilegarden.

Issue #624, opened 2018-12-11
Modify CSP to allow loading Tilegarden tiles on staging/production


flibbertigibbet:
Add an exception to the nginx Content Security Policy to allow Tilegarden tiles be loaded within the staging and production sites.

Connects #595.

Console error:
```
Refused to load the image '<URL>' because it violates the following Content Security Policy directive: "default-src https: data: blob: 'unsafe-inline' 'unsafe-eval'". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
```

Failed network requests:
![image](https://user-images.githubusercontent.com/960264/49828943-2a665180-fd5b-11e8-8838-67fd8a8830cd.png)

PR #626, opened 2018-12-12
Deploy Tilegarden instance to staging

### LINKED ISSUE #595: Deploy Tilegarden tiler to staging

jeancochrane:
## Overview

Configure Tilegarden for multi-environment deployments, and stand up an instance in staging. In addition, modify CI scripts to include Tilegarden deployments.

Note that this PR doesn't wire up the app with Tilegarden, since that's blocked by #594 and #624. I opened up #625 to keep track of that work.

### Demo

Visit https://2fjf2aov9f.execute-api.us-east-1.amazonaws.com/latest to see a bare instance of Tilegarden:

screen shot 2018-12-12 at 4 54 55 pm

### Notes

* `./scripts/infra` runs the deployment with Claudia based on config files retrieved from S3 remote state, so it should publish deployments for staging and production separately.

## Testing Instructions

You can test the PR by running a manual deployment of Tilegarden to staging. If you'd like to go the extra mile and confirm that your code changes get reflected on staging, edit the Tilegarden source code (perhaps to change the `usage` text that displays on the root endpoint) before running this deployment.

* Shell into the VM with `vagrant ssh`
* Run `./scripts/update` to make sure your images are up to date
* Comment out the following line in `./scripts/infra` (since we don't want to actually run Terraform updates):

https://github.com/azavea/pfb-network-connectivity/blob/92d506ee7e77b0729a9dbafe33afde7b61ec86bc/scripts/infra#L73

* Run `export GIT_COMMIT=latest` (the value can be anything, `./scripts/infra` just expects this variable to be set)
* Run `export PFB_SETTINGS_BUCKET=staging-pfb-config-us-east-1`
* Run `./scripts/infra apply`
* Confirm that the deployment succeeds, and visit the URL that gets printed to the console to hit the staging endpoint

Closes #595.

### DIFF
--- a/scripts/infra
+++ b/scripts/infra
@@ -71,6 +71,17 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                 ;;
             apply)
                 terraform apply "${PFB_SETTINGS_BUCKET}.tfplan"
+
+                popd
+                pushd "${DIR}/.."
+
+                aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/.env" "./src/tilegarden/.env"
+                aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/claudia.json" "./src/tilegarden/claudia.json"
+
+                docker-compose \
+                    -f docker-compose.yml \
+                    -f docker-compose.test.yml \
+                    run --rm --entrypoint yarn tilegarden deploy
                 ;;
             *)
                 echo "ERROR: I don't have support for that Terraform subcommand!"
--- a/src/tilegarden/package.json
+++ b/src/tilegarden/package.json
@@ -32,7 +32,7 @@
     "local": "yarn build-all-xml && node --inspect=0.0.0.0:9229 -- node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module src/api | bunyan -o short",
     "parse-id": "jq -r '.api.id' claudia.json > .api-id",
     "test": "eslint src && jest --coverage",
-    "compile": "yarn build-all-xml && ncc src -o dist && mkdir -p dist/xml && cp src/config/*.xml dist/xml/"
+    "compile": "yarn build-all-xml && mkdir -p dist && cp src/*.js dist && cp -R src/util dist/util && mkdir -p dist/xml && cp src/config/*.xml dist/xml/"
   },
   "devDependencies": {
     "bunyan": "^1.8.12",
@@ -42,7 +42,6 @@
     "eslint-config-airbnb-base": "^12.1.0",
     "eslint-plugin-import": "^2.12.0",
     "jest": "^23.2.0",
-    "@zeit/ncc": "^0.1.16",
     "nodemon": "^1.17.5",
     "rewire": "^4.0.1"
   },
--- a/src/tilegarden/scripts/template-vars.js
+++ b/src/tilegarden/scripts/template-vars.js
@@ -6,8 +6,6 @@
 
 const fs = require('fs')
 
-const envPrefix = process.env.NODE_ENV === 'production' ? 'PROD_' : 'DEV_'
-
 fs.readFile(process.argv[2], 'utf-8', function (err, out) {
     if (err) process.stderr.write(err)
     else {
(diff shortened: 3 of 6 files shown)


scripts/infra
```diff
@@ -71,6 +71,17 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                 ;;
             apply)
                 terraform apply "${PFB_SETTINGS_BUCKET}.tfplan"
+
+                popd
+                pushd "${DIR}/.."
+
+                aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/.env" "./src/tilegarden/.env"
+                aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/claudia.json" "./src/tilegarden/claudia.json"
```
jeancochrane: Claudia creates this `claudia.json` file the first time it deploys a new function. After running that first deployment, I copied the `claudia.json` file into the S3 config bucket so we can use it as remote state.

src/tilegarden/package.json
```diff
@@ -32,7 +32,7 @@
     "local": "yarn build-all-xml && node --inspect=0.0.0.0:9229 -- node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module src/api | bunyan -o short",
     "parse-id": "jq -r '.api.id' claudia.json > .api-id",
     "test": "eslint src && jest --coverage",
-    "compile": "yarn build-all-xml && ncc src -o dist && mkdir -p dist/xml && cp src/config/*.xml dist/xml/"
+    "compile": "yarn build-all-xml && mkdir -p dist && cp src/*.js dist && cp -R src/util dist/util && mkdir -p dist/xml && cp src/config/*.xml dist/xml/"
```
jeancochrane: As @KlaasH and I talked about in our Tilegarden sync, this `ncc` compilation step wasn't really doing anything useful, so I removed it in favor of simply moving the source files to a `dist` directory.

src/tilegarden/scripts/template-vars.js
```diff
@@ -6,8 +6,6 @@
 
 const fs = require('fs')
 
-const envPrefix = process.env.NODE_ENV === 'production' ? 'PROD_' : 'DEV_'
```
jeancochrane: Since we manage all environment variables with an `.env` file, this prefixing is no longer necessary.

jeancochrane:
I noticed today that there were a few problems with the setup that prevented the Lambda functions from A) accessing config files and B) connecting to the database. Fixed those two problems, so you can now see a tile here (misrendered due to #594): https://2fjf2aov9f.execute-api.us-east-1.amazonaws.com/latest/tile/1f350b10-278b-46f4-bd83-617d27eb6da6/ways/16/19085/24804.png

src/tilegarden/package.json
```diff
@@ -32,7 +32,7 @@
     "local": "yarn build-all-xml && node --inspect=0.0.0.0:9229 -- node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module src/api | bunyan -o short",
     "parse-id": "jq -r '.api.id' claudia.json > .api-id",
     "test": "eslint src && jest --coverage",
-    "compile": "yarn build-all-xml && ncc src -o dist && mkdir -p dist/xml && cp src/config/*.xml dist/xml/"
+    "compile": "yarn build-all-xml && mkdir -p dist && cp src/*.js dist && cp -R src/util dist/util && mkdir -p dist/config && cp src/config/*.xml dist/config/"
```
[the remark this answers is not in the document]
jeancochrane: I hear you on the question of whether or not to fudge the compilation step, that's a tough call! It seems reasonable to assume that Tilegarden will want some form of further bundling or minification someday, but at this point it feels like the biggest priority for us is keeping the two implementations as similar as we can. I'll go ahead and fix it to align with how you've set it up on your branch.

If I were going to keep the fake bundling, though, that rsync command is definitely the right way to do it. I tend to forget that rsync has a whole world of flags that can save me a bunch of commands.

jeancochrane:
Thanks for your review @KlaasH! I changed the one line you commented on in order to better match the future of Tilegarden core. Agreed RE: adding documentation, but I'll add that in as part of the follow-up PR since we'll need docs for the additional CloudFront infrastructure. Once tests pass, I'll merge this in!

Issue #616, opened 2018-12-07
django-q container missing from update and build scripts


KlaasH:
I noticed while testing PR #615 that the django-q container was crashing because it couldn't find `django_countries`.  Turns out all it needed was a `docker-compose build`, but that hadn't happened because it's not in the list of containers to build in [`scripts/update`](https://github.com/azavea/pfb-network-connectivity/blob/0.7.0/scripts/update).  It's also missing from `cibuild` and `cipublish`, which seems surprising.

jeancochrane:
I think I've gotten to the bottom of this discrepancy! Looks like `scripts/server` was building the `django-q` image implicitly:

https://github.com/azavea/pfb-network-connectivity/blob/ac0bae71429982291c698e39bdf481f9537c5025/scripts/server#L28

However, [`docker-compose up` will only build images if they don't exist](https://stackoverflow.com/questions/39988844/docker-compose-up-vs-docker-compose-up-build-vs-docker-compose-build-no-cach#answer-39988980). Most likely you saw strange behavior because the code had changed to depend on `django_countries`, but the container didn't update accordingly. Because of issues like this one, Ops tends to prefer `docker-compose --build` in our `server` scripts, since it makes sure that the images are always up do date before serving files.

The `cibuild` and `cipublish` scripts don't build images for `django-q` because the Task Definition for the `django-q` Service uses the `pfb-app` image:

https://github.com/azavea/pfb-network-connectivity/blob/ac0bae71429982291c698e39bdf481f9537c5025/deployment/terraform/app.tf#L183

This makes sense in that the `django-q` service is identical to the app service, it just runs a different command. 

Opened up #623 to make that change to `./scripts/update` that you recommended above, as well as introducing the `--build` flag to `docker-compose up` in `./scripts/server`.

PR #628, opened 2018-12-13
Feature/kak/add analysis upload endpoint#618

### LINKED ISSUE #618: Add POST API endpoint that creates and triggers a new local upload task
POST endpoint should take a Neighborhood UUID and an http(s) url to the zip file of local results. It should trigger a new local upload async task in django-q and return an appropriate http status code to the user to indicate the job has started.

If this task is taken before #617 then the django q task can just be a stubbed job.

Depends on #619

flibbertigibbet:
## Overview

Run a dummy `django-q` task on creating a local analysis upload task in the API.

The endpoint already existed, so this adds some validation and a dummy task.

### Demo

`django-q` processing the upload in the logs:
image

Successful POST response:
image

Attempting to POST with a neighborhood UUID that is a valid UUID, but matches no neighborhood:
image

## Testing Instructions

 * Rebuild containers to address #616, if necessary
 * Post a neighborhood UUID and URL to http://localhost:9200/api/local_upload_tasks/
 * Should create an `AnalysisLocalUploadTask` for the matching neighborhood, with the URL
 * `django_q` should log the (dummy) task processing
 * API response should include expected fields and status
 * Attempting to post invalid data should return an appropriate error

Closes #618

### DIFF
--- a/src/django/pfb_analysis/serializers.py
+++ b/src/django/pfb_analysis/serializers.py
@@ -155,6 +155,12 @@ class AnalysisLocalUploadTaskCreateSerializer(serializers.ModelSerializer):
         validated_data.pop('neighborhood')
         return super(AnalysisLocalUploadTaskCreateSerializer, self).create(validated_data)
 
+    def validate_neighborhood(self, obj):
+        if Neighborhood.objects.filter(uuid=obj).count() != 1:
+            raise serializers.ValidationError(
+                'No matching neighborhood found for UUID {uuid}'.format(uuid=obj))
+        return obj
+
     class Meta:
         model = AnalysisLocalUploadTask
         fields = ('uuid', 'created_at', 'modified_at', 'created_by', 'modified_by',
--- a/src/django/pfb_analysis/tasks.py
+++ b/src/django/pfb_analysis/tasks.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import shutil
 import tempfile
@@ -5,6 +6,8 @@ import tempfile
 from pfb_analysis.models import AnalysisBatch
 from pfb_network_connectivity.utils import download_file
 from users.models import PFBUser
+
+logger = logging.getLogger(__name__)
 
 
 def create_batch_from_remote_shapefile(shapefile_url):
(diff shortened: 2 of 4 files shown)


src/django/pfb_analysis/tasks.py
```diff
@@ -18,3 +18,7 @@ def create_batch_from_remote_shapefile(shapefile_url):
         batch.submit()
     finally:
         shutil.rmtree(tmpdir)
+
+
+def upload_local_analysis(analysis_url):
```
CloudNiner: This should take the `local_upload_task_uuid` since the job will need the entire task model to look up and use the related AnalysisJob and Neighborhood models as well as set Task status, etc.

src/django/pfb_analysis/serializers.py
```diff
@@ -155,6 +155,12 @@ def create(self, validated_data):
         validated_data.pop('neighborhood')
         return super(AnalysisLocalUploadTaskCreateSerializer, self).create(validated_data)
 
+    def validate_neighborhood(self, obj):
+        if Neighborhood.objects.filter(uuid=obj).count() != 1:
```
CloudNiner: 👍

src/django/pfb_analysis/views.py
```diff
@@ -164,12 +164,17 @@ def get_serializer_class(self):
...
+        if not serializer.is_valid():
+            return
         neighborhood_id = serializer.validated_data['neighborhood']
+        url = serializer.validated_data['upload_results_url']
         neighborhood = Neighborhood.objects.get(pk=neighborhood_id)
         user = self.request.user
         job = AnalysisJob.objects.create(neighborhood=neighborhood,
                                          created_by=user, modified_by=user)
-        serializer.save(job=job, created_by=user, modified_by=user)
+        serializer.save(job=job, upload_results_url=url, created_by=user, modified_by=user)
```
CloudNiner: It shouldn't be necessary to add `upload_results_url=url` here. By default, the serializer saves all properties in `validated_data`. The things passed here explicitly are passed because they're required model properties that require a value and aren't already part of `validated_data`.

flibbertigibbet:
To check the passed upload task ID, I also added `django_q` to the logging apps. Console output with the task UUID:
image

The created task (UUIDs match):
image

CloudNiner:
👍

PR #632, opened 2018-12-19
Switch GPG keyservers used in tilemaker docker build

### LINKED ISSUE #630: Investigate GPG keyserver timeouts in Jenkins builds
Every so often, Jenkins fails to build the Django container, throwing an intermittent GPG keyserver error when trying to import remote keys:

```
gpg: requesting key 7E37093B from hkp server keyserver.pgp.com
gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
```

In the short term, the error can be mitigated by rerunning the build.

See:
* http://urbanappsci.internal.azavea.com/job/azavea/job/pfb-network-connectivity/job/PR-629/4/

flibbertigibbet:
## Overview

Address GPG keyserver timeout issues on `tilemaker` container build.

## Testing Instructions

 * Clean out local docker images
 * `./scripts/update` should succeed reliably without GPG keyserver timeouts

Fixes #630.

### DIFF
--- a/src/tilemaker/Dockerfile
+++ b/src/tilemaker/Dockerfile
@@ -22,10 +22,11 @@ RUN set -ex \
     B9AE9905FFD7803F25714661B63B535A4C206CA9 \
     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
     56730D5401028683275BD23C23EFEFE93C4CFFFE \
+    6A010C5166006599AA17F08146C2130DFD2497F5 \
   ; do \
-    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" || \
-    gpg --keyserver pgp.mit.edu --recv-keys "$key" || \
-    gpg --keyserver keyserver.pgp.com --recv-keys "$key" ; \
+    gpg --keyserver hkp://ipv4.pool.sks-keyservers.net --no-tty --recv-keys "$key" || \
+    gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --no-tty --recv-keys "$key" || \
+    gpg --keyserver keyserver.pgp.com --no-tty --recv-keys "$key" ; \
   done
 
 ENV NPM_CONFIG_LOGLEVEL info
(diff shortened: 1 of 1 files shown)


src/tilemaker/Dockerfile
```diff
@@ -22,10 +22,11 @@ RUN set -ex \
...
     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
     56730D5401028683275BD23C23EFEFE93C4CFFFE \
+    6A010C5166006599AA17F08146C2130DFD2497F5 \
   ; do \
-    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" || \
-    gpg --keyserver pgp.mit.edu --recv-keys "$key" || \
-    gpg --keyserver keyserver.pgp.com --recv-keys "$key" ; \
+    gpg --keyserver hkp://ipv4.pool.sks-keyservers.net --no-tty --recv-keys "$key" || \
+    gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --no-tty --recv-keys "$key" || \
+    gpg --keyserver keyserver.pgp.com --no-tty --recv-keys "$key" ; \
```
jeancochrane: Good catch! We fixed a similar pseudo-TTY issue in `docker-django` by [using the `--batch` flag](https://github.com/azavea/docker-django/commit/c5a5f0037ac78403b76144e76874841bb0ff16a8). My understanding is that `--no-tty` should do the same thing.

What's your reasoning behind the hostname and protocol change for the keyserver URIs? Seems like it works fine, but I'm curious what the difference is.

flibbertigibbet: No tty errors surfaced in my testing here, but I thought I'd throw in the `--no-tty` for good measure.

Regarding the host and protocol changes, I'd run into a similar issue getting Postgres keys on a Bionic container [here](https://github.com/flibbertigibbet/django-layermapping-null-demo/blob/master/layermapping/Dockerfile#L20), where the fix was to request ipv4 keys. These specific servers were suggested as a workaround on an issue listing I no longer have a reference to.

flibbertigibbet: [This](https://github.com/docker-library/official-images/issues/4252) isn't what I saw, but seems to be a good discussion of the issue generally.

jeancochrane: Gotcha! That context is helpful. IIRC Taylor wound up using a script to poll a bunch of different keyservers in order to get around a similar issue in `docker-django`: https://github.com/azavea/docker-django/commit/0cb90cfc7b7f0b8e35fcc5ef6d9ea6a974821841 It might make sense to set up something similar here if we continue to see build failures after this comes in, but for now I think your solution is a perfectly fine start.

PR #638, opened 2018-12-27
Fix comparison places not on current page being lost

### LINKED ISSUE #523: Compare feature doesn't retain original selection
After selecting one city (NYC) from the search, I viewed the detail page and chose Compare. 
image

I then chose "Search Places" under "Add a Place to Compare", which takes me back to the search, but doesn't retain my first choice (NYC). 
image

flibbertigibbet:
## Overview

Separately request neighborhood data for comparison places.
Fixes comparison places not in current page of places on list view being lost on reloading list view.

### Notes

Original issue cannot be reproduced locally without having more than one page of analysis results to display.

## Testing Instructions

 * `./scripts/update`
 * Visit front-end site at dynamically loaded port (not nginx) http://localhost:9301
 * Navigate between comparison page, details, and list, adding places to compare
 * Reload list page with places to compare set
 * Comparison place list in list view should be retained as expected

Fixes #523

### DIFF
--- a/src/angularjs/src/app/places/list/places-list.controller.js
+++ b/src/angularjs/src/app/places/list/places-list.controller.js
@@ -99,7 +99,7 @@
 
             ctl.getPlaces = getPlaces;
 
-            getPlaces();
+            getPlaces(true);
         }
 
         /**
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Updated to not re-query on page or filter change.

PR #643, opened 2019-01-07
Configure tile URL based on settings

### LINKED ISSUE #641: Enable toggling tile URL between Tilegarden and S3
On PR #613, I noted that I had just switched the tile URL to Tilegarden rather than adding a toggle because I couldn't remember why I had said, on issue #592, why a toggle would be useful.
Turns out a toggle is useful because we need to be able to deploy the code that imports AnalysisJobs so we can load the geometries before switching over to using them for tiling.

flibbertigibbet:
## Overview

Select either new Tilegarden endpoint URL if `TILEGARDEN_ROOT` has been defined on the environment, or fall back to previous URL structure.

### Notes

We will want to either revert this PR once the intermediate production deployment it's here to support is done, or re-set the production `tilegarden_root` variables in `app.tf` when we're ready to switch production to use Tilegarden.

## Testing Instructions

 * Should still load Tilegarden tiles
 * Remove `TILEGARDEN_ROOT` passed env var from `common.yml`, or set to an empty string
 * Restart `./scripts/server`
 * Tiles should be pulled from S3 (unless the last neighborhood job was a re-import, which won't have any)

Closes #641.


KlaasH:
This makes it toggleable by changing the variable in development but just turns it off in staging/production, right?  How hard would it be to define a new variable so that it can be toggled in the deployed environments by switching a value in `.tfvars` (or, hypothetically, on a running task) rather than requiring a code change?  I'm not sure how many places a new tfvar has to be recorded in, and since the plan is to remove the old tiling setup altogether it's not worth building to much machinery for a toggle, but it would be nice not to have the intermediate deploy/data migration blocking Staging from using Tilegarden.

flibbertigibbet:
I think the commit I just added will leave the Tilegarden URL unset for production, but set for staging via the  `staging-pfb-config-us-east-1.tfvars` override, but it might be good if someone from @azavea/operations could confirm.

hectcastro:
I think that this will lead to `PFB_TILEGARDEN_ROOT` being an empty string when `tilegarden_root` is unset vs. leaving the environment variable itself unset. It looks like that _could_ mess up the conditional logic here:

https://github.com/azavea/pfb-network-connectivity/pull/643/files#diff-0f94d6b7bcd28a2c56e5baa7bcb8bc74R576

By causing it to go down the `if` condition when you'd want it to go down the `else` condition.

flibbertigibbet:
Setting an empty string is what was intended. The second condition will be false for an empty string.

PR #669, opened 2019-01-22
Add analysis import admin UI

### LINKED ISSUE #647: Add “Upload local analysis” view in admin front-end
- Add a new button on the ["Analysis Jobs" page](http://localhost:9301/#/admin/analysis-jobs/)
- Add a view to make a local analysis job.  The form should be just a droplist of Neighborhoods and a field for the zip file URL (extremely similar to the Run Analysis view).

On submission, it should send the request to the `local_upload_task` endpoint and show a Toastr message with the response.

Displaying status/errors of the `AnalysisLocalUploadTask` after it has been successfully created will be the subject of a separate issue.

flibbertigibbet:
## Overview

Add view at /admin/analysis-jobs/import to import
existing analysis results.

### Demo

New button:

image

Page:
image

## Testing Instructions

 * Go to admin UI analysis job page [here](http://localhost:9301/#/admin/analysis-jobs/)
 * Click to import analysis should go to new import page
 * Should be able to start analysis import
 * Should redirect to analysis job list page after submit

Closes #647.

### DIFF
--- a/src/angularjs/src/app/analysis-jobs/import/analysis-jobs-import.controller.js
+++ b/src/angularjs/src/app/analysis-jobs/import/analysis-jobs-import.controller.js
@@ -1,0 +1,45 @@
+/**
+ * @ngdoc controller
+ * @name pfb.analysisJobs.import.controller:AnalysisJobImportController
+ *
+ * @description
+ * Controller for importing analysis results for a job run elsewhere
+ *
+ */
+(function() {
+    'use strict';
+
+    /** @ngInject */
+    function AnalysisJobImportController($log, $state, $filter, toastr, Neighborhood, AnalysisJobImport) {
+        var ctl = this;
+
+        function initialize() {
+            ctl.neighborhoods = Neighborhood.all().$promise.then(function(data) {
+                ctl.neighborhoods = data.results;
+            });
+        }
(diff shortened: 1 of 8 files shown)


src/angularjs/src/app/analysis-jobs/import/analysis-jobs-import.controller.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @description
+ * Controller for importing analysis results for a job run elsewhere
+ *
+ */
+(function() {
+    'use strict';
+
+    /** @ngInject */
+    function AnalysisJobImportController($log, $state, $filter, toastr, Neighborhood, AnalysisJobImport) {
```
KlaasH: A couple unused imports:
```suggestion
    function AnalysisJobImportController($state, toastr, Neighborhood, AnalysisJobImport) {
```

flibbertigibbet:
I went with putting this in a new `import` directory to separate out this view which does not kick of jobs, from the existing two views in the `create` directory, which do. Also I thought the `import` naming made sense and was easier to read than further adding to the `create` views, which I found somewhat hard to read with all the similar `create` naming.

PR #672, opened 2019-01-23
Feature/kak/cancel q tasks#671

### LINKED ISSUE #671: Failed job import tasks retry indefinitely
If an `AnalysisLocalUploadTask` fails, `django-q` continues to log job retry attempts, seemingly indefinitely.

If the task's associated model object has been set to `ERROR` status, the associated django queue task should also be cancelled.

We should also probably consider limiting the number of retry attempts for queued tasks.

flibbertigibbet:
## Overview

Remove failed `django-q` tasks from the task queue, instead of retrying indefinitely. Applies to both analysis import tasks and batch analysis job tasks.

## Demo

A single task failing in the logs:
image

### Notes

Limiting the number of retries to a low number might be preferable to limiting to one attempt, in case of some intermittent failure. However, `django-q` appears to not offer retry attempt count limits as a feature.

Tasks that fail currently have their related Django model object marked with `ERROR` status, but the `django-q` task would continue to retry, even after calling `delete` on the queue task. This change will not stop any existing failing queued tasks from continuing to retry, but will cause new tasks to not retry after failure.

## Testing Instructions

 * Create an analysis upload task that will fail (as with a URL that is not a zip file) from the DRF admin [here](http://localhost:9200/api/local_upload_tasks/)
 * `django-q` should log the task failure, once
 * No retries with the same new queue task name should appear in the logs

Fixes #671

### DIFF
--- a/src/django/pfb_analysis/views.py
+++ b/src/django/pfb_analysis/views.py
@@ -131,7 +131,10 @@ class AnalysisBatchViewSet(ViewSet):
             ClientMethod='get_object',
             Params={'Bucket': settings.AWS_STORAGE_BUCKET_NAME, 'Key': key}
         )
-        async('pfb_analysis.tasks.create_batch_from_remote_shapefile', url, ack_failure=True)
+        async('pfb_analysis.tasks.create_batch_from_remote_shapefile',
+            url,
+            group='create_analysis_batch',
+            ack_failure=True)
 
         return Response({
             'shapefile_url': url,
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Thanks for the DB `delete` command. I'd tried finding a way via the Django shell to cancel zombie tasks, without success. I added group names to the tasks here, so that we might be able to use the group delete and related commands in future, as you noted, which aren't available for tasks without a group.

Issue #634, opened 2018-12-21
Add S3 tile caching setup to terraform config


KlaasH:
See the [issue comment](https://github.com/azavea/tilegarden/issues/126#issuecomment-441776311) and [branch](https://github.com/azavea/tilegarden/tree/feature/kjh/s3-tile-cache).  The idea is to set things up so that
- tile requests come into the CloudFront distribution (as they do already)
- CloudFront sends them to an S3 Website set up to serve a bucket of cached tiles
- The S3 website configuration is set to redirect back to the CloudFront distribution, prepending `latest/` to the request path
- The CloudFront distribution sends all requests that start with `latest/` to the API gateway

The first request for a tile will thus be filled by API Gateway, i.e. by an actual Lambda function spinning up and making the tile. But before returning it, the Lambda function will write the tile to the bucket so later requests for the same tile will find it there.  The code changes are in a separate issue: #___.

So the task is to set up the Terraform config to provision CloudFront and S3.  Until the code side is done, all requests will fall back to the API gateway (though we could manually add tiles to the bucket to test the cache response).

jeancochrane:
Currently blocked waiting on #680.

PR #673, opened 2019-01-23
Write tiles to S3 cache

### LINKED ISSUE #635: Write tiles to S3 cache bucket
This is the code side of adding tile caching.  #634 is the infrastructure side.  See also the [issue comment](https://github.com/azavea/tilegarden/issues/126#issuecomment-441776311) and [branch](https://github.com/azavea/tilegarden/tree/feature/kjh/s3-tile-cache) on the main Tilegarden project.

The task is to add code to `api.js` so that, if there is a variable configured in the environment with the cache bucket name, it writes all tiles to that bucket, using the request path as the key

This depends on #634 insofar as it needs to use the cache bucket configured there, though it could be done in parallel by manually creating a bucket and supplying a variable with its name.

Note that I'm not sure it's possible to actually set up a cached endpoint in development (maybe related to #593), but it should at least be possible to have a local Tilegarden instance write tiles to S3.

### LINKED ISSUE #636: Add script to clear S3 tile cache
Once we have tiles cached in S3 (see #634 and #635), we'll need an easy way to clear them if we need to, i.e. if we're changing the styles.  Even if this is a simple aws-cli command, it would be nice to encapsulate it in a way that makes it easy to discover and run.

KlaasH:
## Overview

Adds a step in the tile generation process where it checks for a `PFB_TILEGARDEN_CACHE_BUCKET` variable in the environment and, if it's set, writes the tile into that bucket, using the request path as the key.

Currently there's no way to serve the tiles from cache.  #634 covers that part.

Also adds a script to clear all files in the configured cache bucket.  I'm not sure that will be the way to do it in deployed environments (where it's harder to get a shell, and where the number of items might make the "change settings to make all objects expire, change settings back" deletion strategy preferable), but it seems useful in development.

### Demo
```
vagrant@pfb-network-connectivity:/vagrant$ aws s3 ls --recursive s3://dev-pfb-tilecache-us-east-1/
2019-01-23 17:00:57        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2103/2894.png
2019-01-23 17:00:56        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2103/2895.png
2019-01-23 17:00:56        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2103/2896.png
2019-01-23 17:00:56        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2103/2897.png
2019-01-23 17:00:57        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2103/2898.png
2019-01-23 17:00:56        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2104/2898.png
2019-01-23 17:00:57        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2105/2898.png
2019-01-23 17:00:57       3395 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2106/2898.png
2019-01-23 17:00:56       4447 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2107/2898.png
2019-01-23 17:00:56        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2108/2898.png
2019-01-23 17:00:57        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/census_blocks/13/2109/2898.png
2019-01-23 17:52:21        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2103/2894.png
2019-01-23 17:52:20        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2103/2895.png
2019-01-23 17:52:20        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2103/2896.png
2019-01-23 17:52:21        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2103/2897.png
2019-01-23 17:52:21        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2103/2898.png
2019-01-23 17:52:19        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2104/2898.png
2019-01-23 17:52:20        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2105/2898.png
2019-01-23 17:52:19       4204 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2106/2898.png
2019-01-23 17:52:19       7566 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2107/2898.png
2019-01-23 17:52:19      16902 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2108/2898.png
2019-01-23 17:52:21        334 tile/80bbb2a8-5399-40f2-95a8-537917f4197c/ways/13/2109/2898.png
```

### Notes

- Since I'm expecting serving from cache to be set up as an S3 Website, the key has to match the full request path exactly.  So the tiles have to go in the root of the bucket, not a subdirectory.  Or possibly there's a way to configure an S3 website to strip off a prefix, and it could be more flexible.  But I didn't manage to find an option like that in my limited looking.
- I don't think it's going to be possible to test the full caching setup locally, since Cloudfront can't use your local machine as an origin.  See 593.
- This also switches to using the `AWS_PROXY` request format for Claudia API Builder. See [docs](https://github.com/claudiajs/claudia-api-builder/blob/v3.0.2/docs/request-object.md#the-api-builder-request-object).
- Unfortunately `claudia-local-api` doesn't fully reproduce the AWS Proxy-style request object (see https://github.com/azavea/claudia-local-api/issues/1).  As noted in the code, this works around that by reconstructing the path from the parameters.

## Testing Instructions

- In the VM, run `./scripts/update` and `./scripts/server`
- Go to a neighborhood in [your dev front-end](http://localhost:9301/#/places////). Make sure the tiles are coming from Tilegarden, not S3 (i.e. that your TILEGARDEN_ROOT variable is set for the container).
- Any tiles you've just generated should now be in the `dev-pfb-tilecache-us-east-1` bucket (you can run `./scripts/clear-tile-cache` to clear out the clutter if it's hard to tell).

Resolves #635 
Resolves #636

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -18,6 +18,7 @@ services:
       - PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION
       - PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME
       - PFB_TILEGARDEN_ROOT=http://localhost:9400
+      - PFB_TILEGARDEN_CACHE_BUCKET=dev-pfb-tilecache-us-east-1
     volumes:
       - $HOME/.aws:/root/.aws:ro
 
--- a/scripts/clear-tile-cache
+++ b/scripts/clear-tile-cache
@@ -2,14 +2,16 @@
 
 set -e
 
-BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
+DEFAULT_BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
 
 function usage() {
     echo -n \
-         "Usage: $(basename "$0") [--list]
+         "Usage: $(basename "$0") [--list] [BUCKET]
 
-Removes all cached tiles (that is, all files) from the configured S3 bucket:
-    $BUCKET_NAME
+Removes all cached tiles (that is, all files) from the configured S3 cache bucket:
+    $DEFAULT_BUCKET_NAME
+
+The target bucket can be overridden by passing a different bucket name as an argument.
 
 With the --list option, lists and summarizes the contents before asking for confirmation.
 "
@@ -20,7 +22,16 @@ then
     if [ "${1:-}" = "--help" ]; then
         usage
         exit
-    elif [ "${1:-}" = "--list" ]; then
+    fi
+
+    if [ "${1:-}" = "--list" ]; then
+        SHOW_LIST=true
+        shift
+    fi
+
+    BUCKET_NAME="${1:-$DEFAULT_BUCKET_NAME}"
+
+    if [ $SHOW_LIST ]; then
         aws s3 ls --recursive --summarize "s3://${BUCKET_NAME}"
         echo ""
     fi
--- a/scripts/clear-tile-cache
+++ b/scripts/clear-tile-cache
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -e
+
+BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
+
+function usage() {
+    echo -n \
+         "Usage: $(basename "$0") [--list]
+
+Removes all cached tiles (that is, all files) from the configured S3 bucket:
+    $BUCKET_NAME
+
+With the --list option, lists and summarizes the contents before asking for confirmation.
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+        exit
+    elif [ "${1:-}" = "--list" ]; then
+        aws s3 ls --recursive --summarize "s3://${BUCKET_NAME}"
+        echo ""
+    fi
+
+    echo "This will delete all files from the '$BUCKET_NAME' S3 bucket."
+
+    read -r -p "Are you sure? [y/N] " response
+    response=${response,,}    # tolower
+    if [[ "$response" =~ ^(yes|y)$ ]]
+    then
+        aws s3 rm --recursive "s3://${BUCKET_NAME}"
+    fi
+fi
--- a/src/tilegarden/.env.dev
+++ b/src/tilegarden/.env.dev
@@ -1,5 +1,7 @@
+AWS_PROFILE
 PFB_DB_HOST=database.service.pfb.internal
 PFB_DB_DATABASE=pfb
 PFB_DB_PASSWORD=pfb
 PFB_DB_PORT=5432
 PFB_DB_USER=pfb
+PFB_TILEGARDEN_CACHE_BUCKET=dev-pfb-tilecache-us-east-1
--- a/src/tilegarden/.eslintrc.json
+++ b/src/tilegarden/.eslintrc.json
@@ -4,15 +4,17 @@
   },
   "extends": "airbnb-base",
   "rules": {
-    "semi": [2, "never"],
-    "no-unexpected-multiline": 2,
-    "indent": ["error", 4],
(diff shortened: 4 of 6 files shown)


scripts/clear-tile-cache
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -e
+
+BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
```
jeancochrane: Do you think it's worth allowing the user to optionally set `BUCKET_NAME` if they want to clear a bucket other than the one that's configured for dev? e.g.:

```bash
if [[ -z "${BUCKET_NAME}" ]]; then
    BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
fi
```

I agree with what you wrote in the PR body, that it seems like the usefulness of this script is probably limited to the isolated case where you want to clear the dev cache for testing -- still, I wonder if more extensibility will come in handy. I leave the call up to you.

jeancochrane: Do you think it's worth allowing the user to set this bucket themselves, without having to edit `common.yml`? e.g.:

```bash
if [[ -z "${BUCKET_NAME}" ]]; then
    BUCKET_NAME=$(docker-compose run --rm tilegarden bash -c 'echo -n $PFB_TILEGARDEN_CACHE_BUCKET')
fi
```

I agree with your comment in the PR body that it's hard to think of cases where this script would be useful other than clearing the preconfigured dev tile cache, so I leave the decision up to you.

KlaasH:
Whoops, I guess I rushed this out the door a bit.  Will clean up linting and tests.

common.yml
```diff
@@ -18,6 +18,7 @@ services:
       - PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME_REVISION
       - PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME
       - PFB_TILEGARDEN_ROOT=http://localhost:9400
+      - PFB_TILEGARDEN_CACHE_BUCKET=dev-pfb-tilecache-us-east-1
```
flibbertigibbet: Why not put the tiles for development users somewhere in the existing `${DEV_USER}-pfb-storage-us-east-1` bucket we're using for other S3 storage?

scripts/clear-tile-cache
KlaasH: Yeah, on reflection that's probably worth having.  It might not be useful, but then again it might, in which case it's a lot easier to write it in now than muddle through without it later.

common.yml
KlaasH: I'm expecting that for serving tiles from the cache we'll want to use an [S3 Website](https://docs.aws.amazon.com/AmazonS3/latest/dev/HowDoIWebsiteConfiguration.html), since that's what I did in [the Tilegarden branch](https://github.com/azavea/tilegarden/tree/feature/kjh/s3-tile-cache) where I've been working on caching.  And as far as I can tell that serves the whole bucket.  If there turns out to be a way to restrict it and use the existing bucket, that would be worth pursuing.

It's true that a shared dev cache could cause some inconvenience, though given that tiles are namespaced by job ID, it would be more in the way of clutter than any weird tiler behavior.  I'm inclined to run with it, though--with the fallback of making a personal bucket and overriding the variable if one is needed--until we know whether it's a problem in practice.

KlaasH:
Linter appeased, bucket argument added to `clear-tile-cache`.  Will merge shortly, if the build succeeds.

PR #677, opened 2019-01-24
Feature/kak/job upload details#648

### LINKED ISSUE #648: Add upload details to job detail view for uploaded jobs
We need to expose some details of the local analysis upload process to the user.  In particular, some errors (like invalid zip file) that we want the user to see get written onto the `AnalysisLocalUploadTask`.

Rather than have a separate view or list for keeping track of `AnalysisLocalUploadTask` instances and exposing those errors, since [the job is created during the synchronous part of the creation request](https://github.com/azavea/pfb-network-connectivity/blob/e1cbd7d077527187fffe511b8ee9cc3c0b28e0e3/src/django/pfb_analysis/views.py#L172-L173), we should be able to show upload information on the job detail page.

The upload details could be added into the `analysis_job` response or the page could do a separate query to the `local_upload_tasks` endpoint.  Probably the latter is better, to avoid cluttering up the `analysis_job`, but either way seems potentially fine.

We might want to add a status for `AnalysisJob` to make the process more visible (e.g. `IMPORTING`), though that's not strictly necessary as long as we make sure it ends up in either `COMPLETE` or `ERROR`.

flibbertigibbet:
## Overview

Display analysis import task status for its related analysis job in the web admin UI page list view. Also display additional analysis import task information on the detail page for its related analysis job.

Adds some import task fields to the related analysis job serializer. Also adds the ability to filter the import task endpoint directly by analysis job UUID, which is currently unused by the front-end site.

### Demo

Analysis job list view, with one job created via import task:
image

The analysis job detail section for its import task:
image

Serialized analysis job response with related import task:
image

### Notes

I went with serializing a few extra fields on the analysis job rather than adding separate front-end queries for the related import task for each job, as doing so doesn't add much data to the existing analysis jobs query, should be more efficient network use than multiple additional queries, and more straightforward to implement than chained queries. Related batches, also shown in the analysis job list view and details, are also serialized directly with the analysis jobs.

## Testing Instructions

 * Create a new analysis import task, which should redirect on success to the analysis job list
 * Status for the task should show with its related analysis job in the list view
 * The related analysis job detail view should should the import task details as expected

Close #648


src/django/pfb_analysis/serializers.py
```diff
@@ -98,6 +98,37 @@ class Meta:
         read_only_fields = fields
 
 
+class AnalysisLocalUploadTaskSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = AnalysisLocalUploadTask
+        fields = ('uuid', 'error', 'job', 'status', 'upload_results_url',)
```
KlaasH: Since these show up in this order on the detail page, putting 'status' before 'error' makes it a bit more readable--it's natural to look below the ERROR status for the message, and when the status isn't ERROR having it first avoids having to look below to figure out why the "local upload task error" field is blank.

Also, having 'job' included in the embedded serialization and putting it on the detail page as "local upload task job" is a bit odd.  It took me a moment to realize that's not some other related entity, just a repetition of the ID of the job whose detail page I'm looking at.  I guess that's related to
> Also adds the ability to filter the import task endpoint directly by analysis job UUID, which is currently unused by the front-end site.

How hard would it be to suppress that field in the embedded serialization?

src/angularjs/src/app/analysis-jobs/list/analysis-jobs-list.html
```diff
@@ -24,6 +24,7 @@ <h1 class="no-margin-top">Analysis Jobs</h1>
             <td>{{analysisJob.status | displayStatus}}</td>
             <td>{{analysisJob.neighborhood.label}}</td>
             <td>{{analysisJob.batch}}</td>
+            <td>{{analysisJob.local_upload_task.status | displayStatus}}</td>
```
KlaasH: Just a note: this sort of works by coincidence--the `AnalysisUploadTask` statuses are a subset of the `AnalysisJob` statuses, but they're parallel constructs that could have been different.  If we wanted to add a status for the uploads that didn't match one already defined on jobs, we'd have to decide whether to make the filter a superset or split it.  But fortunately we don't.

KlaasH: So indeed.  The capital letters drew my eye and I missed the comment.  Excellent.

src/django/pfb_analysis/serializers.py
flibbertigibbet: I missed the `job` was coming through nested, good catch. I'll split out the serializer into two versions.

flibbertigibbet:
This should be ready for another look.

Serialization of a job:
image

Job status updates on the job were supposed to be in sync with the import task in case of error, but I suppose that was lost at some point. I fixed that here.

Issue #589, opened 2018-11-01
Review and merge pedestrian paths PR


CloudNiner:
#586 

Blocked pending additional changes from @beckymasond 

We should:
- [ ] ensure no regressions on performance of jobs
- [ ] ensure no issues with job outputs

jcahail:
Per email from @beckymasond, no changes will be made at this time.

KlaasH:
No changes means this is is unblocked and we can review and merge as-is (assuming no issues arise in regression testing).

flibbertigibbet:
I ran a local Boulder analysis, then imported the most recent public production Boulder analysis from [here](https://bna.peopleforbikes.org/#/admin/analysis-jobs/eaf9fff0-951f-4a96-be68-b88b0e503807/). The results are very different, though; the population counts I wouldn't expect to change, but they're way off from each other.

![image](https://user-images.githubusercontent.com/960264/52005426-a6496400-2497-11e9-8617-be1e1caa0c90.png)

flibbertigibbet:
I re-ran analysis for Boulder locally on the `develop` branch, and got results that look much closer; the populations now match.

(left, pedestrian PR branch build; middle, develop branch; right, re-imported production results)
![image](https://user-images.githubusercontent.com/960264/52013214-75bef580-24aa-11e9-97fa-9c7afc081031.png)

flibbertigibbet:
Full summary results comparison:
![image](https://user-images.githubusercontent.com/960264/52015039-0d264780-24af-11e9-8a1e-aabc4c0edc94.png)

flibbertigibbet:
Merged #682

PR #682, opened 2019-01-29
Rebased: removing pedestrian paths and fixing one-way evaluation

flibbertigibbet:
Same as #586, rebased against develop.


flibbertigibbet:
See #589 for screenshots. This is ready to merge.

Issue #685, opened 2019-01-29
Cannot provision new VM


flibbertigibbet:
Attempting to provision a fresh VM fails on attempting to install `docker-compose`.

<Details>
TASK [azavea.docker : Configure Docker] ****************************************
changed: [default] => {"changed": true, "checksum": "89dee2cdbad5d5190b31432a4fa095ea428a2bfe", "dest": "/etc/default/docker", "gid": 0, "group": "root", "md5sum": "da9a056689d42b30025051b81962ecc7", "mode": "0644", "owner": "root", "size": 651, "src": "/home/vagrant/.ansible/tmp/ansible-tmp-1548786324.06-135183084308066/source", "state": "file", "uid": 0}

TASK [pfb.docker : Install docker-compose] *************************************
[...]

PR #690, opened 2019-01-30
Expanding bikeable service roads to include 'bicycle'='yes'

beckymasond:
## Overview

Currently the BNA only considers roads with the functional class "service" to be bikeable roads if they include the tag bicycle='designated'. We'd like to include the value 'yes' as an acceptable alternative to 'designated'. 

### Notes

I have not tested this change because I am running into the same issues running the BNA locally as noted in #685.


KlaasH:
I'm going to close this and make an issue for reopening and evaluating it.  Until we get our issues with concurrent CI builds sorted out, having it open will keep causing other builds to fail.

Issue #694, opened 2019-01-31
Make tile strategy fully toggleable


KlaasH:
After some post-checkin discussion yesterday and in light of #693, rather than try to provisionally deploy Tilegarden then rip the old tiling machinery out and deploy again, I think it makes sense to go further down the road of being able to toggle between the options.  That is, set things up so that the new or old tiling strategies both work and either can be activated by setting or unsetting a toggle variable.

This seems better than having bits of dead/irrelevant code from the old tile setup hanging around, and it provides more flexibility/requires less coordination around deployment than removing the old strategy immediately--we can deploy Tilegarden first then clear out the old setup at our leisure, or if it seems preferable we'd have the option of deploying the rest of the new work without activating Tilegarden, then switching over later.

This invalidates #650 and makes #611 a post-deployment task.

Issue #611, opened 2018-11-30
Remove old tiling container, code, and config


KlaasH:
Including:
- The `tilemaker` container and source tree
- The related code and settings in the Django app
- Related infrastructure configuration (e.g. the `tilemaker` SQS queues and job definitions)

Requires #650

KlaasH:
Also need to make sure the analysis job ends with `COMPLETE` status, since it currently puts the job in `EXPORTED` status when it's done and it's the `tilemaker` container that [marks it Complete](https://github.com/azavea/pfb-network-connectivity/blob/d6ffcd6cc1b186171981852a7a62e68c4e9f82c0/src/tilemaker/entrypoint.sh#L26).

flibbertigibbet:
Connects #650

flibbertigibbet:
Also remove the tiler toggle logic from #650.

PR #696, opened 2019-01-31
Expand implementation of tilegarden toggle

KlaasH:
## Overview

As noted on #694, there are advantages to being able to easily switch between the tilemaker/S3 tiler and Tilegarden.  Also, per https://github.com/azavea/pfb-network-connectivity/issues/611#issuecomment-454157314, there was a bit more to do to fully integrate Tilegarden.

This switches the toggling behavior to use a new variable, `PFB_USE_TILEGARDEN`, and changes the analysis model to check it to determine both what tile URLs to send to the front-end and what to do after its analysis job runs--generate tiles with `tilemaker` if Tilegarden is disabled, mark the job complete if it's enabled.

This should make it easy to stay with `tilemaker` (by setting the variable to `false`), use Tilegarden immediately (by setting it to `true`), or start with the former then switch to the latter.  The geometry import at the end of the analysis is not conditional, so even if `USE_TILEGARDEN` is false, any new analysis runs will still have their geometries imported, so nothing extra needs to be done to use Tilegarden for their tiles.
The one scenario that this doesn't fully enable is activating Tilegarden, running new jobs, then reverting to the S3 tile source.  The tiles for jobs run with `USE_TILEGARDEN` set to true will not have been generated.

### Notes

- The ugliest part of this (in my opinion, anyway) is the special case in `AnalysisJob.update_status`.  If we were intending to keep the `tilemaker` option available forever, it would probably be worth doing the work of passing the information into the analysis container that there's not going to be a tiling job so it should end with COMPLETE rather than EXPORTED.  But as things stand now, I don't think it's worth the trouble.

## Testing Instructions

- With `PFB_USE_TILEGARDEN=true` in `common.yml` (which this sets as the default):
   - The tile URLs for your places should point to the configured TILEGARDEN_ROOT
   - If you [kick off an analysis](http://localhost:9301/#/admin/analysis-jobs/create/), the log output will show the "Can't actually run development analysis jobs on AWS. Try this:" message with the command to run locally, but won't include the "Can't actually run development tiling jobs on AWS. Try this:" message.
   - If you run the analysis command, it will get all the way to COMPLETE and show up on your site, with working tiles.
- If you change `PFB_USE_TILEGARDEN=false` in `common.yml` and restart your instance:
   - The tile URLs for your places should point to S3
   - If you [kick off an analysis](http://localhost:9301/#/admin/analysis-jobs/create/), the log output *will* include the "Can't actually run development tiling jobs on AWS. Try this:" message and tilemaker command.
   - If you run the analysis command, it will succeed and leave the job in EXPORTED status, so it won't show up on your site.  If you then run the `tilemaker` command, it will set the job to COMPLETE when it's done, at which point it should show up on your site, with tiles coming from S3.

Connects #694.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -19,6 +19,7 @@ services:
       - PFB_AWS_BATCH_TILEMAKER_JOB_DEFINITION_NAME
       - PFB_TILEGARDEN_ROOT=http://localhost:9400
       - PFB_TILEGARDEN_CACHE_BUCKET=dev-pfb-tilecache-us-east-1
+      - PFB_USE_TILEGARDEN=true
     volumes:
       - $HOME/.aws:/root/.aws:ro
 
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -152,8 +152,9 @@ data "template_file" "pfb_app_https_ecs_task" {
     batch_analysis_job_definition_name_revision  = "${var.batch_analysis_job_definition_name_revision}"
     batch_tilemaker_job_queue_name               = "${var.batch_tilemaker_job_queue_name}"
     batch_tilemaker_job_definition_name_revision = "${var.batch_tilemaker_job_definition_name_revision}"
-    tilegarden_root                              = "${var.tilegarden_root}"
+    tilegarden_root                              = "https://${aws_route53_record.tilegarden.fqdn}"
     tilegarden_cache_bucket                      = "${lower(var.environment)}-pfb-tile-cache-${var.aws_region}}"
+    use_tilegarden                               = "${var.use_tilegarden}"
   }
 }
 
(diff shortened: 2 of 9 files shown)


KlaasH:
> On setting `PFB_USE_TILEGARDEN` to `false` in `common.yml` and restarting, the tile requests still go to S3 (and get 403s).

With `PFB_USE_TILEGARDEN=false`, the tile requests should go to S3.  If the jobs you're looking at are new or you didn't run the old tile generation for them (which has to be run manually for development jobs, like the analysis), the tiles won't be there.

KlaasH:
> might there be caching on staging/production that could cause issues with toggling the tiler reference? (Would a user with a stale browser cache get broken tiles after a tiler switch?)

That could be a concern if the URL pattern were coded into the front-end and the actual URLs were composed based on the job ID or something, but in this case the tile URLs come with the analysis job details from the API.  The [analysis jobs service](https://github.com/azavea/pfb-network-connectivity/blob/0.8.1/src/angularjs/src/app/components/analysis-jobs/analysis-jobs.service.js#L19-L22) doesn't have `cache: true` set, and I'm not seeing any caching when I try it, so I don't think this will be an issue.

PR #697, opened 2019-01-31
Feature/kak/pass city fips#608

### LINKED ISSUE #608: Store and pass city FIPS for neighborhood
Store added `PFB_CITY_FIPS` value added in #585 on the `Neighborhood` model, and pass it to analysis runs via environment variable, as the state FIPS is passed [here](https://github.com/azavea/pfb-network-connectivity/blob/0.7.0/src/django/pfb_analysis/models.py#L651).

Depends on #585.

flibbertigibbet:
## Overview

Add the city FIPS to the neighborhood model, and pass and use it through job processing and in UI.

### Demo

The web UI neighborhood list page, now with added FIPS codes:
pfb_neighborhood_list_with_fips

The web UI neighborhood creation page, with front-end validation for the FIPS:
pfb_city_fips_form_error

Some custom errors the API will return on checking the FIPS:
pfb_city_fips_start_state_error

pfb_city_fips_error_numeric

Development log output after kicking off a batch job. FIPS included in the command:
pfb_django_batch_logs

During analysis run for one of the jobs, the default city speed was found (meaning FIPS passed correctly):
pfb_boulder_found_fips_speed

### Notes

I wrote a script to look up city FIPS from [this Census list of FIPS for place names](https://www2.census.gov/geo/docs/reference/codes/files/national_places.txt) and add them to a given batch analysis shapefile, and ran it on the 100 cities batch file from S3. [The result is on S3 now](https://s3.amazonaws.com/pfb-batch-run-inputs-us-east-1/PFB_100_cities_cityfips.zip). I also made a test file with three of the 100 cities (Boulder, Sioux Falls, and New Haven, picked because they're small-ish.) [Test batch shapefile here](https://s3.amazonaws.com/pfb-batch-run-inputs-us-east-1/PFB_test_cities_cityfips.zip).

We should probably run the script on the other batch shapefiles in that bucket, if the one I've made already looks good.

## Testing Instructions

 * run migrations
 * should be able to create neighborhood in web UI or DRF UI, with or without city FIPS
 * should get appropriate error message for a bad FIPS
 * running analysis individually or in batch should use and pass the city FIPS, if present
 * logged analysis command should include FIPS, if present on the neighborhood
 * city speed should be found by city FIPS, if defined in `city_fips_speed.csv`

Closes #608


flibbertigibbet:
Management command added:

pfb_add_fips_cmd

src/django/pfb_analysis/management/commands/add_city_fips.py
```diff
@@ -0,0 +1,145 @@
...
+            place_count = len(shp)
+            found = 0
+            for feature in shp:
+                prop = feature.get('properties')
+                city = prop['city'].upper()
+                state_places = lookup.get(prop['state']).keys()
+                found_fips = None
+                if state_places:
+                    for place in state_places:
+                        if place.find(city) > -1:
```
KlaasH:
```suggestion
                        if place.find(city) == 0:
```
The file I was trying had Sioux City in it, which matches both "Sioux City" and "South Sioux City" with this check.  Matching only at the beginning resolves that one, though it wouldn't help with "Grand Rapids" vs. "Grand Rapids charter township".  Not that this has to be perfect, but we might as well get as close as we conveniently can.

flibbertigibbet:
Ready for another look

src/django/pfb_analysis/management/commands/add_city_fips.py
```diff
@@ -0,0 +1,148 @@
...
+                city = prop['city'].upper()
+                state_places = lookup.get(prop['state']).keys()
+                found_fips = None
+                if state_places:
+                    for place in state_places:
+                        if place.find(city) == 0:
+                            fips = lookup[prop['state']][place]
+                            logger.info("""Found place {place} for {city} in {state}
+                                with FIPS {fips}""".format(place=place, city=city,
+                                                           state=prop['state'], fips=fips))
```
KlaasH: Thanks!  Showing both the `place` and the `city` should help with catching fuzzy matching errors, too.
NBD, but if this were
```
logger.info("Found place {place} for {city} in {state} "
            "with FIPS {fips}".format(
                place=place, city=city, state=prop['state'], fips=fips))
```
the line break wouldn't come through in the log message.

PR #707, opened 2019-02-11
Tilegarden performance improvement

### LINKED ISSUE #693: Profile and tune Tilegarden tile generation
Using Tilegarden on staging, the tile generation requests are taking longer than I expected.  Though the local and remote environments are somewhat different, I would expect the performance to be a bit more similar, and closer to what we saw on the Tilegarden demo.

The tile generation tasks don't use much memory, but turning up the memory allocated to the Lambda function also increases the CPU allocation, which does speed them up at least somewhat.  But it looks like the effect isn't linear for higher settings, so it seems like there might be another bottleneck.

So the first task is to add logging to the tile generation code path to get a better sense of what's taking the most time.  Then the next task is to try to improve whatever seems like the most important target and/or the lowest-hanging fruit.

KlaasH:
## Overview

Improves Tilegarden performance by applying some optimization parameters (from [the Mapnik documentation](https://github.com/mapnik/mapnik/wiki/OptimizeRenderingWithPostGIS)).

Since there's no way to pass those parameters through the `carto` compilation from mml/mss to Mapnik XML, this also changes the build and deploy commands to remove that step, so that now the XML is canonical (and no longer in .gitignore) and the mml/mss files and the script to compile them into Mapnik XML are there for developer convenience.

Another side-effect of this change is that the compilation step that pulled the environment variables into the map config was cut out.  I moved the template strings into the Mapnik XML and the substitution step into the Lambda function itself.  This has the benefit of making the map config environment-independent (formerly the mml/mss files were but the compiled XML was not).

To facilitate understanding where the time is going, this also adds a logging package and a bunch of debug logging statements to the function.  The only non-debug logging statement remains the "Uploaded to S3" one, but it gets pretty verbose if you set `DEBUG` in the environment.  It turns out that the time is mostly going to `render`, which isn't surprising, but it's nice to be sure...

I did some performance comparisons at different memory allocations.  Though the sample size is too small to put too much trust in the numbers, the results I got were that 256MB ran about 37% faster than 128MB, 512MB ran another 30% faster, and 1024MB ran 19% faster than that.  So 512MB seems like a reasonable choice to get good performance without too much diminishing returns.  The functions don't actually use anywhere close to that amount of memory, but the higher CPU and/or the higher network throughput allocated along with the higher memory clearly makes a difference.

### Demo

image

### Notes

- Possibly the mml/mss files and compilation script should be removed altogether, but it really is nicer to write styles in that format when experimenting with possibilities.  So I think the option to compile to Mapnik XML then just do a polish and optimization pass on the XML is worth keeping.
- Since the database parameters are being drawn from the environment now, this also adds them to the Lambda environment by adding them to the `claudia` commands.
- Since those `claudia` commands were already very unwieldy, inlined in the `package.json` `scripts` section, and this would have made them even worse, I moved them into tiny bash scripts in the Tilegarden scripts directory.
- I did some comparison testing to figure out which parameters and choices actually made for the fastest tile generation.  
   - `extent_from_subquery` was by far the most important.
   - Specifying the srid of the source table didn't make a measurable difference, but it can't hurt.
   - I didn't expect `persist_connection` to make a difference one way or the other, and that seemed to be the case.  I added it because it shouldn't do any harm and might help prevent database connection saturation from becoming a problem.
   - I initially thought that defining one `neighborhood_ways` layer and putting both styles on it would be faster than defining two different layers that pull the same data.  That turned out to be false--not by much, but the 2-layer setup was consistently slightly faster.
   - I also thought setting `cursor_size` to something large enough to capture all the features in a tile would help by reducing network round trips.  This turned out to be false--it actually made things slightly slower.
- To facilitate that comparison testing, I rehabilitated the Tilegarden feature whereby it can read the config file from an S3 bucket if given the bucket name as a query parameter.  Most of the feature was still working, I just had to restore the bit where it actually reads in the bucket name.  With that in place, I could try multiple configurations without having to switch anything but the URL.  To avoid the cache and reduce variability from network delay, I tested by making requests from the bastion directly to the API Gateway endpoint (the bad encoding doesn't matter when all you care about is timing).

## Testing Instructions

- Tile generation should work as before, both locally and on staging (where these changes will show up if `test/kjh/tilegarden-performance` was built since the last `develop` build)
- It should be significantly faster (except for cold starts)

Resolves #693.


KlaasH:
> Are these tiler logs going to a logging service in staging or production? If so, we should be sure debug logging is turned off in those environments, so as to not overwhelm the logs.

The Tilegarden logs are going to CloudWatch.  I confirmed that the function pushed by the test deployment is only logging the one `info` message.  There's no provision for passing `DEBUG` into the environment on deployment, so the only way to activate it is to set it manually on the deployed Lambda function (i.e. in the AWS console. Or I assume with the CLI, though I haven't tried).

PR #708, opened 2019-02-11
Update census tile color ramp to deciles

### LINKED ISSUE #700: Change census blocks map color ramp to 10 equal segments
The census blocks style already uses 10 steps, but with nine 6% bands and one 46% band. Change it so that all the intervals are 10%.

- [ ] [Tilegarden style](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/tilegarden/src/config/census_blocks.mss)
- [ ] [Map legend](https://github.com/azavea/pfb-network-connectivity/blob/0.8.1/src/angularjs/src/app/components/map/map.constants.js#L36-L37)

KlaasH:
## Overview

Sets the color ramp for census block connectivity to 10 equal segments (rather than nine 6%-wide segments and one 46% top segment).

### Demo

Before:
image

After:
image

## Notes

- This PR is pointed at the branch for #707, since that changes the style setup.  I'll rebase after that's merged.
- I updated `census_blocks.mss` as well. I'm still of two minds about whether to keep that, but while it's here it seems better to have it in sync than not.

## Testing Instructions

Rebuild and restart your Tilegarden container, then load a neighborhood and view the "Census blocks with access" layer.  Everything should look a bit more dire.

Resolves #700.


KlaasH:
> The label ranges shouldn't overlap (i.e., should be 80-89, not 80-90).

Good point.  Fixed.  I also rebased onto develop, now that #707 is merged.

Issue #710, opened 2019-02-13
Add Tilegearden executor role to Terraform config


KlaasH:
The Tilegarden Lambda function needs a few IAM permissions to run.  Claudia.js creates a role with the permissions necessary (writing to logs and creating VPC network interfaces and logs), but for [caching on S3](https://github.com/azavea/pfb-network-connectivity/pull/673) it also needs some S3 permissions.  And the Lambda-warming setup will require some Lambda execution permissions.

There's no way to tell Claudia what permissions to add to the role it creates, but you can tell it to use a pre-existing role.  So we should use Terraform to create a role with the permissions we need then pass that role name (using a fixed, environment-based format so we don't have to add a variable) to Claudia.

At the moment, the permission has been defined by hand and attached by hand to the staging executor role, so we'll need to do the same for production before deploying the caching functionality.

PR #712, opened 2019-02-14
Enable Lambda warming on Tilegarden function

### LINKED ISSUE #701: Create and implement a Lambda-warming strategy
The cold start time for the Tilegarden lambda functions is too long.  My understanding is that this is likely due mainly to the latency in establishing VPC connections, so there's not much we can do to optimize it away.

The number of concurrent requests is fairly high, since the first interaction with the tiler is loading a map full of tiles.  So warming one copy won't be much of an improvement over loading none--it would likely handle one tile then, even if it's ready to process more, the others would already be dispatched to new instances and be waiting on those to spin up.

One thought to explore: people will almost always be getting to the map view via the places list, so rather than a purely schedule-based warming strategy, we could think about whether it would make sense to use one that responds to user activity--e.g. when any request is made of `analysis-jobs` endpoint that drives the places list, place a task that launches several concurrent lambda-warming requests.  Then if users spend at least a few seconds on the list page, they would have a decent number of workers ready when they view their first map.

KlaasH:
## Overview

Since the Tilegarden Lambda functions are inside a VPC (which they need to be, since they talk to the database), [their cold start times are bad](https://medium.freecodecamp.org/lambda-vpc-cold-starts-a-latency-killer-5408323278dd).  This adds the [`lambda-warmer`](https://github.com/jeremydaly/lambda-warmer) package and configures the Lambda function's entrypoint to handle the type of invocations that trigger it.

I initially did something much more complicated (see [branch](https://github.com/azavea/pfb-network-connectivity/tree/feature/kjh/lambda-warming-endpoint)), involving a warming endpoint that could be triggered from outside and would in turn trigger a `lambda-warmer` event.  This was based on the fact that 
- to be effective, we'll need to have a decent number of workers warmed up and 
- the site doesn't get continuous usage, so keeping the Lambdas warm 24/7 would waste a lot of invocations.

I changed to this simpler strategy when I realized 
- the triggering endpoint would itself incur a cold start delay, so the time from sending the triggering request to the tiler being ready would be double the cold start time, which would be much too long
- keeping a small fleet of Lambda functions warm is really not expensive.  Using [this calculator](https://dashbird.io/lambda-cost-calculator/), with the 512MB memory size and assuming triggering every 15 minutes and the cumulative execution time of all the warming invocations per event adding up to 10 seconds (which I think is reasonable to conservative), it would cost 25 cents per month.

This adds a section to the deployment README about setting up the scheduled warming event.  I'm not sure how to do it in Terraform, and though it doesn't look super complicated, it would create another circular dependency between the Terraform config and the resources produced by the Tilegarden deployment script.  And resolving that seems somewhat complicated.  So I'm adding instructions to do it manually and [an issue](https://github.com/azavea/pfb-network-connectivity/issues/714) to hopefully automate it later.

It also adds a section to the deployment README about manually setting permissions on the executor role, which is issue #710.

### Demo

image

### Notes

A note on warming interval: There are no guarantees or even stated norms for how long a Lambda instance will sit idle before getting destroyed.  People have [done experiments](https://read.acloud.guru/how-long-does-aws-lambda-keep-your-idle-functions-around-before-a-cold-start-bf715d3b810) and come up with "usually about 45 minutes, but sometimes longer and sometimes shorter".  Using a timeout of 15 minutes seems reasonable to me--conservative but not excessively so.

## Testing Instructions

I don't think this can really be tested locally.  At the moment, it has been pushed to staging via a test branch and is running there.

Resolves #701

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -136,7 +136,27 @@ vagrant@pfb-network-connectivity:/vagrant$ docker-compose \
                                              tilegarden deploy-new
 ```
 
-### 4. Update remote state for Claudia and Terraform
+### 4. Manually add permissions
+
+The Tilegarden Lambda function needs certain permissions to work. Some are set automatically by
+Claudia, but the ones related to caching and warming aren't (see [issue #710](https://github.com/azavea/pfb-network-connectivity/issues/710)).
+
+After both Terraform and the Tilegarden deploy have run (the former to create the cache bucket,
+the latter to create the executor role), find the IAM role corresponding to your Tilegarden deployment
+(it will be the `PROJECT_NAME` you sent in the `.env` file plus `-executor`) and add two permissions:
+- An S3 policy to allow writing to the cache bucket (currently it just blindly writes, so read
+permissions aren't required, though as long as the policy is restricted to the bucket there's not
+much harm in it granting extra permissions.)
+- A Lambda policy to grant `InvokeFunction`, so that the Lambda function can trigger itself to
+generate concurrent warming invocations.
+
+### 5. Manually add a scheduled warming event
+
+In the [CloudWatch Rules console](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#rules:),
+add a scheduled event to generate warming invocations to prevent users from being subjected to cold starts.
+See [issue #714](https://github.com/azavea/pfb-network-connectivity/issues/714).
+
+### 6. Update remote state for Claudia and Terraform
 
 Once the deployment has completed, upload your `.env` file and Claudia metadata
 file to the remote state bucket so that CI can update Tilegarden automatically:
--- a/src/tilegarden/package.json
+++ b/src/tilegarden/package.json
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "claudia-api-builder": "^4.1.0",
+    "lambda-warmer": "^1.1.0",
     "mapnik": "^3.7.2",
     "sql-escape-string": "^1.1.0",
     "winston": "^3.2.1",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Upcoming release]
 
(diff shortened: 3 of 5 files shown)


src/tilegarden/package.json
```diff
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "claudia-api-builder": "^4.1.0",
+    "lambda-warmer": "^1.0.0",
```
flibbertigibbet: The latest release is `1.1.0`.

KlaasH: Weird.  I thought I had added that via `yarn add` just a few days ago.  Will fix.

deployment/README.md
```diff
@@ -136,7 +136,27 @@ vagrant@pfb-network-connectivity:/vagrant$ docker-compose \
...
+permissions aren't required, though as long as the policy is restricted to the bucket there's not
+much harm in it granting extra permissions.)
+- A Lambda policy to grant `InvokeFunction`, so that the Lambda function can trigger itself to
+generate concurrent warming invocations.
+
+### 5. Manually add a scheduled warming event
+
+In the [CloudWatch Rules console](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#rules:),
+add a scheduled event to generate warming invocations to prevent users from being subjected to cold starts.
+See [issue #714](https://github.com/azavea/pfb-network-connectivity/issues/714).
```
jeancochrane: Have you set up either the permissions or the rule in production yet? If not, we should make sure that the issue for the next production deploy captures this work.

KlaasH: This made me realize that the resources already exist and having the permission attached before the code to use it is deployed won't do any harm, so I just added the permission.  The CloudWatch rule should happen after the deploy, since the code that's up now won't know what to do with the events, so it would just generate errors.  I made an issue for the deploy (#717) and added a note to it.

src/tilegarden/package.json
KlaasH: Fixed.  It was actually installing 1.1.0 already, since `^` only pins major version...

KlaasH:
I rebased this onto develop and pushed it to the `test/` branch to get it back on staging.

PR #713, opened 2019-02-14
Ignore job ID in imported shapefiles

### LINKED ISSUE #692: Job ID for imported analysis should always be set to the specified job
I have two dev instances now, and I ran an analysis job in one then zipped up the results and tried to import it into the other.  It failed with many lines like:
> Ignoring Feature ID 3068 because: No ForeignKey AnalysisJob model found with keyword arguments
: {'uuid': u'c6646c2b-5509-4626-aaae-b749f37796de'}

Turns out we're handling "no JOB_ID" (e.g. from ad hoc local analysis runs) and "reimport to the same job" (e.g. importing geometry for the existing jobs) OK, but having a job ID in the shapefile properties that doesn't match the target job makes the layer imports not work.

KlaasH:
## Overview

When importing a shapefile, the job ID should always be set to the job we're importing it for.  We've hit a few different issues from missing or mis-matched job IDs in input shapefiles, but we can just ignore the value (or lack thereof) in the shapefile completely and set it to the target job ID after import.

So this does that. Since there doesn't seem to be a way to set the value of a field to a literal string via the layer mapping, this relies (as we were doing before in the event of job id errors) on setting the job ID after import on all geometries with no job ID.  Since this could theoretically cause an issue if multiple imports were running simultaneously, this puts the import in a transaction so there won't ever be ID-less geometries actually committed to the database.

This also changes the error behavior on failed import so that it always raises, so we won't get uploads that get marked `COMPLETE` and show up on the site when they didn't import cleanly.

### Demo

image

## Testing Instructions

The import that failed on production was: [this file](https://s3.amazonaws.com/production-pfb-storage-us-east-1/results_2019_city_speed/provincetown-ma.zip), for [Provincetown, MA](https://s3.amazonaws.com/production-pfb-storage-us-east-1/neighborhood_boundaries/provincetown-ma/provincetown-ma.zip).  Without this change it should get a `ValidationError` from the attempted import then move on, resulting in an import that's marked complete but has no geometries.  With this change, it should import successfully.

Resolves #692.

### DIFF
--- a/src/django/pfb_analysis/management/commands/import_results_shapefiles.py
+++ b/src/django/pfb_analysis/management/commands/import_results_shapefiles.py
@@ -10,8 +10,9 @@ import zipfile
 import boto3
 
 from django.conf import settings
-from django.contrib.gis.utils import LayerMapping, LayerMapError
+from django.contrib.gis.utils import LayerMapping
 from django.core.exceptions import ValidationError
+from django.db import transaction
 
 from pfb_analysis.models import (
     AnalysisJob,
@@ -26,7 +27,6 @@ logger = logging.getLogger(__name__)
 CENSUS_BLOCK_LAYER_MAPPING = {
     'geom': 'POLYGON',
     'overall_score': 'OVERALL_SC',
-    'job': {'uuid': 'JOB_ID'},
 }
 
 NEIGHBORHOOD_WAYS_LAYER_MAPPING = {
@@ -37,7 +37,6 @@ NEIGHBORHOOD_WAYS_LAYER_MAPPING = {
     'ft_bike_in': 'FT_BIKE_IN',
     'tf_bike_in': 'TF_BIKE_IN',
     'functional': 'FUNCTIONAL',
-    'job': {'uuid': 'JOB_ID'},
 }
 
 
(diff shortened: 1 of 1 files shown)


src/django/pfb_analysis/management/commands/import_results_shapefiles.py
```diff
@@ -74,23 +75,13 @@ def import_shapefile(job, shpfile_name, model, layer_mapping):
                                             import_file,
                                             layer_mapping)
             import_layer_map.save()
-        except LayerMapError as e:
-            # If we failed with an error related to JOB_ID not found, it's because this is an
-            # old job that doesn't contain the column in the shapefile. So we try again with
```
flibbertigibbet: Are imports of jobs before the export column was added no longer supported now then? If it fails due to it missing, will that be clear to the user?

flibbertigibbet: Oh, it's not going to be used at all, okay.

KlaasH:
> I suppose it's fine the job ID is still written to the Shapefile, although it is now ignored.

Yeah, I think #607 was focused on the case of importing geometries as the last step of a regular analysis run and I was very focused here on the case of local uploads.  I do think this approach works fine--it handles a wider range of cases and does just as good a job of handling the case of importing during an analysis run--but it's a little odd that we came full circle.  And yeah, I think having the job ID in the shapefile doesn't do any harm, and there could conceivably be a benefit to being able to trace a shapefile back to its job ID even if it gets separated from its original context.

PR #719, opened 2019-02-27
Define Tilegarden executor role in Terraform config

### LINKED ISSUE #710: Add Tilegearden executor role to Terraform config

KlaasH:
## Overview

The Tilegarden Lambda function needs some permissions, specifically
- write logs
- attach to a VPC
- write to the S3 cache
- invoke itself for warming

If you don't provide a role to Claudia.js, it makes one that has the first two, but since the other two are for custom Tilegarden functionality, they need to be added.  The most straightforward way to do that is to define the whole role in Terraform and just hand it to Claudia.js in the .env file.

So this adds the Terraform code to do that, moves the `LAMBDA_ROLE` variable out of the "Optional" section of the .env.example file, and removes the section from the deployment instructions saying you have to do it manually.

### Demo

I ran terraform `plan` and `apply` manually for staging, so here's a screenshot of the created role:
image

### Notes

- I put a few extra permissions on the "write to cache" role because I'm imagining future worlds in which the function does some amount of cache evaluation/invalidation.  I don't think there's any harm in having them there, though.  It's pretty restricted (only a few permissions and only on the cache bucket).
- It seems like Claudia only sets the executor role on `create` (i.e. our `deploy-new` command), not on update.  So setting the role in the .env file and deploying didn't cause it to change.  So I changed it manually.  I didn't try it here, but from other experience I believe it would do the right thing (assign the specified role) for `deploy-new`.

## Testing Instructions

I already deployed this manually to staging.  It's probably not worth creating a whole new stack to test it from scratch, though you could delete the role and confirm that it recreates it properly.

## Checklist

Resolves #710

### DIFF
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -192,3 +192,101 @@ resource "aws_iam_instance_profile" "batch_container_instance" {
   name = "${aws_iam_role.batch_container_instance_ec2.name}"
   role = "${aws_iam_role.batch_container_instance_ec2.name}"
 }
+
+#
+# Tilegarden executor
+#
+data "aws_iam_policy_document" "lambda_assume_role" {
+  statement {
+    effect = "Allow"
+
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+
+    actions = ["sts:AssumeRole"]
+  }
+}
+
+data "aws_iam_policy_document" "lambda_invoke" {
+  statement {
+    effect = "Allow"
+    resources = ["*"]
+    actions = [
+      "lambda:InvokeFunction",
+    ]
+  }
+}
+
+data "aws_iam_policy_document" "logs_create_and_write" {
+  statement {
+    effect = "Allow"
+    resources = ["arn:aws:logs:*:*:*"],
+    actions = [
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+    ]
+  }
+}
+
+data "aws_iam_policy_document" "vpc_access" {
+  statement {
+    effect = "Allow"
+    resources = ["*"],
+    actions = [
+      "ec2:CreateNetworkInterface",
+      "ec2:DeleteNetworkInterface",
+      "ec2:DescribeNetworkInterfaces",
+    ]
+  }
+}
+
+data "aws_iam_policy_document" "s3_write_tile_cache" {
+  statement {
+    effect = "Allow"
+    resources = [
+      "arn:aws:s3:::${lower(var.environment)}-pfb-tile-cache-${var.aws_region}",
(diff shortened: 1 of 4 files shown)


deployment/terraform/iam.tf
```diff
@@ -192,3 +192,101 @@ resource "aws_iam_instance_profile" "batch_container_instance" {
...
+    }
+
+    actions = ["sts:AssumeRole"]
+  }
+}
+
+data "aws_iam_policy_document" "lambda_invoke" {
+  statement {
+    effect = "Allow"
+    resources = ["*"]
```
jeancochrane: Just want to confirm that this policy was originally scoped to the specific Lambda function in question, but that we're going to open it up to all resources because the Lambda function is defined in Claudia. I think this is fine given the deployment constraints we're under, but I also want to call out that I don't understand the security implications of it.

deployment/terraform/iam.tf
```diff
@@ -192,3 +192,101 @@ resource "aws_iam_instance_profile" "batch_container_instance" {
...
+      "arn:aws:s3:::${lower(var.environment)}-pfb-tile-cache-${var.aws_region}/*",
+    ]
+    actions = [
+      "s3:ListBucket",
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:DeleteObject",
+      "s3:GetLifecycleConfiguration",
+      "s3:PutLifecycleConfiguration",
+      "s3:GetBucketWebsite",
```
jeancochrane: This is nice and simple compared to the permissions we wound up setting manually! Glad you were able to pare it down to the essentials.

deployment/terraform/iam.tf
KlaasH: Yeah, good point.  This would mean that rather than only being able to invoke itself, the Tilegarden Lambda would be able to invoke any other Lambda function.  Locking it down more would be better, but I think would require defining a new Terraform variable (so that the permission could say e.g. `resources = ["arn:aws:lambda:::function:${var.tilegarden_project_name}"]`) and instructing future deployers of new stacks to make sure the `PROJECT_NAME` in `.env` matches the Terraform variable.

Not a huge deal, but since I don't think there's any real risk from leaving the permission less restricted, I don't think it's worth the added complexity.

jeancochrane: Agreed 👍

Issue #715, opened 2019-02-14
Delete old pre-rendered tiles from S3


KlaasH:
The tiles from the old "render everything and serve it from S3" strategy aren't being used anymore, and they add up to quite a bit of space on S3.  We should delete them.

Unfortunately they're not in their own bucket or even a single path that contains only them--they're nested under `results/{JOB_ID}/tiles`.  Deleting them with `aws s3 rm` would take a really long time, so it would be nice to be able to apply an expiration rule to get rid of all of them in one fell swoop.  But I don't know if it's possible to put a wildcard in the middle of an expiration rule.  And caution is called for, since we wouldn't want to delete the rest of the contents of the results directories.

KlaasH:
Object expiration rules can be applied by prefix or by tag, but not by wildcard.  Tags have to be added to each file, so presumably adding them then setting an expiration rule would be no less trouble than just doing the `rm` command.

Running
```
aws s3 rm --recursive --quiet s3://STORAGE_BUCKET/results/ --exclude "*" --include "*/tiles/*"
```
would take a very long time.  Running it for one job's directory, i.e. adding an ID to the `s3://` path, took about 2 minutes, and there are 172 job directories in staging and 1783 in production.  Though several hundred of those--the ones from the most recent analysis run--won't have tiles in them.  Also, running it from one of the EC2 instances makes it go significantly faster (24s), which makes sense since it's mostly a network operation.

So yeah, I don't think there's a clever way to do this, but it doesn't require much action, just launching a few commands and waiting a long time for them to finish.

KlaasH:
I finally did this.  It kept bugging me that all that cruft was sitting there even though it would be so easy to delete it, just very slow.  So I logged into one of the app instances, started a tmux session, and launched the `rm` commands.  It took 10 days, but required basically no supervision—I just checked on it occasionally to see if it was still going.  The final tally was more than 6 million files deleted from the staging bucket and more than 60 million from production.

PR #720, opened 2019-02-28
Remove tilemaker

### LINKED ISSUE #611: Remove old tiling container, code, and config

KlaasH:
## Overview

Removes all code and config related to the old "tilemaker" static tile pre-rendering process.

### Notes

- I didn't remove the artifacts like batch queues, containers in ECR, and tiles on S3.  The latter is the subject of issue #715.  I don't think the other things really have a marginal cost.
- The `use_tilegarden` variable should be removed from the staging tfvars file when this is merged and from the production file when this is deployed.

## Testing Instructions

- Everything should work the same as before, provided you had `USE_TILEGARDEN` set to true before.
- I ran an analysis job locally to check for regressions and to make sure it went to COMPLETE and showed up on the site.  Running a batch upload seemed like overkill, since there's no reason this should affect that specifically.

## Checklist

Resolves #611


KlaasH:
> pfb_aws_batch_analysis_job_queue_name and pfb_aws_batch_analysis_job_definition_name are still defined in group_vars/all.example; it looks like those are no longer needed.

There's just one queue now, so possibly not.  I took out the tilemaker queue vars.  But the model is set up to specify the queue, which it's drawing from settings, and I think doing it explicitly like that isn't necessarily worse.

PR #728, opened 2019-03-07
Add documentation section and infra commands for migration

### LINKED ISSUE #664: Improve and document database migration deployment process
When we add migrations that have to be applied before the related new code will work properly, I don't think we have a way of staging the deploy to apply them after the new container definitions are published (which is necessary since we need to run the new version of the management command that has the new migrations) but before the new version of the Django service is activated.

If there's a way to do that in an orderly fashion with the current tools (e.g. apply task changes first then apply the remainder, including updating the services to use the new tasks, after migrations have been run) we should document it in the [deployment README](https://github.com/azavea/pfb-network-connectivity/blob/develop/deployment/README.md).  If not, we should create a process.

KlaasH:
## Overview

Adds a section to the deployment documentation about how to safely apply migrations.  Since in some cases migrations need to be applied before the API service is updated, also adds a pair of commands to `infra` to deploy only the management task, so that migrations can be run before the matching API service is updated and starts handling requests that might rely on the schema being updated.

### Demo

```
The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Your plan was also saved to the path below. Call the "apply" subcommand
with this plan file and Terraform will exactly execute this execution
plan.

Path: staging-pfb-config-us-east-1-mgmt.tfplan

-/+ aws_ecs_task_definition.pfb_app_management
    arn:                   "arn:aws:ecs:us-east-1:950872791630:task-definition/StagingManagement:401" => "<computed>"
    container_definitions: "62deef93ae6e6aa7824fff26374548bddabe0dae" => "d3eda56040ec17d18a4a7b6c778e48c1044a7770" (forces new resource)
    family:                "StagingManagement" => "StagingManagement"
    network_mode:          "" => "<computed>"
    revision:              "401" => "<computed>"

Plan: 1 to add, 0 to change, 1 to destroy.
vagrant@pfb-network-connectivity:/vagrant$ ./scripts/infra apply-mgmt

Attempting to deploy application version [a759192]...
-----------------------------------------------------
download: s3://staging-pfb-config-us-east-1/terraform/terraform.tfvars to ./staging-pfb-config-us-east-1.tfvars
aws_ecs_task_definition.pfb_app_management: Destroying... (ID: StagingManagement)
aws_ecs_task_definition.pfb_app_management: Destruction complete
aws_ecs_task_definition.pfb_app_management: Creating...
  arn:                   "" => "<computed>"
  container_definitions: "" => "d3eda56040ec17d18a4a7b6c778e48c1044a7770"
  family:                "" => "StagingManagement"
  network_mode:          "" => "<computed>"
  revision:              "" => "<computed>"
aws_ecs_task_definition.pfb_app_management: Creation complete (ID: StagingManagement)
Apply complete! Resources: 1 added, 0 changed, 1 destroyed.
```

### Notes

The Management task requires the `batch_analysis_job_definition_name_revision` variable, which means `plan-mgmt` still has to run the `update-job-defs` container to make new job revisions.  I pulled that section out into a function so it could be run by both `plan` commands.  Per the TODO in that code, it doesn't have a way of knowing whether the analysis container has actually been changed, so it just makes a new revision every time.  This doesn't do any harm, but it does mean that doing a full `plan/apply` after doing `plan-mgmt/apply-mgmt` will update the batch definitions again, which will mean that the Management task definition will get updated again.

## Testing Instructions

- Follow the deployment instructions to set environment variables (use `ENVIRONMENT=staging`) and build containers
- Run `./scripts/infra plan-mgmt`.  It should create a plan to update only one resource, `aws_ecs_task_definition.pfb_app_management`
- Run `/.scripts/infra apply-mgmt`.  It should apply the plan and update the task definition.

Actually runing a management task, since there's not enough extra memory on the staging EC2 instance to add a task, would require either adding an EC2 instance to the cluster or temporarily killing the HTTPS service's task.

## Checklist

Resolves #664.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Use Terraform to create and configure the Tilegarden executor role
 - Show country name in neighborhood labe for non-US places
 - Treat US territories as states, not countries
+- Add management-only deployment command for applying migrations before deployment
 
 #### Removed
 - Removed 'tilemaker' container and all code and config related to static tiling
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -44,6 +44,23 @@ Then edit the file and upload the modified copy, using default server-side encry
 aws s3 cp "${ENVIRONMENT}.tfvars" "s3://${ENVIRONMENT}-pfb-config-us-east-1/terraform/terraform.tfvars" --sse
 ```
 
+### Migrations
+In some circumstances, there are availability considerations around when migrations are applied relative to the Django service getting updated.  The new definition of the Management task is normally created by `infra apply` at the same time as the new definition of the API service task.  If the new API code is compatible with the pre-migration database state (e.g. an optional field is being removed, so the new code will just ignore it), you can run migrations at any time after the new tasks have been created.
+
+If the new code is incompatible with the pre-migration database schema (e.g. a new field is being added), migrations need to be applied before the API service definitions are updated with the new code.  To do this, follow the "Deploying" instructions from the next section but instead of `./scripts/infra plan` and `./scripts/infra apply`, run `./scripts/infra plan-mgmt` and `./scripts/infra apply-mgmt`.  Then after applying migrations, run the normal `plan` and `apply` commands to deploy the rest of the services.
+
+To apply migrations, once the task has been updated:
+- Open the [Elastic Container Service](https://console.aws.amazon.com/ecs/home?region=us-east-1) console
+- Select the cluster for your target environment
+- Under the "Tasks" tab, click "Run New Task"
+- Choose the "Management" task that matches the environment
+- Under "Advanced Options > Container Overrides" set the "Command override" to `migrate`.
+- Click "Run Task"
+It will show either a green "Created tasks successfully" or a red "Run tasks failed" message at the top
+of the page.  If it fails with "Reasons : ["RESOURCE:MEMORY"].", you might need to spin up a new EC2 instance or temporarily reduce the task count to enable the migration task to run.
+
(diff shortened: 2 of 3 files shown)


KlaasH:
The health check for PFB uses [django-watchman](https://django-watchman.readthedocs.io/en/latest/) with [the 'database' check](https://github.com/azavea/pfb-network-connectivity/blob/0.9.1/src/django/pfb_network_connectivity/settings.py#L238-L243), so [the endpoint](https://bna.peopleforbikes.org/healthcheck/) will say `{"databases": [{"default": {"ok": true}}]}` when all's well.  But that's just connecting to the database, not making any assertions about its state.  If the Django app fails to initialize (e.g. due to a syntax error or a missing required setting), that would make the health check endpoint not respond at all, but there's no more detailed app-level test than that.  So if there's a model with a new field, that will cause an error for any query on that model (since Django explicitly selects fields, so it will be asking for one that the database doesn't have), but the health check won't know there's a problem.

I don't know of a way to detect unapplied migrations programmatically.  [This StackOverflow thread](https://stackoverflow.com/questions/31222695/django-you-have-unapplied-migrations-which-ones) doesn't have anything better than "run `manage.py showmigrations` and grep the output for unchecked lines", which is kind of terrible.  If there were a way to do it, that could be turned into a custom django-watchman test that could accomplish the scenario you describe--keep the new API tasks from being added to the load balancer until the migrations have been applied.

But that would help in the usual scenario (old code can work with new DB state, new code breaks with old DB state) but actually be a problem in the other scenario (old code breaks with new DB state, new code can work with old DB state).  So I think having the flexibility to do it manually before or after the rest of the deploy makes sense.  Or at least is the best currently-available option.

(P.S. I really want to be able to label the above scenarios with "forward" and "backward" and "compatible" or "incompatible", but I can't get the categories to line up well enough in my head, so so far I'm stuck with the super-wordy version.)

PR #729, opened 2019-03-07
Add Terraform var for Tilegarden function name

### LINKED ISSUE #714: Add lambda-warming scheduled task to Terraform config
As noted on PR #712, the Lambda-warming strategy for Tilegarden requires a scheduled Cloudwatch event.  That PR adds a little section the deployment README about doing it by hand, but it would be much preferable to have Terraform do it.  The main complication is that it adds to the circular dependency between the Terraform config and the stuff created by Claudia.  In this case, it's the ARN of the Lambda function that would be needed but not exist the first time Terraform is run.

KlaasH:
## Overview

Adds a tfvar for the name of the Tilegarden Lambda function and uses it to restrict the Lambda execution permission on the TilegardenExecutor role.

See [discussion on PR #719](https://github.com/azavea/pfb-network-connectivity/pull/719#discussion_r260855692)

Also
- Uses the tfvar to set the name of the executor role.
- Renames the variable on the Tilegarden side to `LAMBDA_FUNCTION_NAME`, since that's what it's used for.
- Removes the `LAMBDA_ROLE` variable from the Tilegarden .env example and uses the `LAMBDA_FUNCTION_NAME` value with 'Executor' appended to match what Terraform does when creating it.

What this does *not* do is resolve #714 by incorporating the Lambda-warming scheduled task into Terraform.  It turns out giving CloudWatch the permission to invoke a Lambda function is done by [defining a policy on the Lambda function](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/resource-based-policies-cwe.html#lambda-permissions) and can't be done via IAM role or any sort of resource wildcard.

So trying to define the scheduled warming event before the actual Lambda function exists isn't possible, which means putting it in the Terraform config would add a bunch more complexity to the circular dependency between the Terraform and Claudia deployment steps.

Using a variable to set the function name is useful, though, and since I did that first on the way to configuring the CloudWatch rule and discovering this problem, we might as well keep it.

### Notes

[Here's a branch](https://github.com/azavea/pfb-network-connectivity/tree/feature/kjh/lambda-warming-in-terraform) with [the CloudWatch rule/event config](https://github.com/azavea/pfb-network-connectivity/blob/feature/kjh/lambda-warming-in-terraform/deployment/terraform/events.tf).

## Testing Instructions

If run for Staging, starting with current `develop` deployed:
- With the `tilegarden_function_name` variable set to the existing function name (pfbTilegardenStaging) `infra plan` (after the usual setup per the deployment instructions) should show only the usual service task updates (caused by the batch job revision changing) plus a change to the `InvokeLambda` permission, changing it to a more restrictive `resources` clause.
- With the variable set to something else, `infra plan` should show that plus a bunch of changes to the Executor role to remove the one with the old name and create one with the new name.

## Checklist

~- [ ] Add entry to CHANGELOG.md~

### DIFF
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -162,3 +162,6 @@ variable "tilegarden_api_gateway_domain_name" {}
 variable "cloudfront_price_class" {
   default = "PriceClass_100"
 }
+
+# Should be environment-specific and match the value in the Tilegarden .env file
+variable "tilegarden_function_name" {}
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -131,7 +131,7 @@ $ cp ./src/tilegarden/.env.example ./src/tilegarden/.env
 
 Edit the new file to fill in or adjust variables.  The required variables are:
 - `AWS_PROFILE`: the name of the AWS credentials profile you created above, e.g. "pfb"
-- `PROJECT_NAME`: a name to identify this deployment, which should include the environment name
+- `LAMBDA_FUNCTION_NAME`: a name to identify this deployment, which should include the environment name
 - `LAMBDA_REGION`
 - `LAMBDA_ROLE`: the role the Lambda function should run under. Use the one created by Terraform, e.g. "pfbStagingTilegardenExecutor"
 
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -212,7 +212,7 @@ data "aws_iam_policy_document" "lambda_assume_role" {
 data "aws_iam_policy_document" "lambda_invoke" {
   statement {
     effect = "Allow"
-    resources = ["*"]
+    resources = ["arn:aws:lambda:*:*:function:${var.tilegarden_function_name}"]
     actions = [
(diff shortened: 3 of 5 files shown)


deployment/terraform/variables.tf
```diff
@@ -162,3 +162,6 @@ variable "tilegarden_api_gateway_domain_name" {}
 variable "cloudfront_price_class" {
   default = "PriceClass_100"
 }
+
+# Should be environment-specific and match the value in the Tilegarden .env file
+variable "tilegarden_function_name" {}
```
ddohler: Does this need to get added to the `.tfvars` files on S3?

KlaasH: Yup, will do.

KlaasH:
> I noticed that GitHub's issue-resolving logic isn't smart enough to understand the negation

I wondered whether it would pick that out.  I wasn't sure if keywords like that had to be at the beginning of a line.  I guess it's probably nice that they don't in some cases, but yeah, false positives...

Issue #734, opened 2019-04-03
Add country and state/province filtering to the Places list


KlaasH:
Country and state information should be pulled from the `countries` API endpoint.  The state/province filter will have to depend dynamically on what's selected in the country filter.

Issue #733, opened 2019-04-03
Customize neighborhood display labels by country


KlaasH:
For US and Canadian cities, the labels should use 2-letter abbrevs for state and country, i.e. "City, ST, CO".  People For Bikes will provide details of what the labels should be for specific other countries and what the default should be for countries not explicitly configured.

Since the config lives in the Django settings, it probably makes sense to just compose the labels and add them to the API response rather than sending the config and the elements to the front-end and composing the label there.

Issue #732, opened 2019-04-03
Store 'state/province' for non-US neighborhoods


KlaasH:
Update the Neighborhood entry form to ask for state/province for countries where we want to track that information.

Which countries we want to track state/province for should already be defined in the country config dictionary in setting.py.

This will require a model change, since the current field is a `USStateField`.

PR #736, opened 2019-04-04
Add subdivisions to 'countries' list, and country config

### LINKED ISSUE #731: Upgrade countries endpoint to provide state/province information
- Use `pycountry` rather than `django-countries`, since it includes state/province information
- Add country config to settings.py to control which countries we want states for
    - This will also be where we put label customization, so a dictionary by country probably makes sense, rather than just a list.
    - People For Bikes will provide this information along with the label customization 
- Since the front-end will need to dynamically adjust the state/province filter by country, it might make sense to just return all country and state/province information in a single response.  It would be cacheable, and if it's not prohibitively large would save the extra work of doing additional API calls to get state/province information per country.

KlaasH:
## Overview

Main changes:
- Adds a config dictionary in settings to record what countries we want to track subdivisions for.
- Adds the `pycountry` package, which provides subdivisions
- Uses the `pycountry` and the config to add subdivisions to the list returned by the `countries` endpoint, only for the countries whitelisted in the config.

Incidental changes:
- Renames the country abbreviation field from `abbr` to `alpha_2`. That's what it's called in pycountry. Not that we have to be bound by that, but I like it better--more explicit.

This leaves the `state` endpoint alone. I think it should go away, but that seems to fall under issue #732.

Though it adds `pycountry`, this doesn't remove `django-countries` or `us`. It does reduce the usage of `us`, so I think we should be able to remove that by the time this is all done.  Whether `django-countries` can/should go away will depend on whether we want to keep using `CountryField`.

### Demo

image

## Testing Instructions

- Go to http://localhost:9202/api/countries/.  It should have states for the US and provinces for Canada.
- The "Upload Neighborhood" form should work as before.

## Checklist

~- [ ] Add entry to CHANGELOG.md~
This could go in the CHANGELOG, but I don't think it's necessarily worth mentioning separately from the more visible things that will be built on it (e.g. #732, #733, and #734).

Resolves #731

### DIFF
--- a/src/django/pfb_analysis/countries.py
+++ b/src/django/pfb_analysis/countries.py
@@ -1,0 +1,55 @@
+from django.conf import settings
+import pycountry
+
+
+def get_country_config(alpha_2):
+    return settings.COUNTRY_CONFIG.get(alpha_2, settings.COUNTRY_CONFIG['default'])
+
+
+def use_subdivisions(alpha_2):
+    return settings.COUNTRY_CONFIG.get(alpha_2, settings.COUNTRY_CONFIG['default'])['subdivisions']
+
+
+def subdivisions_for_country(alpha_2):
+    """ Subdivisions for the given country, if we want to track them
+
+    Returns a list of subdivisions, as {name, code, type}, for the given country, but
+    only if the country config in settings says to track subdivisions for the country. Otherwise
+    returns None.
+    """
+    if use_subdivisions(alpha_2):
+        return sorted(
+            # Note: pycountry's subdivision codes all start with the alpha_2 country code and
+            # a hyphen. Since we're attaching them to the country and don't expect to be
+            # mingling subdivisions from different countries, this strips off the prefix and
+            # keeps only the actual subdivision code.
+            [{'name': s.name, 'code': s.code[3:], 'type': s.type} for s in
+             pycountry.subdivisions.get(country_code=alpha_2)],
+            key=lambda s: s['name']
+        )
+    return None
+
+
+def build_country_list():
+    """ Build a list of countries, without US territories and with subdivisions if desired
+
+    Builds a list of {alpha_2, name, subdivisions} dictionaries, where 'subdivisions' is
+    a dictionary of {code, name, type} subdivisions, but is only present if the country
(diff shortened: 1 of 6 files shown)


flibbertigibbet:
Why are there Docker base image changes here? Would that be helpful to note in the changelog?

src/django/pfb_analysis/countries.py
```diff
@@ -0,0 +1,51 @@
...
+def subdivisions_for_country(alpha_2):
+    """ Subdivisions for the given country, if we want to track them
+
+    Returns a list of subdivisions, as {name, code, type}, for the given country, but
+    only if the country config in settings says to track subdivisions for the country. Otherwise
+    returns None.
+    """
+    if use_subdivisions(alpha_2):
+        return sorted(
+            [{'name': s.name, 'code': s.code[3:], 'type': s.type} for s in
```
flibbertigibbet: What is in the field getting sliced for the code? Could `code` ever be empty or too short for the slice?

src/analysis/Dockerfile
```diff
@@ -53,7 +53,8 @@ RUN set -xe && \
       git clone --branch $GIT_BRANCH_QUANTILE https://github.com/tvondra/quantile.git && \
         (cd quantile && make install) && \
     \
-    pip install --upgrade pip && \
+    pip install --upgrade pip setuptools && \
+    hash -r && \
```
flibbertigibbet: A comment here for the reason for using `hash` might come in handy later.

KlaasH:
Whoops, I based this off of the branch for https://github.com/azavea/pfb-network-connectivity/pull/737 to get the CI build to work, but forgot to point the PR there.  I'm still working on the container fixes, but I believe they're working well enough for the purposes of these changes.

src/django/pfb_analysis/countries.py
KlaasH: Oops, should have commented that.  The subdivision codes all start with the alpha_2 country code and a hyphen.  We could use the whole thing and strip the prefix for presentation, but I don't think having the prefixes is likely to be useful, so it seems simpler to strip them off up front.

Issue #735, opened 2019-04-03
Change "State" terminology to "State/Province"


KlaasH:
On the public site and in the admin views.

PR #739, opened 2019-04-09
Support states/provinces for non-US cities

### LINKED ISSUE #732: Store 'state/province' for non-US neighborhoods

KlaasH:
## Overview

Converts the `Neighborhood.state_abbrev` field from supporting only US states to supporting states/provinces for any country.  Whether we actually collect state/province for a given country depends on the `COUNTRY_CONFIG` in settings.  The front-end relies on the subdivision information included in the `countries` API endpoint, and the back-end validation checks that the selected country should have subdivisions and that the selected subdivision abbreviation is one of the valid ones.

### Demo
image

### Notes

This doesn't change the labeling (see #733 for that), so the change isn't really perceptible on the public site.  On the admin site, though, the form has changed a bit--Country is before State, now, since having a lower question affect the display of an earlier one seems a bit confusing, and the State field will now stick around when Canada is selected. (see #735 for adding Province to the nomenclature)

## Testing Instructions

- Go to the [Create Neighborhood form](http://localhost:9301/#/admin/neighborhoods/create/).
   - Entering a US city should work as before (except with the fields in a different order). State should be required.
   - Entering a Canadian city should now require that you select the province.
   - Selecting any country besides those two should make the State field disappear.
   - If you pick a state then change to a different country that doesn't support states, the state should not get submitted with the save request.
- Go to the [API interface](http://localhost:9202/api/neighborhoods/) and try to submit an invalid neighborhood, such as
   - US or Canada with no state
   - US or Canada with an invalid state
   - Any other country with a value for state
- The public site Places list and details should be unchanged.

## Checklist

Resolves #732


src/angularjs/src/app/neighborhoods/create/neighborhoods-create.html
```diff
@@ -18,30 +18,30 @@ <h1>Create Neighborhood</h1>
...
-                  ng-options="state as state.name for state in neighborhoodCreate.states
-                  track by state.abbr"
-                  class="form-control" id="role" type="text">
+                  ng-options="state as state.name for state
+                    in neighborhoodCreate.country.subdivisions track by state.code"
+                  class="form-control" id="role" type="text" required>
           </select>
         </div>
-        <div class="form-group" ng-if="neighborhoodCreate.isDefaultCountry()">
+        <div class="form-group animationIf" ng-if="neighborhoodCreate.isDefaultCountry()">
```
flibbertigibbet: Fancy fading :art:

PR #744, opened 2019-04-12
Sort places by name, remove sorting by state

### LINKED ISSUE #730: Change default sorting of Places list to alphabetical by name
Sorting by state will no longer be an option, and we can remove the code to support it.  We will be adding filtering by state instead.

KlaasH:
## Overview

Removes the option to sort places by state then name, which had been the default.  Now the default is alphabetical by place name across all states and countries.

(Issue #734 will add filtering to solve the problem of finding a place by country or state.)

### Demo

image

### Notes

This also includes a small code refactor to the next and previous buttons.  As noted in the commit message, they tripped me up for a minute and then bothered me enough to spend a few minutes reconfiguring them a bit to reduce moving parts and repetition.  I believe they're functionally unchanged.

## Testing Instructions

- Load the places list, behold the alphabetical sorting.
- Use the "Previous" and "Next" buttons (set `REST_FRAMEWORK['PAGE_SIZE']` to something small if you don't have enough places for the default page size) and confirm that they still work as expected.

## Checklist

Resolves #730.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 #### Changed
 - Updated containers based on Debian Jessie to Stretch
 - Support entering state/province for non-US neighborhoods
+- Default to sorting places list by name and remove order-by-state option
 
 ## [0.9.2] - 2019-03-11
 
--- a/src/angularjs/src/app/places/list/place-list.html
+++ b/src/angularjs/src/app/places/list/place-list.html
@@ -111,10 +111,10 @@
         <section class="paginate">
             <button class="btn btn-default paginate-previous"
                 ng-click="placeList.getPrev()"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 #### Changed
 - Updated containers based on Debian Jessie to Stretch
 - Support entering state/province for non-US neighborhoods
+- Sort Places by name by default and remove order-by-state option
```
flibbertigibbet: Is the capitalization here intentional?

KlaasH: I think I erratically capitalize that.  It's not really a term of art in this context, but it's not 100% not one.  It's also closely related to the Neighborhood model, and I do usually capitalize model names.  So my answer is "maybe" :slightly_smiling_face:

PR #755, opened 2019-08-12
Migrates analysis scripts to Python 3

### LINKED ISSUE #752: Upgrade python scripts in 'src/analysis' to Python 3
See http://python-future.org/automatic_conversion.html#futurize-quick-start-guide.
The analysis code is mostly bash and SQL, but there are 2 python scripts. The changes are likely minimal.

ddohler:
## Overview

Updates the Python scripts in `/src/analysis` to Python 3.

### Notes

Although backward compatibility with Python2 was not a requirement, and the code required only minimal changes to achieve Python3 compatibility, I discovered that the import statements necessary to achieve backwards compatibility can serve a useful signaling function to both automated and human readers that the code has been audited for Py3 compatibility. For example, performing automated conversion on `detect_utm_zone.py` initially resulted in the following diff:
```
--- scripts/detect_utm_zone.py	(original)
+++ scripts/detect_utm_zone.py	(refactored)
@@ -5,6 +5,8 @@
 Based on https://github.com/jbranigan/geo-scripts-python/blob/master/latlng2utm/detect-utm-zone.py
 """
 from __future__ import print_function
+from __future__ import division
+from past.utils import old_div
 
 import argparse
 import math
@@ -32,7 +34,7 @@
     """ Finds the UTM zone of a WGS84 coordinate """
     # There are 60 longitudinal projection zones numbered 1 to 60 starting at 180W
     # So that's -180 = 1, -174 = 2, -168 = 3
-    zone = ((coord - -180) / 6.0)
+    zone = (old_div((coord - -180), 6.0))
     return int(math.ceil(zone))
 
 
@@ -79,10 +81,10 @@
     check_latlng(bbox)
     check_width(bbox)
 
-    avg_longitude = ((bbox[1] - bbox[0]) / 2) + bbox[0]
+    avg_longitude = (old_div((bbox[1] - bbox[0]), 2)) + bbox[0]
     utm_zone = get_zone(avg_longitude)
 
-    avg_latitude = ((bbox[3] - bbox[2]) / 2) + bbox[2]
+    avg_latitude = (old_div((bbox[3] - bbox[2]), 2)) + bbox[2]
 
     # convert UTM zone to SRID
     # SRID for a given UTM ZONE: 32[6 if N|7 if S]<zone>
```

This is because division changed between Python 2 and Python 3, so the conversion script is doing its best to prevent a change in script behavior. However, the Python 2 code is actually somewhat problematic here -- in the unlikely situation that pure integers managed to sneak into the inputs, we would consider integer division to be a bug rather than a feature because it would make the average incorrect in most cases. So it's preferable to adopt the Python 3 behavior, which can be done by manually adding `from __future__ import division` to the file. This makes the conversion tool ignore that particular fix. This could also be done by switching off the fix from the command line, but this way is more transparent and durable because it signals to all future readers that the change in division behavior has been accounted for in this file.

Maybe this kind of signaling won't be necessary once Python 3 is widespread enough that everyone just assumes that code is in Python 3, but for the immediate future it seems like a helpful way to broadcast "this code has been migrated".

This doesn't migrate all the way to Python 3.7. Given the difficulties that caused on Cicero, it seems like waiting a year is more likely to be the most efficient use of available project budget since it should hopefully become easier to do the upgrade then.

## Testing Instructions

 * Edit `download_osm_extract.py` to add `logger.info(sys.version)` near the top of `main()`
 * Follow the README instructions to rebuild the `pfb` container
 * If you don't have it already, grab the `gtown_westside.zip` shapefile from the project S3 bucket
 * Run something like `docker run -e PFB_SHPFILE=/data/gtown_westside/gtown_westside.shp -e PFB_STATE=pa -e PFB_STATE_FIPS=42 -e NB_INPUT_SRID=4326 -e NB_BOUNDARY_BUFFER=1 -v $(pwd)/data/:/data/ pfb`
 * Confirm that your log message says the script is running under Python 3 when `download_osm_extract.py` starts up.
 * Confirm that the script completes successfully and that the analysis completes successfully.

## Checklist

Resolves #752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Support entering state/province for non-US neighborhoods
 - Default to sorting places list by name and remove order-by-state option
 - Adjust label formatting to show state/province and country
+- Upgraded analysis scripts to run under Python 3
 
 ## [0.9.2] - 2019-03-11
 
--- a/src/analysis/Dockerfile
+++ b/src/analysis/Dockerfile
@@ -41,13 +41,13 @@ RUN set -xe && \
         parallel \
         postgresql-plpython-$PG_MAJOR \
         postgresql-$PG_MAJOR-pgrouting \
-        python-gdal \
+        python3-gdal \
         gdal-bin \
         unzip \
         zip \
         postgis \
-        python-dev \
-        python-pip" && \
+        python3-dev \
+        python3-pip" && \
     apt-get update && apt-get install -y ${BUILD_DEPS} ${DEPS} --no-install-recommends && \
     \
     mkdir /tmp/build/ && cd /tmp/build && \
@@ -58,9 +58,9 @@ RUN set -xe && \
       git clone --branch $GIT_BRANCH_QUANTILE https://github.com/tvondra/quantile.git && \
         (cd quantile && make install) && \
     \
-    pip install --upgrade pip setuptools && \
+    pip3 install --upgrade pip setuptools && \
     hash -r && \
-    pip install -r /opt/pfb/django/requirements.txt && \
+    pip3 install -r /opt/pfb/django/requirements.txt && \
     \
     cd /tmp/ && rm -rf /tmp/build/ /var/lib/apt/lists/* && \
     apt-get purge -y --auto-remove ${BUILD_DEPS}
--- a/src/analysis/README.md
+++ b/src/analysis/README.md
@@ -2,10 +2,11 @@
 
 ## Docker
 
-To run the analysis in docker, first build the docker image:
+To run the analysis in docker, first build the docker image.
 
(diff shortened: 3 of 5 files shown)


ddohler:
I think your initial idea was correct and that the issue you saw can't be (directly) resolved until we upgrade the full Django app. The error is coming from `update_status` which shells out to `/opt/pfb/django/manage.py import_results_shapefiles "${PFB_JOB_ID}"`. In turn, `manage.py` specifies that it should be executed with `#!/usr/bin/env python`, so it's trying to operate inside a Py2 environment where Django no longer exists.

If we want to keep things functional for the time being, we could duplicate the Python package installation steps so that the container has copies of all the necessary libraries for both Py2 and Py3. That seems somewhat superfluous since we're planning to migrate the Django app soon anyway, but it should work. Alternatively, we could make a tracking branch to avoid polluting `develop` until we're finished. I'm good with either one -- what do you think?

Issue #756, opened 2019-08-19
Deploy Python 3 upgrade


KlaasH:
Depends on #751 and #752.

Special instructions: see https://github.com/azavea/pfb-network-connectivity/issues/579#issuecomment-520512903 re replacing the EC2 instances.

KlaasH:
Done.  The "raise ASG desired count, drain old instances, lower desired count" routine did the job for replacing the EC2 instances.

It turned out we had neglected to apply [migration 39](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/django/pfb_analysis/migrations/0039_alter_neighborhood_state_abbrev_field.py) when we deployed it, but it's applied now, along with the small handful of migrations that came with the Django upgrade.

PR #761, opened 2019-10-30
Fix build errors from Ansible version and osm2pg* installation

KlaasH:
## Overview

When I went to rebuild my instance, first it said this:
> The requested Ansible version (2.8.4) was not found on the guest.
Please check the Ansible installation on your Vagrant guest system (currently: 2.8.6),
or adapt the provisioner `version` option in your Vagrantfile.

I just removed the Ansible version specifier from the Vagrantfile entirely.  I don't see a reason not to just accept what comes with the VM, and if we're doing that, writing it down in the Vagrantfile just means it'll crash occasionally and make us update it to match.

Once I fixed that, I got a long crash beginning with
```
In file included from /usr/include/geos/geom/GeometryFactory.h:23:0,                                          
                 from /tmp/build/osm2pgsql/geometry-builder.cpp:39:                                           
/usr/include/geos/geom/Geometry.h:26:3: warning: #warning "The GEOS C++ API is unstable, please use the C API 
instead" [-Wcpp]                                                                                              
 # warning "The GEOS C++ API is unstable, please use the C API instead"                                       
   ^~~~~~~                                                                                                    
/usr/include/geos/geom/Geometry.h:27:3: warning: #warning "HINT: #include geos_c.h" [-Wcpp]                   
 # warning "HINT: #include geos_c.h"                                                                          
   ^~~~~~~                                                                                                    
/tmp/build/osm2pgsql/geometry-builder.cpp: In member function ‘geometry_builder::pg_geom_t geometry_builder::g
et_wkb_simple(const nodelist_t&, int) const’:                                                                 
/tmp/build/osm2pgsql/geometry-builder.cpp:230:25: error: ‘geos::geom::GeometryFactory::GeometryFactory()’ is p
rotected within this context                                                                                  
         GeometryFactory gf;                                                                                  
                         ^~                                                                                   
In file included from /tmp/build/osm2pgsql/geometry-builder.cpp:39:0:                                         
/usr/include/geos/geom/GeometryFactory.h:411:2: note: declared protected here                                 
  GeometryFactory();                                                                                          
  ^~~~~~~~~~~~~~~
```

We've been installing `osm2pgrouting` and `osm2pgsql` in the `analysis` container by cloning the repos and compiling them.  Presumably that was because the versions available in the Trusty `apt` sources were too old.  But now we're on Xenial, and the versions available from `apt` are slightly ahead of the ones we've been using (2.2.0 vs 2.1.0 for `osm2pgrouting`, 0.92.0 vs 0.90.1 for `osm2pgsql`).

So this removes the manual compilation and adds those two packages to the list of dependencies we're installing with `apt`.

The other manual compilation we've been doing is for `quantile`, a PostgreSQL extension.  I left that in place, since the alternatives don't seem easier.

### Notes

- This also includes a change I had sitting in my code, that I must have made while working on the Python/Django upgrades but after my last PR, to add the `django-q` container to the list in `scripts/console`.
- The versions don't match exactly, but I don't see any evidence of breaking changes in the [`osm2pgrouting` release notes](https://github.com/pgRouting/osm2pgrouting/releases).  `osm2pgsql` doesn't seem to have release notes or a current changelog, but hopefully it doesn't change much across minor versions.
- I'm not sure why this changed, but Python 2 was getting uninstalled along with the `BUILD_DEPS`. And it's needed for the AWS CLI installation. So I added it to the `DEPS` list to make it stick around.

## Testing Instructions

- Rebuild your containers (either run `scripts/update` from within the VM or destroy the VM and run `scripts/setup`)
- Everything should work as before.  I ran 1 analysis as a smoke test.

## Checklist

Connects #756

### DIFF
--- a/src/analysis/Dockerfile
+++ b/src/analysis/Dockerfile
@@ -2,10 +2,6 @@ FROM quay.io/azavea/postgis:2.3-postgres9.6-slim
 MAINTAINER Azavea
 LABEL type=pfb-analysis
 
-ENV GIT_BRANCH_OSM2PGROUTING osm2pgrouting-2.1.0
-ENV GIT_BRANCH_OSM2PGSQL 0.90.1
-ENV GIT_BRANCH_QUANTILE master
-
 # Copy django requirements (will copy the rest later, so that any change to
 # the code doesn't trigger re-running the whole provisioning step)
 COPY ./django/requirements.txt /opt/pfb/django/requirements.txt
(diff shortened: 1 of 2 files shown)


src/analysis/Dockerfile
```diff
@@ -46,17 +42,16 @@ RUN set -xe && \
...
+        python \
         python3-dev \
-        python3-pip" && \
+        python3-pip \
+        osm2pgrouting \
+        osm2pgsql" && \
     apt-get update && apt-get install -y ${BUILD_DEPS} ${DEPS} --no-install-recommends && \
     \
     mkdir /tmp/build/ && cd /tmp/build && \
-      git clone --branch $GIT_BRANCH_OSM2PGROUTING https://github.com/pgRouting/osm2pgrouting.git && \
```
ddohler: 🎉

PR #768, opened 2019-11-27
Enable neighborhood deletion, with confirmation

### LINKED ISSUE #765: Enable deleting neighborhoods
Add a delete link, with a confirmation dialog, to the [neighborhood list view](http://localhost:9301/#/admin/neighborhoods/).

The confirmation dialog should probably mention that all the associated analysis jobs will be deleted as well.

The same permissions rules that apply to adding neighborhoods should apply to deleting them.

KlaasH:
## Overview

Fixes the "Delete" column in the neighborhood list (turns out there was one there, it just wasn't working) and adds a confirmation modal to prevent rash or accidental deletion.

The problem with existing column was that it was using the wrong icon class.  Also we didn't have a suitable icon in our Fontello set, so I updated it to add one.  I also added an "Edit" icon while I was at it, since we'll be wanting one eventually.

The endpoint is the same `NeighborhoodViewset` class that handles creation, and it already has suitable permission classes ("anyone with an account can view, only the admin user in the admin org can add/edit/delete").

I used [uibModal](https://angular-ui.github.io/bootstrap/#!#modal), and put the template and controller in a new `components/` subdirectory.  I made it fairly generic, which was easy because that's how the example I was working from was set up, and also I like it because it keeps the machinery in one place and the context-specific details (text, post-confirmation action) in the context where they're relevant.

We already had `angular-bootstrap` installed, but I had to add the `bootstrap-sass-official` package to get the styles.  It took a little fiddling to figure out how to include what we need in `index.scss`, but it ended up being very straightforward.

### Demo

image

## Testing Instructions

- You'll need to have some neighborhoods in your dev database. An easy way to get them is to import a batch file--it will create a bunch of neighborhoods and analysis job instances, but the jobs won't run, they'll just print their "run this command to actually do the analysis" messages.
- Click the trash can next to one of the neighborhoods in the [Neighborhood list](http://localhost:9301/#/admin/neighborhoods/).  It should show a confirmation modal.
- Clicking "Cancel" will put you right back where you were a moment ago. Clicking "Delete" will also put you back on the neighborhood list page, but the neighborhood you just clicked will disappear.
- Any analysis jobs that existed for the neighborhood will also be gone from [the Analysis Jobs list view](http://localhost:9301/#/admin/analysis-jobs/).

## Checklist

Resolves #765


ddohler:
I will mention that I'm not quite sure what to make of the CI failure; I got the same error when I checked out this branch while `scripts/server` was running, but running `scripts/update` cleaned it up and I never saw it again.

KlaasH:
Yeah, the error is coming from the first new file getting pulled in in `index.scss` from the new `boostrap-sass-official` package, so it seems like the problem must be that it's trying to run that `gulp build` command before the files are in place.  But I don't understand why the files aren't there, when the step to install them definitely already happened.  And they're there on retry.  I don't know if it's a file caching/sync thing or if there's actually something out of order--e.g. we copy the `bower_components` directory into the location where we're going to look for them before we actually run `bower install`.
But yeah, if the symptoms are just "whenever we add a bower package we have to re-run CI once", I don't think it's worth the trouble to track it down.

PR #777, opened 2019-12-05
Delete Neighborhood and AnalysisJob files from S3 on instance deletion

### LINKED ISSUE #187: Cleanup Neighborhood resources on AWS when deleted
Currently just the uploaded shapefile

KlaasH:
## Overview

Now that we can delete neighborhoods--which means we can also delete analysis jobs, since they get deleted along with their neighborhood--we run the risk of creating a mess of orphaned files on S3.  The Neighborhood model keeps its boundary file on S3, and the results of the AnalysisJob processing are stored on S3 as well.

This adds post-delete handlers for those two models to delete the files from S3 after the instances have been deleted. For Neighborhood, which uses a `FileField`, it's just a matter of calling `.delete()` on it.  AnalysisJob doesn't have a field to track its own assets, but it does know the S3 path where they live, so we can use that to delete them with boto3.

### Demo

image

## Testing Instructions

Note: to make it easy to see files appearing and disappearing on S3, I decided to clear out the `results/` path on my development S3 bucket.  None of the files were connected to jobs I still had in my database, since I've destroyed and rebuilt my containers recently, so it was all cruft.
I did so by running this in `shell_plus`:
```
import boto3
bucket = boto3.resource('s3').Bucket(settings.AWS_STORAGE_BUCKET_NAME)
deletion_results = bucket.objects.filter(Prefix='results/').delete()
```
It took quite a while, because it turned out it was deleting 328,401 files.  I had a lot of old analysis results that included tilesets.  But now I can do `aws s3 ls s3://MY_STORAGE_BUCKET_NAME/results/ --recursive` and easily see the files I just created (or not see the ones I deleted).

- Make a neighborhood if you don't already have one.
- If you don't have a completed analysis, import one. You can use any import file URL. Here's one: `https://s3.amazonaws.com/khoekema-pfb-storage-us-east-1/uploads/marquette_results.zip`
- You should see the boundary file somewhere in `s3://YOUR_STORAGE_BUCKET_NAME/neighborhood_boundaries/` and the imported analysis results in `s3://YOUR_STORAGE_BUCKET_NAME/results/` (in a subdirectory for the analysis job's UUID).
- From the [Neighborhoods list view](http://localhost:9301/#/admin/neighborhoods/), delete the neighborhood.
- In addition to the neighborhood and job being gone from the front-end, you should see a couple messages in the log, and the files on S3 should now be gone.

## Checklist

Resolves #187

### DIFF
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -22,6 +22,8 @@ from django.contrib.gis.geos import GEOSGeometry, MultiPolygon, Polygon
 from django.contrib.postgres.fields import JSONField
 from django.core.files import File
 from django.db import models
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
 from django.utils.text import slugify
 
 import botocore
@@ -324,6 +326,12 @@ class Neighborhood(PFBModel):
     class Meta:
         # Note that uniqueness fields should also be used in the upload file path
         unique_together = ('name', 'country', 'state_abbrev', 'organization',)
+
+
+@receiver(post_delete, sender=Neighborhood)
+def delete_boundary_file(sender, instance, **kwargs):
+    instance.boundary_file.delete(save=False)
+    logger.info("Deleted boundary file for {}, {}".format(instance.label, instance.label_suffix))
 
 
 class AnalysisBatchManager(models.Manager):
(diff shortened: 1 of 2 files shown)


src/django/pfb_analysis/models.py
```diff
@@ -762,6 +770,22 @@ def _s3_url_for_result_resource(self, filename):
...
 
+@receiver(post_delete, sender=AnalysisJob)
+def delete_analysisjob_s3_results(sender, instance, **kwargs):
+    s3 = boto3.resource('s3')
+    bucket = s3.Bucket(settings.AWS_STORAGE_BUCKET_NAME)
+    path = instance.s3_results_path
+    delete_result = bucket.objects.filter(Prefix=path).delete()
+    try:
+        logger.info("Deleted {} results files from {}".format(
+            sum(len(batch['Deleted']) for batch in delete_result),
```
ddohler: One tiny nitpick: if a neighborhood is deleted that has a queued job, the job path won't exist on S3 yet, but it'll try to delete it anyway. However, this doesn't seem to cause any errors, simply: `Deleted 0 results files from results/810d88c2-0b51-4118-a6df-45467dd76d54` .

I don't think this is worth addressing now but wanted to flag it so that the behavior is known.

KlaasH: Yeah, true.  That seems fine to me, though.  The only ways I can think of to know whether there's anything to delete would be to infer from the job status or check the bucket.  The former seems fiddly and potentially brittle, and since the `delete()` call has no problem deleting zero items, adding an additional `list` call doesn't seem worth the trouble.

PR #784, opened 2019-12-17
Update LODES data retrieval to latest available (2017/2016)

beckymasond:
## Overview

The BNA retrieves 2014 LODES data from the Census (2013 data if 2014 is not available for a state). 2017 or 2016 data are now available for all states, so this PR updates the data retrieval to use the latest data available for each state.

## Notes

The LODES automated data retrieval has failed frequently lately in testing, apparently due to problems on the Census Bureau's end. When the data is manually downloaded and provided to the BNA it runs fine.

## Testing Instructions

 * Run BNA for any U.S. state.
 * For all U.S. states except Alaska and South Dakota, analysis will proceed as usual and produce Employment score at finish.
 * For Alaska and South Dakota, analysis will fail to find 2017 data and will instead default to downloading 2016 data, then proceed as usual.

### DIFF
--- a/src/analysis/import/import_jobs.sh
+++ b/src/analysis/import/import_jobs.sh
@@ -45,7 +45,7 @@ function import_job_data() {
 
     NB_STATE_ABBREV="${1}"
     NB_DATA_TYPE="${2:-main}"    # Either 'main' or 'aux'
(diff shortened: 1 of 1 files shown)


KlaasH:
Looks good :+1: 
I didn't have problems downloading census files on my few test runs, but that seems to be a problem that we can expect to keep seeing.  I made issue #786, which would be one way to deal with it.

PR #789, opened 2020-01-21
Lanes

beckymasond:
## Overview

This PR accomplishes two objectives:
1. Correcting the lanes.sql script to accommodate all likely values for the "oneway" OSM tag including 1, yes, no, -1, and NULL. This script was edited last year in #682 but that solution was only a partial fix ultimately. This solution should be a complete fix.
2. Reducing the width requirement in feet for certain infrastructure and parking combinations in order to accommodate metric entries. 1.5 meters converts to 4.9 feet, which can result in a facility + parking width of 12.9 feet (less than the required 13 feet). 1.5 meters should be an acceptable lane width so this change provides leeway for that conversion. 

### Demo

1. The changes to the lanes.sql script in 2019 corrected errors in the evaluation of oneway streets as described in #586. However, it created new errors in the interpretation of lane counts. The error is visible in Bradford Street Extension in Provincetown, MA. In the current BNA results, Bradford Street Extension has no lane counts listed in the TF_Lanes and FT_Lanes columns despite that a lane count was entered in the "lanes" tag in OSM at the time of the analysis. Using the revised lanes.sql script, the lane count appears in the TF_Lanes and FT_Lanes columns, and the street rating changes to low stress.

2. Moors Road in Provincetown, MA has a bike lane that is entered as 1.5 m wide in OSM. In the current BNA results this road rates as high stress. With the changes implemented here, Moors Road is low stress.

Original:
image

Revised:
image

## Testing Instructions

 * Run BNA for Provincetown, MA using current and revised versions of lanes.sql script. 
 * Compare output for Bradford St Ext. and Moors Road using both versions.
 * Any other BNA results including streets with a value in the "lanes" tag could be used to evaluate objective 1. Examples of bike lanes entered in meters are rare so objective 2 is harder to evaluate across cities.

### DIFF
--- a/src/analysis/features/lanes.sql
+++ b/src/analysis/features/lanes.sql
@@ -15,7 +15,7 @@ SET     ft_lanes =
                                     ),
                                     1       -- only one dimension
                                 )
-                    WHEN osm."turn:lanes" IS NOT NULL AND one_way_car = 'ft'
+                    WHEN osm."turn:lanes" IS NOT NULL AND osm."oneway" IN ('1', 'yes')
                         THEN    array_length(
                                     regexp_split_to_array(
                                         osm."turn:lanes",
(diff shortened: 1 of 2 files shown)


KlaasH:
:+1: Looks good.

PR #790, opened 2020-01-22
Revise calculation of population scores for destinations

beckymasond:
## Overview

When the user clicks on a destination in the BNA map "pop low stress," "pop high stress," and "pop score" values appear in the pop-up. For destinations retrieved from OSM as points (single nodes) residing within a single census block area, these values are accurate. For destinations retrieved from OSM as polygons that overlap multiple census blocks, these values are inaccurate. For single block destinations these values are accurate because they sum the population in all blocks within the bikeshed. For multi-block destinations these values are incorrect because they repeat the summation of the bikeshed population for every destination block without distinguishing between blocks that have already been counted. In the fixes implemented here, only the population of each **unique** block in the bikeshed for all of the destination blocks is summed.

### Demo

Running the BNA for the Algiers region of New Orleans and selecting the destination in the Universities output, University of the Holy Cross, gives a pop_low_stress score of 5609, pop_high_stress score of 248036, and a pop_score of 0.0226. The pop_high_stress score is clearly too high because the population of the Algiers region is only 51755 and the population of all of New Orleans is only 345392. Using the corrected access_universities.sql script included here, pop_low_stress=4463, pop_high_stress=34097, and pop_score=0.13. As another point of comparison, University of the Holy Cross is also tagged as a school as a single node destination in close proximity to the university tag and the pop_high_stress value for that destination is 31164. 

Original:
image

Revised:
image

### Notes

The same changes has been applied to all destinations retrieved as points or polygons from OSM: 
* colleges
* community centers
* dentists
* doctors
* hospitals
* parks
* pharmacies
* retail
* schools
* social services
* supermarkets
* transit 
* universities

## Testing Instructions

 * Run BNA with and without these changes for a sample city.
 * Compare destination pop scores for all destination types listed above, particularly for multi-polygon destinations. 
 * The output for single census block destinations should not differ. The output for destinations covering multiple census blocks will be different in most cases, and accurate.

### DIFF
--- a/src/analysis/connectivity/access_colleges.sql
+++ b/src/analysis/connectivity/access_colleges.sql
@@ -67,19 +67,25 @@ SET     colleges_score =    CASE
 -- set population shed for each college in the neighborhood
 UPDATE  neighborhood_colleges
 SET     pop_high_stress = (
-            SELECT  SUM(cb.pop10)
-            FROM    neighborhood_census_blocks cb,
-                    neighborhood_connected_census_blocks cbs
-            WHERE   cbs.source_blockid10 = cb.blockid10
-            AND     cbs.target_blockid10 = ANY(neighborhood_colleges.blockid10)
+            SELECT  SUM(shed.pop)
+            FROM    ( 
+                    SELECT  cb.blockid10, MAX(cb.pop10) as pop 
+                    FROM    neighborhood_census_blocks cb,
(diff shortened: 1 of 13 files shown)


KlaasH:
:+1: Looks good.  I found a park in Kalamazoo that was reporting 163,000 high-stress population when the one right next to it said 32,000.  With the change, they both say 32,000.

PR #806, opened 2020-10-05
Add filter to Neighborhoods list

### LINKED ISSUE #787: Add column filters for Neighborhoods list
A note from Rebecca, in response to the new neighborhood editing functionality:
> finding a city in the list of Neighborhoods can be tedious if it was loaded a long time ago (there can be many pages to scroll back and scan).

Definitely true, and being able edit a neighborhood only helps if you can find it.  I think column filters like what we have on the Analysis Jobs list would make sense.  Sorting could work, too, but filtering seems like it gets at the use case more directly, and given that we already have it on another list, it shouldn't be hard to implement.

kevinearldenny:
Add filter to Neighborhoods list component

## Overview

Added a filter to the table in the Neighborhood list, repurposing the code from the filter that already existed on the Analysis Results table. Allow for users to filter by neighborhood name, state and country.

### Demo

image

### Notes

This filter operates more as a search - only exact matches are returned. For example, if you type 'Phila'... no results are returned, but if you type 'Philadelphia' the expected result is returned.

## Testing Instructions

* Load a test dataset into DB
 * Navigate to http://localhost:9200/#/admin/neighborhoods/
* Try filtering by neighborhood name, state or country

## Checklist

Resolves #787


KlaasH (commented):
I didn't follow the code path, but I see the "Neighborhood" search on the Jobs page is querying with `?search=`, which returns substring matches.  I think this should behave more like that.  Being able to match without typing out the whole exact name seems pretty important to the usefulness of this filter (and the current behavior--all the results disappearing when you start to type--is not very friendly).  Fortunately, from a little experimentation it seems like using `?search=` works to match substrings on the `neighborhoods` endpoint as well, so it might only take a few more changes to the front-end to get substring matching working.

kevinearldenny:
After a little further investigation here, it seems like the Job search might be leveraging [this filter](https://github.com/azavea/pfb-network-connectivity/blob/d2848f1616195017856f8ccf981ba19f1b4dd672/src/django/pfb_analysis/filters.py#L14) - we might need another one of these filters to be created as well.

KlaasH:
You're right, I should have looked closer at the results I was getting from `search` on the `neighborhoods` endpoint. I thought it was working because it returned results, but it was just ignoring the parameter and returning everything.

So yeah, seems like we need a class like that for Neighborhood, with similar `search` field/`filter_search` method (which will also mean removing the `filter_fields` property from `NeighborhoodViewSet` and putting the list in the FilterSet's `Meta` class).

src/angularjs/src/app/components/filters/neighborhood-filter.directive.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * @ngdoc directive
+     * @scope
+     * @name pfb.components.filters.NeighborhoodFilterDirective:pfbNeighborhoodFilter
+     *
+     * @description
+     * Directive for the neighborhood table header
+     * Filters by neighborhood and status with client-side auto-complete
```
KlaasH, line 39: Trivial, but this line is the same as the one in `AnalysisJobFilterDirective` and doesn't quite apply here. Could revise to fit, or just drop the line.

KlaasH (approved):
:+1: Looks great!

PR #807, opened 2020-10-09
Add default speed to AnalysisJob model and display on the frontend

### LINKED ISSUE #804: Add default speed limit information to analysis results and show in front-end
The analysis knows the default speed limit--either from a city-specific one or from the state-level defaults--and uses it for stress calculations.
We want to display that information to the user when they're looking at a neighborhood's scores etc, so we need to
1. export it into the analysis results files
2. import it into the AnalysisJob model (it doesn't really belong with `overall_scores`, so probably it makes sense as a new field)
3. display it in the front-end.  I don't think there's a design for this, but I believe there was discussion about showing it in a corner of the map.

kevinearldenny:
## Overview

The default speed limit for a Neighborhood is an important parameter in the BNA analysis, and it needs to be passed along so it can be displayed on the frontend. It is now being exported as part of the analysis process, and then imported into the `default_speed_limit` field on the AnalysisJob model through the `import_default_speed` management command. 

It is then displayed on the frontend in a row in the sidebar table, and as a Leaflet control on the bottom of the map.

### Demo

image

image

## Testing Instructions

 * Create a Neighborhood, ensuring that it's FIPS code is correct since this is used in the matching of default_speed
 * Run an analysis job for the newly created neighborhood (Boulder shown in screenshots)
 * Validate that default_speed_limit field has been copied to the AnalysisJob in the [Analysis Jobs admin page](http://localhost:9200/#/admin/analysis-jobs/), as shown in the 2nd screenshot
 * Navigate to the Place Detail page. The default speed limit should be displayed in a Leaflet Control in the bottom right next to the map legend.

## Checklist

Resolves #804


kevinearldenny:
@KlaasH I've updated the display on the frontend, and added an export of `default_speed.csv` - this should be ready for review now.

KlaasH (changes requested):
The speed limit box and the code behind it look good.  Is the fixed positioning necessary, though?  I thought that was needed to get it to show up to the left but for stacking it we could just use `position: "bottomright" ` and let Leaflet worry about how it interacts with the other legend.

Right now it's still shown as `bottomleft` in `map.constants.js`, but seemingly that's being fully overridden, since it's showing up in the bottom right.  It's also doing this:
image

On the analysis/back-end stuff, I'm going to try to draft up my thoughts as code.

Issue #808, opened 2020-10-16
Missing migrations


flibbertigibbet:
Attempting to run Django migrations within a newly built VM (a step in the README) results in the error message: `Your models have changes that are not yet reflected in a migration, and so won't be applied. Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.`

Output from `makemigrations`:
![pfb_missing_migrations](https://user-images.githubusercontent.com/960264/96270281-6e4ec100-0f99-11eb-9213-09c5eccb01f7.png)

PR #809, opened 2020-10-16
Cache Census LODES data on S3

### LINKED ISSUE #786: Store all Census files used by analysis on S3
Issue #566 talks about applying the same caching strategy we set up for the OSM extracts to the census block files.  There's [a comment there](https://github.com/azavea/pfb-network-connectivity/issues/566#issuecomment-464068754) that caching would also help with the problem of analysis failures resulting from the download from census.gov failing.  There's also a note on PR #784 that the LODES jobs data downloads were causing similar problems.

The caching strategy for the OSM extracts isn't wildly complex, but it's not simple.  It includes lifecycle rules to clear out old files and locking so that concurrent jobs can download a file exactly once then share it.  That's all necessary because OSM data changes all the time and it's not easy to know ahead of time exactly what we'll want when.

But the census data doesn't change often--the LODES files are annual and we've been using the same block files since we launched this project.  So we could do away with the automated cache invalidation and negotiation and just make a simple script, to be run manually as needed, that downloads everything we need and dumps it in an S3 bucket.

The analysis could still fall back to downloading from census.gov if the file it's looking for isn't present in the cache, though the most likely cause of that would be a local job not having the cache bucket configured.

So the proposal is to set up a bucket, make a script to copy all the Census files we use into it, and change the analysis to look there first.  This would be in lieu if #566.

flibbertigibbet:
## Overview

Cache Census LODES data on S3. Only fetch files directly from the Census Bureau website if they aren't in the S3 cache, and then upload them to the S3 cache.

This also adds a convenience script `./scripts/run-lodes-cache-update` to populate the S3 cache for all available states and both data types.

### Demo

Running `./scripts/run-lodes-cache-update` the first time:
pfb_cache_lodes_initial

Running `./scripts/run-lodes-cache-update` again (with a populated cache):
pfb_cache_lodes_secondary

Analysis results, with employment data set:
pfb_reston_results_with_employment

### Notes

I could run the cache-populating script for the staging and/or production environments, if someone could point me to the appropriate S3 bucket locations for those environments.

Previously, the import script would check S3 first before downloading LODES data from the Census Bureau site but not upload to it to S3 if it wasn't there, which makes me think that a one-off script had been run at some point to populate the S3 cache, but that script was not run again when the LODES years in use got updated. I didn't find any pre-existing script in the repository to do that.

I ran into #808 while working on this.

## Testing Instructions

 * Set `PFB_DEBUG` in the VM if you'd like to see full debug output from the script
 * Run `./scripts/run-lodes-cache-update` in the VM
 * On first run, it should not find cached LODES files in your personal dev environment S3 bucket
 * Kill the script after it finishes at least one upload
 * Re-run the script. Previously processed states should be on S3; others should download from Census and upload to S3
 * Run an analysis
 * Analysis results should still look as expected and contain an employment accessibility figure

## Checklist

Resolves #786

### DIFF
--- a/src/analysis/scripts/download_census_lodes.py
+++ b/src/analysis/scripts/download_census_lodes.py
@@ -40,7 +40,7 @@ LODES_DATA_TYPES = ('main', 'aux')
 LODES_MOST_RECENT_YEAR = '2017'
 LODES_FALLBACK_YEAR = '2016'
 LODES_URL = 'http://lehd.ces.census.gov/data/lodes/LODES7'
-S3_LODES_DIRECTORY = 'data/'
+S3_LODES_DIRECTORY = 'data'
 STATE_ABBREVIATIONS = [state.abbr.lower() for state in us.STATES]
 
 
--- a/src/analysis/scripts/download_census_lodes.py
+++ b/src/analysis/scripts/download_census_lodes.py
@@ -168,12 +168,12 @@ def process_download(state, data_type, local_dir, bucket=None, check_only=False)
 
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--local_dir", default=None,
+    parser.add_argument("--local-dir", default=None,
                         help="The directory to put downloaded LODES files in")
-    parser.add_argument("--state_abbrev", default=None, help="state abbreviation")
-    parser.add_argument("--data_type", default=None, choices=LODES_DATA_TYPES,
+    parser.add_argument("--state-abbrev", default=None, help="state abbreviation")
+    parser.add_argument("--data-type", default=None, choices=LODES_DATA_TYPES,
                         help="data type (main or aux)")
-    parser.add_argument("--storage_bucket", default=None, help="S3 storage bucket")
+    parser.add_argument("--storage-bucket", default=None, help="S3 storage bucket")
     parser.add_argument('--verbose', '-v', action="store_true")
     args = parser.parse_args()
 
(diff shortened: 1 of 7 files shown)


src/analysis/scripts/download_census_lodes.py
```diff
@@ -0,0 +1,229 @@
...
+    datefmt='%Y-%m-%d %H:%M:%S'
+)
+logger = logging.getLogger(__file__)
+
+LODES_DATA_TYPES = ('main', 'aux')
+# fall back to an older year if the most recent in use is not available
+LODES_MOST_RECENT_YEAR = '2017'
+LODES_FALLBACK_YEAR = '2016'
+LODES_URL = 'http://lehd.ces.census.gov/data/lodes/LODES7'
+S3_LODES_DIRECTORY = 'data/'
```
KlaasH, line 43: I think this trailing slash is causing the S3 keys to come out like `data//hi_od_main_JT00_2017.csv.gz`.

KlaasH (commented):
Started looking at this, haven't gone over it comprehensively yet, but per the comment below, I initially had trouble finding the files on S3 because they've got a double slash in the S3 key.

And this is only doing the jobs files, but the census block files should get the same treatment.

flibbertigibbet:
I've updated this to add caching scripts for the Census block data as well.

KlaasH:
Closing in favor of PR #812

PR #810, opened 2020-10-16
Update Tilegarden node to 12.x

### LINKED ISSUE #801: Upgrade Node runtime for tilegarden
A staging deploy failed with "InvalidParameterValueException: The runtime parameter of nodejs8.10 is no longer supported for creating or updating AWS Lambda functions. We recommend you use the new runtime (nodejs12.x) while creating or updating functions."

The existing production tiler is working fine.  I skipped that part of the deployment last time, since there were no changes, and we could presumably do that again if necessary.  Until either we want to change something or something else happens that requires a full redeploy.  Best to be ready.

I don't think there's any reason to expect problems upgrading the runtime, though we'll also want to upgrade dependencies, which means there could be trouble from `node-mapnik`.

flibbertigibbet:
## Overview

Updates the Tilegarden node version to 12 locally, and updates all node dependencies.

This PR is intended to test what happens in the AWS environment when the update gets deployed to staging; if it works fine, this will close #801. I expect however that there will be AWS Lambda environment-specific errors with `node-mapnik`.

### Notes

I've manually updated the node runtime on AWS for the staging environment to 12 in preparation for merging this. I couldn't find anything in the deployment directory to set the runtime, so it appears to be a manual configuration. The Tilegarden manual deployment instructions look to be very involved, so I think the easiest way to test how this behaves in the AWS environment would be to let CI deploy to staging once this PR is merged in.

Given [this issue with node-mapnik](https://github.com/mapnik/node-mapnik/issues/937), I expect that there will be `node-mapnik` errors on AWS that do not occur locally. I think it may be helpful to confirm those errors exist before proceeding to implement a custom runtime (as described in the linked issue).

## Testing Instructions

 * Run `./scripts/update` to rebuild Tilegarden
 * Tiles should still serve locally without error

## Checklist

Connects #801

### DIFF
--- a/src/tilegarden/Dockerfile
+++ b/src/tilegarden/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.15-stretch-slim
+FROM node:12.19-stretch-slim
 
 ENV BASE_DIR /opt/pfb/tilegarden
 
--- a/src/tilegarden/package.json
+++ b/src/tilegarden/package.json
@@ -48,6 +48,7 @@
     "claudia-api-builder": "^4.1.2",
     "lambda-warmer": "^1.2.1",
     "mapnik": "^4.5.2",
+    "pg": "^8.4.1",
     "sql-escape-string": "^1.1.0",
     "winston": "^3.3.3",
     "xml2js": "^0.4.23"
--- a/src/tilegarden/package.json
+++ b/src/tilegarden/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tilegarden-pfb",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "AWS Lambda mapnik renderer",
   "contributors": [
     {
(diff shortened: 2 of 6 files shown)


KlaasH:
I pushed a test branch which Jenkins built and deployed to staging (branch [here](https://github.com/azavea/pfb-network-connectivity/tree/test/update-tilegarden-node%23801), CI build [here](http://urbanappsci.internal.azavea.com/job/azavea/job/pfb-network-connectivity/job/test%252Fupdate-tilegarden-node%2523801/)).
As expected, all is not well in Mapnik-land.

The [logs](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252FpfbTilegardenStaging/log-events$3Fstart$3D-604800000) are full of 
> Mapnik LOG> 2020-10-19 21:40:56: Problem loading plugin library: /var/task/node_modules/mapnik/lib/binding/lib/mapnik/input/postgis.input (dlopen failed - plugin likely has an unsatisfied dependency or incompatible ABI)

and

> 2020-10-19T21:40:56.610Z	b8f31252-d77c-4cc8-8ee0-2aeeb97c3f93	ERROR	[Error: Could not create datasource for type: 'postgis' (searched for datasource plugins in '/var/task/node_modules/mapnik/lib/binding/lib/mapnik/input')  encountered during parsing of layer 'neighborhood_waysTF' in Layer]

flibbertigibbet:
I just pushed a commit to here and the `test` branch that adds the node PostgreSQL package `pg` as a dependency, in the hopes that will pick up some missing binary, as [this comment](https://github.com/mapnik/node-mapnik/issues/863#issuecomment-644997043) seems to suggest is the issue.

KlaasH:
It looks like it's crashing the same way.

Since the build takes like 40 minutes to fully cycle, a tighter trial-and-error loop would be helpful.  I believe this will work (from the default directory, `/vagrant/`, within the VM) to manually deploy just the Tilegarden service on staging:
```
export ENVIRONMENT="staging"
export PFB_SETTINGS_BUCKET="${ENVIRONMENT}-pfb-config-us-east-1"
export GIT_COMMIT=$(git rev-parse --short HEAD)
aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/.env" "./src/tilegarden/.env"
aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/tilegarden/claudia.json" "./src/tilegarden/claudia/claudia.json"

docker-compose \
    -f docker-compose.yml \
    -f docker-compose.test.yml \
    run --rm --entrypoint yarn tilegarden deploy
```

flibbertigibbet:
I'm going to close this PR but leave the branch. There appears to be some package/binary missing from the base image in the more recent Node Lambda environments that was present previously. It's not immediately clear to me how to fix the Lambda environment issues, particularly given the use of Claudia for building and deploying the Lambda.

PR #812, opened 2020-11-03
Cache downloaded census data files to S3 within analysis

### LINKED ISSUE #786: Store all Census files used by analysis on S3

KlaasH:
## Overview

The analysis already looks on S3 for cached census data files (jobs and blocks), but there's no facility for actually filling that cache. This makes the analysis scripts that download and use those files also push them up to the S3 cache so they'll be there for the next analysis run. The files don't change (new versions would be released as new files, with a newer year in the filenames), so we don't need to worry about rotating or invalidating the cache.

This is in place of PR #809, which has the advantage of providing a way to load the cache with all the files for all states, but has the disadvantage of adding a lot more code--two Python scripts that are similar but not the same, plus two driver scripts. This achieves the same end as far as what the analysis does--using cached files if they're there and uploading them if they're not--within the existing bash scripts.

Just caching the files after first use will make a big difference to the rate of analysis failures from census.gov errors, but it would leave the door open for the first few runs in any given state to have problems. So there's value in preloading the cache. But not necessarily enough value to justify writing another script to do it. So I took what seemed like the path of least resistance, even though it's an odd road: I ran the scripts from PR #809 (`scripts/run-blocks-cache-update` and `scripts/run-lodes-cache-update`), which saved all the files to my development S3 bucket, then I downloaded them from there and uploaded them to the production bucket. So the production cache is loaded, but if we want to do it again someday we'll have to either check out the branch with those scripts (https://github.com/azavea/pfb-network-connectivity/tree/feature/kak/cache-census-s3%23786) or write a new script to do it.

Resolves #786

## Testing Instructions

- Check that your cache directory is empty: `aws --profile pfb s3 ls s3://USERNAME-pfb-storage-us-east-1/data/` (or at least see what's in it so you'll know what changes)
- Run an analysis
   - You can do a small one, or you can cancel it out once it's done with the jobs and blocks import steps to tying up resources for a long analysis run.
   - If you do a South Dakota or Alaska one, you'll see the "fall back to 2016" behavior for the LODES files.
- Check that the three files used by that analysis are now in your cache directory on S3.
- Run another analysis for the same state (same neighborhood is fine) and confirm that it gets the Census files from S3 rather than downloading them again from the source.

## Checklist

### DIFF
--- a/src/analysis/import/import_jobs.sh
+++ b/src/analysis/import/import_jobs.sh
@@ -45,28 +45,45 @@ function import_job_data() {
 
     NB_STATE_ABBREV="${1}"
     NB_DATA_TYPE="${2:-main}"    # Either 'main' or 'aux'
+
     NB_JOB_FILENAME="${NB_STATE_ABBREV}_od_${NB_DATA_TYPE}_JT00_2017.csv"
+    S3_PATH="s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_JOB_FILENAME}.gz"
 
     if [ -f "/data/${NB_JOB_FILENAME}.gz" ]; then
         JOB_DOWNLOAD="/data/${NB_JOB_FILENAME}.gz"
-    elif [ "${AWS_STORAGE_BUCKET_NAME}" ] && aws s3 ls "s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_JOB_FILENAME}.gz"; then
+        echo "Using local job data file ${JOB_DOWNLOAD}"
+    elif [ "${AWS_STORAGE_BUCKET_NAME}" ] && aws s3 ls "${S3_PATH}"; then
         JOB_DOWNLOAD="${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz"
-        aws s3 cp "s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_JOB_FILENAME}.gz" "${JOB_DOWNLOAD}"
+        aws s3 cp "${S3_PATH}" "${JOB_DOWNLOAD}"
+        echo "Downloaded job data file ${JOB_DOWNLOAD} from S3"
     else
         JOB_DOWNLOAD="${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz"
         set +e
         wget -nv -O "${JOB_DOWNLOAD}" "http://lehd.ces.census.gov/data/lodes/LODES7/${NB_STATE_ABBREV}/od/${NB_JOB_FILENAME}.gz"
         WGET_STATUS=$?
         set -e
+        # If the 2017 file isn't there (SD and AK), do the check/download/cache again for 2016
         if [[ $WGET_STATUS -eq 8 ]]; then
             echo "No 2017 job data available, falling back to 2016 data..."
-            JOB_DOWNLOAD="${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz"
             NB_JOB_FILENAME="${NB_STATE_ABBREV}_od_${NB_DATA_TYPE}_JT00_2016.csv"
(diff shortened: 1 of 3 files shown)


src/analysis/import/import_jobs.sh
```diff
@@ -45,29 +45,44 @@ function import_job_data() {
...
-            wget -nv -O "${JOB_DOWNLOAD}" "http://lehd.ces.census.gov/data/lodes/LODES7/${NB_STATE_ABBREV}/od/${NB_JOB_FILENAME}.gz"
+            S3_PATH="s3://${AWS_STORAGE_BUCKET_NAME}/data/${NB_JOB_FILENAME}.gz"
+            JOB_DOWNLOAD="${NB_TEMPDIR}/${NB_JOB_FILENAME}.gz"
+            if [ "${AWS_STORAGE_BUCKET_NAME}" ] && aws s3 ls "${S3_PATH}"; then
+                aws s3 cp "${S3_PATH}" "${JOB_DOWNLOAD}"
+                echo "Downloaded job data file from S3"
+            else
+                wget -nv -O "${JOB_DOWNLOAD}" "http://lehd.ces.census.gov/data/lodes/LODES7/${NB_STATE_ABBREV}/od/${NB_JOB_FILENAME}.gz"
+                echo "Uploading job data file to S3 cache"
+                aws s3 cp "${JOB_DOWNLOAD}" "${S3_PATH}"
```
ddohler, line 77: I flailed a bit remembering how to run local analyses, which caused me to do a run of this where `AWS_STORAGE_BUCKET_NAME` was an empty string, and I ended up with an error at this point because I hadn't provided it with a bucket to upload to (`S3_PATH` was `s3:///data/...`). The Analysis README mentions that `AWS_STORAGE BUCKET' will be "Ignored unless `PFB_JOB_ID` and `AWS_PROFILE` are set," and this doesn't seem to be quite doing that because it's causing a failure if unset. But I'm not sure how relevant or necessary it is to be able to run analyses without enabling this functionality, or whether in practice anyone ever runs without those two variables set, but I wanted to bring it up in case this needs to be put inside a guard of some kind.

ddohler (approved):
Looks good! I confirmed that the census files made it up to S3 (I had no `/data` folder at all before reviewing this). I had one comment but I'm not sure whether it's relevant and should be an easy fix in any case.

src/analysis/import/import_jobs.sh
KlaasH: Good point. I know PFB runs local analyses, though I'm not sure whether they set `AWS_STORAGE_BUCKET_NAME` or not.  In any case, we definitely don't want to break that use case. I added conditions.

ddohler (approved):
The new guards seem to be working; I was able to run one of the previously broken commands. It's for a bigger geometry, so it hasn't finished yet, but I expect that it will.

PR #813, opened 2020-11-03
Upgrade Django and psycopg2 to match base container

KlaasH:
## Overview

The base container has gotten some patch version upgrades and dependabot is saying we should upgrade Django. Staying in sync with the base container isn't strictly necessary, but seems like a good plan. So this does that, and adjusts the comment in the requirements file to better describe the situation.

This is lieu of PR #803.

## Testing Instructions

- Run `./scripts/update` to rebuild the containers then `./scripts/server` and make sure the app still works.
- That's probably about as far as we need to go for moving this relatively simple Django app up a few patch versions.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### Added
 - S3 caching of Census data files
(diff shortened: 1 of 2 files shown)


ddohler (approved):
Seems to be working fine!

Issue #796, opened 2020-02-20
Assign analysis batch task to logged-in user


KlaasH:
The view that kicks off an analysis batch [always assigns it to the default root user](https://github.com/azavea/pfb-network-connectivity/blob/0.13.0/src/django/pfb_analysis/tasks.py#L42-L43).  This is confusing, and I don't see any reason why it shouldn't be passing in the authenticated user, which should be readily available on the `request`.

The `async_task` function [lets you provide multiple args as a tuple](https://django-q.readthedocs.io/en/latest/tasks.html#django_q.async_task).

PR #814, opened 2020-11-09
Remove filters for created_by, modified_by and fips_code for Analysis Batch

### LINKED ISSUE #795: Don't filter on user or city_fips when looking for neighborhoods to update during batch upload
When processing a batch upload, for each neighborhood contained in the file we check to see if it already exists then update it if so and create it if not.
But the check for whether a neighborhood exists is using fields it shouldn't. See https://github.com/azavea/pfb-network-connectivity/blob/0.13.0/src/django/pfb_analysis/models.py#L394-L412
We're making a dictionary of all the neighborhood attributes, including `city_fips`, `created_by`, and `modified_by` then searching for exactly that. But we should only be using the fields that define a unique neighborhood (see https://github.com/azavea/pfb-network-connectivity/blob/0.13.0/src/django/pfb_analysis/models.py#L395-L412).

This meant that when Rebecca started uploading batch files, using her own account, many of them failed because it searched for an existing neighborhood created by her, didn't find one because the existing instances were created with the Azavea user, then crashed when trying to add them because their name, state, label, and organization conflict with the existing instances.

So we need to change the lookup to use only the four fields that define a unique neighborhood, then either create a new neighborhood or only update the other fields (i.e. `modified_by` and `city_fips`) on the existing one.

kevinearldenny:
## Overview

Ensure that neighborhoods created as part of an AnalysisBatch job are matched with existing neighborhoods, even if they were created by a different user within the organization or if they don't have a fips_code that matches with the existing neighborhood's fips_code. Neighborhoods are now only matched on the fields defined in [unique_together](https://github.com/azavea/pfb-network-connectivity/blob/af9fc26510dfea0ebe806ba5ebc26b952dd55527/src/django/pfb_analysis/models.py#L332). Test cases added to validate AnalysisBatch neighborhood matching logic.

### Demo

N/A

### Notes

I'm running into some issues running `./scripts/test` locally, but I am _hopeful_ that the test cases will work as written. I'm continuing to troubleshoot this on my end - I have attached the output logs of this command here for review in case someone is able to assist in rectifying the issue here.

## Testing Instructions

Path 1 (admin frontend)
 * Reload the servers with 
* `vagrant ssh` then `./scripts/update` and `./scripts/server`
 * Create a batch upload AnalysisJob from the admin frontend while logged in as the systems-pfb@azavea.com user
 * Create another user and log in as the newly created user
 * Create another batch upload AnalysisJob as the newly created user using the same shapefile - there should not be any new neighborhoods created as part of this second AnalysisJob

Path 2 (Django test suite)
* `vagrant ssh` then `./scripts/update`
* Run ./scripts/test

## Checklist

- [ x ] Add entry to CHANGELOG.md

Resolves #795 

[django_test.log](https://github.com/azavea/pfb-network-connectivity/files/5513573/django_test.log)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### Changed
 - Add filter to Neighborhood List
+- Remove filters for created_by, modified_by and fips_code for matching an AnalysisJob to existing Neighborhood records
 
 #### Added
 - S3 caching of Census data files
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Requirements:
 #### Notes for non-Windows users
 
 1. An NFS daemon must be running on the host machine. This should be enabled by default on MacOS. Linux computers may require the installation of an additional package such as nfs-kernel-server on Ubuntu.
+2. For some commands (e.g., `./scripts/test`), you may need to add the ENV variable `PFB_SHARED_FOLDER_TYPE=virtualbox` for the shared folders to work as expected with Django. 
 
 ### Setting up AWS credentials
 
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -391,25 +391,31 @@ class AnalysisBatchManager(models.Manager):
                 label = city
                 name = Neighborhood.name_for_label(label)
 
-                # Get or create neighborhood for feature
+                # Get or create neighborhood for feature using fields in Neighborhood `unique_together` clause
                 neighborhood_dict = {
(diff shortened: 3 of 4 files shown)


KlaasH (approved):
Looks good!  One style edit that I think is worth making in the 4 places the test assertion appears, but good to go after that :+1: 

I got tripped up a bit in testing because I made a whole other org and things behaved strangely. Turned out that was due to #796 (all batch jobs run as the default root user).  But when I tried the scenario that actually revealed this problem in production--making a neighborhood with a user besides the default root user then running a batch that included the same neighborhood--it worked fine and I could see that the `created_by` and `modified_by` fields were different.

PR #817, opened 2020-11-12
Sub-PR to #807: alternate approach to the analysis part of showing default speed limits

KlaasH:
## Overview

This is a sub-PR for #807, replacing the analysis-side parts with a different approach.

My concerns with the upstream approach are:
- it repeats itself, writing an extra copy of the speed limit files that are already on disk and running the logic for looking up the right speed limits from those files multiple times
- it adds analysis logic to the Django app. Which we already sort of do, with the status updates and loading overall scores, but this was doing more, including writing out a results file from the Django app, which won't work in the case of a local analysis run that's not connected to a Django app.
- it's missing the step to read the speed limit and save it on the job instance during import

I wasn't sure how to do a good job of describing what I was picturing instead, nor was I sure it was actually coherent and complete, so I did it in code.

This version of things:
- makes a new table to hold only the speed limit value(s) we looked up for the neighborhood, so we only do the lookup/filter logic once
- exports a CSV directly in the analysis script, so it won't depend on the app being present
- when running an analysis with an actual job ID, runs the management command to load the speed limit onto the instance (like the upstream version does, but based on the new exported CSV rather than the full tables)
- makes the `upload_local_analysis` task look for the file and load the speed limit if it's there (failing quietly if not, so it won't cause older exported analysis runs to crash)

### Notes

- The model changes included a `speed_limit_src` field that I don't think was being filled in. I made the `load_speed_limit` function set that value as well.  It doesn't show up in the front-end, but it seems interesting enough to keep, to me.
- Related to the above, I decided to export both state and local limits rather than setting the table up as just one number and its source.  I'm not sure that's the right way to do it--it does mean there's logic in the app that repeats part of what the analysis is doing--but it seemed simpler to implement (by which I guess I mean "less conditional logic in `bash`")

## Testing Instructions

There are a few different permutations of this thing:
- Running an analysis with a job ID
- Running one without, then zipping up the contents of the results directory and importing it
   - For the import part, I have an example file I put on S3, for Appleton, Wisconsin: `https://s3.amazonaws.com/khoekema-pfb-storage-us-east-1/uploads/appleton_wi_results.zip`
- Importing an old results file (e.g. `https://s3.amazonaws.com/khoekema-pfb-storage-us-east-1/uploads/green_bay_results.zip`), which should work but not show the speed limit box

Connects to #804


src/analysis/import/import_osm.sh
```diff
@@ -207,6 +207,27 @@ else
...
+     -c "CREATE TABLE IF NOT EXISTS \"residential_speed_limit\" (
+            fips_code_state char(2),
+            fips_code_city char(7),
+            state_speed smallint,
+            city_speed smallint
+        );"
+psql -h $NB_POSTGRESQL_HOST -U $NB_POSTGRESQL_USER -d $NB_POSTGRESQL_DB \
+     -c "INSERT INTO \"residential_speed_limit\" (
+            state_fips_code,
+            city_fips_code,
```
kevinearldenny, line 221: These two lines are throwing an error for me currently when trying to run a new analysis - looks like a minor typo. 

Should be `fips_code_state` and `fips_code_city` instead of `state_fips_code` and `city_fips_code`

kevinearldenny (changes requested):
PR is not working for me currently, is throwing an error at the line where I commented. I'm trying to run the analysis again now with the typo fixed and will update with any additional feedback

src/django/pfb_analysis/tasks.py
```diff
@@ -11,6 +11,7 @@
 
 from pfb_analysis.management.commands.import_results_shapefiles import add_results_geoms
 from pfb_analysis.management.commands.load_overall_scores import load_scores
+from pfb_analysis.management.commands.load_residential_speed_limit import load_speed_limit
```
kevinearldenny, line 14: I think you may have forgotten to add `management/load_residential_speed_limit.py` to the commit

kevinearldenny: This is currently failing for me with the following error:

```
Unknown command: 'load_residential_speed_limit'
Type 'manage.py help' for usage.
Updating job status: ERROR Failed See job logs for more details.
```

KlaasH: Alas, so I did.  Just added it.

src/analysis/import/import_osm.sh
KlaasH: I changed my mind about these names at the last minute and apparently updated it incompletely then didn't re-check.  These are the names I wanted to switch to, so I just changed the ones above that I missed.  And ran another analysis, which has gotten past this point.

kevinearldenny:
Okay, so after testing the 3 paths, all are working as expected.

1. New analysis for Boulder, CO - Speed limit is propagated and displayed on Neighborhood Detail map
2. Imported analysis for Appleton, WI - Speed limit is included and displayed on Neighborhood Detail map as expected
3. Imported analysis for Green Bay, WI - Speed limit is not included, so it is not displayed on Neighborhood Detail map

Looks like this is good to go 👍

kevinearldenny (approved):

PR #829, opened 2021-01-27
Replace BNA logo with new 2021 design

### LINKED ISSUE #824: Replace BNA logo
Both logos (the big one on the main page and the smaller one that stays in the upper left) should be replaced with this new one: https://drive.google.com/drive/folders/13760fo4rIv0YyJpoOFiG4r-csnp9iThC

I don't think we have specific direction on sizing, and the proportions are slightly different, but we should be able to do something reasonable and follow up if necessary.

kevinearldenny:
## Overview

Adds the new BNA logo provided in #824 to the frontend, and displays it in the center of the homepage and as a smaller icon in the upper left corner.

### Demo

image

### Notes

The logo in the center of the page looks good, but it doesn't seem like using this same logo in the navbar is going to work. It seems like we might want to split the logo in half and use the left portion only ("people for bikes") as the navbar image, but I imagine that we might want @lederer to check it out and opine before I implement this idea.

## Testing Instructions

 * `./scripts/update`
 * Navigate to the homepage and review the newly added logo

## Checklist

Resolves #824


lederer:
I've made some CSS edits to fix the header logo sizing. Agree that it will look better if we can drop the right half – as well as the "people for bikes" beneath the logo. We should inquire if that's acceptable.

image

image

lederer:
Added the new horizontal header logo

image

image

KlaasH (approved):
:+1: Looks good!

PR #830, opened 2021-01-28
Fix docker login command in `cipublish` script

### LINKED ISSUE #823: Fix docker login command in 'cipublish'
When I ran `cipublish` in the VM in the process of doing a production deploy, it failed with
> unknown shorthand flag: 'e' in -e 
> See 'docker login --help'.

That's coming from [the `docker login` command produced by `aws ecr get-login`](https://github.com/azavea/pfb-network-connectivity/blob/0.14.0/scripts/cipublish#L34), which authenticates with ECR to be able to push containers.

Seemingly the AWS CLI is producing an argument that Docker doesn't recognize.  Given that Docker is fairly up to date (v18) and the AWS CLI is pretty old (it's using the default for the Ansible role, 1.11.14), it seems like the problem is likely on the AWS CLI side.  But when I upgrade it and run `aws ecr get-login` again in the container, it still has the `-e none` option.

So it might make sense to upgrade the AWS CLI to version 2.  I don't know if there are commands that would be affected by [breaking changes between the major versions](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html).  My guess would be no, but I didn't go looking.

UPDATE:
I started working on this in conjunction with issue #827, but bailed because it wasn't turning out to be a "while I'm here..." situation.
There's no Python package for the AWS CLI v2. The only instructions they give for installing it are "download the zip, unzip in, run the `install` script".  I found [a 3rd-party Ansible role](https://github.com/deekayen/ansible-role-awscli2) which seems decent, except it doesn't let you pin the version, just always installs latest. So that's not ideal. But we could make our own role by cribbing the steps from that but changing the URL to pull a variable-specified version.

It's also possible that it's not worth the trouble to upgrade to v2 right now and we should just look for the path of least resistance to getting v1 working again, like possibly munging the output of `aws ecr get-login` to remove the deprecated flag.

kevinearldenny:
## Overview

- Create AWS cli ansible task to install aws cli 2
- Update aws login command in `cipublish` script to use AWS cli v2

## Testing Instructions

 * `./scripts/update`
 * SSH into the vm, then
 * `export GIT_COMMIT="export GIT_COMMIT=$(git rev-parse --short HEAD)"`
 * `export ENVIRONMENT="staging|production"`
 * `export PFB_AWS_ECR_ENDPOINT="950872791630.dkr.ecr.us-east-1.amazonaws.com"`
* `./scripts/cibuild`
* `./scripts/cipublish`

## Checklist

- [ x ] Add entry to CHANGELOG.md

Resolves #823

### DIFF
--- a/deployment/ansible/pfb.yml
+++ b/deployment/ansible/pfb.yml
@@ -8,7 +8,7 @@
 
   roles:
     - { role: "azavea.terraform" }
-    - { role: "pfb.aws-cli" }
+    - { role: "pfb.aws-cli2" }
     - { role: "pfb.env" }
     - { role: "pfb.docker" }
     - { role: "pfb.boto3" }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### Changed
 - Update LODES data url to load 2018 data in analysis
+- Upgrade AWS CLI to v2 and update syntax in `cipublish` script 
+
 
 ## [0.14.0] - 2020-12-21
 
--- a/deployment/ansible/roles/pfb.aws-cli/defaults/main.yml
+++ b/deployment/ansible/roles/pfb.aws-cli/defaults/main.yml
@@ -1,3 +1,0 @@
----
-
-aws_cli_version: 1.18.219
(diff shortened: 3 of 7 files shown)


deployment/ansible/pfb.yml
```diff
@@ -8,7 +8,7 @@
 
   roles:
     - { role: "azavea.terraform" }
-    - { role: "pfb.aws-cli" }
+    - { role: "pfb.aws-cli2" }
```
KlaasH, line 11: Since this PR replaces the `pfb.aws-cli` role, it should include a commit to remove it (probably nice to have that as a separate commit so it would be easy to get it back if we wanted).

KlaasH (changes requested):

KlaasH (approved):
Looks good! :+1:

PR #832, opened 2021-02-16
Handle NULL values in city_fips column in batch shapefiles

### LINKED ISSUE #831: Handle null 'city_fips' values in batch shapefile import
The code that loads neighborhoods from a batch shapefile looks for a `city_fips` property to load. If the property is absent, it gets empty string, which works fine.  It clears the value on an existing neighborhood, which seems correct.
If the property is there but null, though, it gets `None` and tries to save that to the neighborhood, which is not allowed and produces an error from the database.

We should add a check after [the line where the value gets loaded](https://github.com/azavea/pfb-network-connectivity/blob/0.14.0/src/django/pfb_analysis/models.py#L389) to set it to empty string if it comes back as `None`.

KlaasH:
## Overview

In the method that processes batch shapefile uploads, which was updated in PR #814 to look for neighborhoods based on their unique attributes then update their other attributes based on the shapefile, having some neighborhoods with `city_fips` values and some without caused a crash, because the ones without returned `None` as their `city_fips` value, which violates the null constraint of the field.

This adds a check to convert `None` values into empty strings.

Resolves #831

## Testing Instructions

Here's a batch shapefile with one town that has a FIPS code: [WI_and_IA_amesfips.zip](https://github.com/azavea/pfb-network-connectivity/files/5990484/WI_and_IA_amesfips.zip)
The FIPS is 1916900, so if you add a line with that code to your `city_fips_speed.csv` file (download it from `s3://pfb-public-documents/city_fips_speed.csv` if you don't already have one) and run the analysis for Ames, IA, (per the command logged by the batch process), you should see it pick up a city speed limit.
Also, in the admin neighborhood list, the other neighborhoods shouldn't have FIPS codes but Ames, IA, should.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update LODES data url to load 2018 data in analysis
 - Upgrade AWS CLI to v2 and update syntax in `cipublish` script 
 - Update BNA logo with updated 2021 design
+- Add check for null city_fips values in batch upload
 
 ## [0.14.0] - 2020-12-21
 
--- a/src/django/pfb_analysis/models.py
+++ b/src/django/pfb_analysis/models.py
@@ -387,11 +387,15 @@ class AnalysisBatchManager(models.Manager):
                 city = feature['properties']['city']
                 state = feature['properties']['state']
                 city_fips = feature['properties'].get('city_fips', '')
+                # Handle NULL values in the city_fips property, which aren't allowed in the model
+                if city_fips is None:
+                    city_fips = ''
                 osm_extract_url = feature['properties'].get('osm_url', None)
                 label = city
                 name = Neighborhood.name_for_label(label)
 
-                # Get or create neighborhood for feature using fields in Neighborhood `unique_together` clause
+                # Get or create neighborhood for feature using only the fields in the
+                # Neighborhood `unique_together` constraint
                 neighborhood_dict = {
                     'name': name,
                     'label': label,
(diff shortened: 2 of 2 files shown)


kevinearldenny:
@KlaasH I'm working to test this PR now. I have done the following steps:

- Download WI / IA shapefile
- Download `city_fips_speed.csv` and place it into the `data/` folder
- Add a line for Ames, like:
`ames,ia,1916900,32`
- Create new neighborhood
- Create a new analysis job on the frontend

I'm not entirely sure what you mean by:

```
(per the command logged by the batch process)
```

I'm currently trying to run: `NB_OUTPUT_DIR=/data/output/ames-test \
    ./scripts/run-local-analysis \
    PFB_JOB_ID=3fe5516c-5b3c-4039-a4ce-62160a80c45b
    /Users/kevindenny/Downloads/WI_and_IA_amesfips.zip ia 19` but am encountering an error here. Can you expand a bit on the correct instructions to test this?

KlaasH:
Oops, yeah, I skipped a step. Go to http://localhost:9301/#/admin/analysis-jobs/create/batch/ and use that zip file to submit an analysis batch.  The log messages from your `scripts/server` should show a series of `run-local-analysis` commands printed by the `django_q_1` task.  One of them (probably the 3rd) will include `ames.zip` in the boundary file path.  Running that one in the VM should pick up the city speed limit.

kevinearldenny:
The speed limit that I entered for Ames is copied to the analysis result, and Ames is the only neighborhood with a FIPS code - looks good 👍

kevinearldenny (approved):

PR #842, opened 2021-06-04
Increase timeout and decrease retries for Django Q tasks; also fix exception handling

### LINKED ISSUE #841: Big analysis imports are timing out, retrying a lot, and filling up the disk
A few weeks ago and then again yesterday, a few imports failed, then a while later they all started failing.
It turns out we have [Django Q](https://django-q.readthedocs.io/en/latest/configure.html#timeout) set to a 10-minute timeout, and that's not long enough for some imports, so they're getting killed before they finish.  Also, when they get killed, they don't have a chance to clean up their working files, and the default number of retries for a failed task (which is what we're using) is 60.  So once a task fails, it retries every 10 minutes, leaving another copy of its working files stranded in a `/tmp/` directory, until the disk gets full and even small imports stop working.

So we need to raise the timeout and decrease the number of retries.

Questions to answer:
- How long do these things take?
- How long should we set the timeout? Ideally long enough that it'll never get triggered for a job that's still plodding along, but not so long that something can stall out then sit there a long time before it's considered dead.
- How many retries should we do? It's not clear what might cause transient errors, so it's possible that a retry will never succeed. Then again, obviously there are unknowns, so it's probably worth doing at least one or two.
- Is there a way to clean up the files from jobs that were killed by the task runner?  Probably the answer would only be "yes" if it sends some sort of soft termination signal that we could catch and do something with. If the process just gets unceremoniously killed, then cleanup would have to be handled separately.

KlaasH:
## Overview

Big imports are timing out and getting killed by Django Q. Also they leave their working files behind when that happens, so when they then get retried dozens of times, the disk ends up full.

To stop all that from happening, this increases the timeout (from 10 minutes to 3 hours) and sets the number of retries at only 1.  (Side note: I can see where "no retries" might not be a good default for a task processing system and any particular number would seem arbitrary, but ["keep retrying until the end of time"](https://django-q.readthedocs.io/en/latest/configure.html#max-attempts) still doesn't seem like the best choice of default behavior to me.)

I don't know what the upper limit is for how long an import could take. I ran one of the imports that failed in production--Oslo, Norway, using [this boundary](https://github.com/azavea/pfb-network-connectivity/files/6594256/oslo.zip) and the analysis results at `https://production-pfb-storage-us-east-1.s3.amazonaws.com/international_results_2021/europe/osloNO.zip`--and it took just under 20 minutes:
image

Since killing a job that's still actually running would be a bummer, and waiting too long to mark an import unsuccessful if it hangs doesn't seem like that big of a problem, I went with 3 hours as a timeout, since that seems well outside the range of what we would expect even very large imports to take.

#### Exception `.message` error

While looking through the logs to figure out what was going wrong here, I also saw some crashes like this:
image

The reason it's failing in the first place is that it can't open the zip file, but then in our code handling that exception, we were trying to use the `.message` attribute of the Exception, which went away in Python 3 (see https://www.python.org/dev/peps/pep-0352/).
So this fixes the places where we're doing that.

Resolves #841

## Testing Instructions

- Start the VM then run `scripts/server` inside it to start the app.
- Make a neighborhood for Oslo, using the file attached above.
- Import the Oslo analysis results using the newly-created neighborhood and the S3 path above. It should run successfully, with no errors or extra attempts, even though it will probably take more than 10 minutes to finish.  (Your `scripts/server` command should show logs, but you can also do `vagrant ssh -c "cd /vagrant && docker-compose logs -f django-q"` to see them.
- Run another import, with any neighborhood and an invalid "Results Zip File URL" (like `https://www.azavea.com/`, which will load but definitely isn't a zip file of analysis results). It should raise an exception and show the stack trace in the logs, but the second error should be a `LocalAnalysisFetchException`, not an `AttributeError`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Upcoming release]
 
+#### Changed
+- Increase timeout and reduce retries for analysis import tasks
+
 ## [0.15.0] - 2021-02-17
 
 #### Changed
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,7 +58,7 @@ Vagrant.configure("2") do |config|
     ansible.install = true
     ansible.install_mode = "pip"
     # Install pip3 for the system version of python3, and make sure 'pip3' works as well
-    ansible.pip_install_cmd = "curl https://bootstrap.pypa.io/3.5/get-pip.py | sudo python3 && sudo ln -s -f /usr/local/bin/pip /usr/local/bin/pip3"
+    ansible.pip_install_cmd = "curl https://bootstrap.pypa.io/pip/3.5/get-pip.py | sudo python3 && sudo ln -s -f /usr/local/bin/pip /usr/local/bin/pip3"
     ansible.version = "2.8.7"
   end
 
--- a/src/django/pfb_analysis/tasks.py
+++ b/src/django/pfb_analysis/tasks.py
@@ -76,7 +76,7 @@ def upload_local_analysis(local_upload_task_uuid):
         except Exception as ex:
             logging.error('Failed to upload analysis results for task {uuid}'.format(
                 uuid=local_upload_task_uuid))
-            raise LocalAnalysisFetchException(ex.message)
+            raise LocalAnalysisFetchException(ex)
 
     def download_and_extract_local_results(tmpdir, task):
         try:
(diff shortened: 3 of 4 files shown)


KlaasH:
It turns out `max_attempts` isn't going to do anything here. I was looking at https://django-q.readthedocs.io/en/stable/configure.html#max-attempts, but we're running version 1.0.2 of Django Q. For some reason the Read the Docs page has lots of really old versions but nothing between 1.0.0 and stable.  Looking in the repo, it turns out the [v1.0.2 version of that documentation page](https://github.com/Koed00/django-q/blob/v1.0.2/docs/configure.rst) doesn't include `max_attempts`.

So I think all the retrying was due to the weird interaction between their `retry` parameter and `timeout`.  It's [explained in the `retry` section](https://django-q.readthedocs.io/en/stable/configure.html#retry), but basically `timeout` is the time before a task is killed and `retry` is the time before it's resubmitted to the queue.  So if the latter is shorter than the former, a bunch of copies can get added to the queue while the first is still running.  Then once the first finally fails and one of the retry tasks starts, I think new copies of that get added in the same way.  So the cycle only stops, at least in our situation, when the jobs start failing within the `retry` time limit because the disk is full.

Another relevant and interesting difference between v1.0.2 and the current version (v1.3.6) is that `retry` used to default to 60 but now defaults to `None`. So now just setting `timeout` and keeping the default for `retry`, as we did, wouldn't cause this type of cycle.

I think we should probably upgrade Django Q some time soonish, since it seems like it's being actively developed and there might be some rough edges that they're reducing over time. But it probably doesn't make sense to do that with this PR, since what's here fixes the immediate problem.

kevinearldenny (approved):
LGTM 👍 

Analysis files successfully imported for the valid URL, and `LocalAnalysisFetchException` error thrown for the invalid URL

PR #856, opened 2021-11-24
Update base VM to fix ansible installation error

KlaasH:
## Overview

VM provisioning has started failing, with a message about Python 3.5 having reached end-of-life. There might be other ways to get around it, but upgrading the base VM would be a good thing anyway, so that's the first thing I tried.

It seems like it worked--I'm able to provision the app and run the analysis. I also updated the Ansible roles for good measure. We only use the VM as a standard environment in which to build and run the containers, so it's not that surprising that it wouldn't be too susceptible to issues from upgrading.

### Notes

There's a chance this will affect deployment, since we do deployments from within the VM.  But the crash is currently blocking Rebecca from running the analysis locally, and I'm not sure it's worth pursuing the deployment question at the moment, when there are no imminent plans to deploy.

## Testing Instructions

- **Update**: Copy the changes from `deployment/ansible/group_vars/all.example` into your `deployment/ansible/group_vars/all` file.
- Provision the project per the README
- If it gets as far as starting the app, that means you're definitely past the point where it was crashing. For a more complete test, you could go through adding a neighborhood and starting an analysis job for it.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### Changed
 - Upgrade AWS CLI to v2 in analysis container
+- Upgrade base VM from Ubuntu 16.04 to 20.04
 
 ## [0.15.1] - 2021-06-04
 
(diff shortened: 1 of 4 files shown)


KlaasH:
I updated the testing instructions above because I changed a couple versions in the `group_vars/all.example` file, which need to be manually copied to the local `all` file.

ktohalloran (approved):
It runs without crashing 🎉  I also ran a local analysis which also worked so it should be ready to go.

Issue #818, opened 2020-11-20
Account creation and password reset emails not sending


KlaasH:
When I submit the password reset form (https://bna.peopleforbikes.org/#/password-reset-request/), it says "Password reset successfully requested", but that's not so.  The request is actually coming back with a 500 error, and this shows up in Papertrail:
![image](https://user-images.githubusercontent.com/6598836/99814592-2d594780-2b17-11eb-9f33-e2f9e39453f3.png)

This seems likely to be fallout from the Django upgrade, though it could also be just a change in the package itself.  In any case, we need to figure out how to properly import it.

We also need to fix the front-end so that it shows an error when there's an error, rather than declaring success regardless.

**Workaround**
I don't think there's a quick/easy way to get the emails to send--even if the fix is a one-line change to settings, as seems likely, it will require a redeploy to make it take effect.  But it's possible to generate a password reset link manually.
- Get a Django shell in the app, i.e.:
  - Load the production SSH key
  - `ssh -A ec2-user@bastion.bna.peopleforbikes.org`
  - Figure out the internal IP of an instance that's running an app task (look in the AWS console), and SSH to that
  - Run `docker ps` and get the container ID of an app task (the image will include `pfb-app` and the command starts with `/usr/local/bin/gunicorn`)
  - Run `docker exec -ti CONTAINER_ID bash` to get a shell in that container
  - Run `./manage.py shell_plus` to get a Django shell
- Run this (replacing the EMAIL_ADDRESS... bit with the desired email address):
  ```
  from django.conf import settings
  from django.core.signing import TimestampSigner
  from users.models import PFBUser

  user = PFBUser.objects.get(email="EMAIL_ADDRESS_OF_USER_TO_RESET")
  signer = TimestampSigner(salt=settings.RESET_SALT)
  token = signer.sign('{}'.format(user.uuid))
  print('https://bna.peopleforbikes.org/#/password-reset/?token={}'.format(token))
  ```
- Copy the resulting URL. It should work to reset that accounts password. It will expire after 24 hours.

PR #870, opened 2022-03-17
Fix email sending

aaronxsu:
## Overview

This PR updates the `EMAIL_BACKEND` in `settings.py` so that it matches the suggested settings in  [`django-amazon-ses v2.1.1`](https://github.com/azavea/django-amazon-ses/tree/2.1.1#django-configuration), since the trackback in the original issue seems to suggest that it ran into module import issue of this dependency, and that this module change in this dep happened in the upgrade of 0.3.2 -> 2.1.1.

This PR also removes `ipython` installation from `deployment/aws-batch/Dockerfile` and `src/verifier/Dockerfile`, and it pins the base image python version in `deployment/aws-batch/Dockerfile`.

## Testing Instructions

 * Push a `test/` branch up to trigger a staging deploy
 * Once the ECS tasks are cycled, go to https://staging.pfb.azavea.com/#/login/ , click on "Reset password", put in the email address that is stored under `PFB Staging Root User` from lastpass
 * Check if there is a password reset email send to that email address

## Checklist

Resolves https://github.com/azavea/pfb-network-connectivity/issues/818

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Upgrade base VM from Ubuntu 16.04 to 20.04
 - Update django-q from 1.0.2 to 1.3.6
 
+#### Fixed
+- Fixe email sending
+
 ## [0.15.1] - 2021-06-04
 
 #### Changed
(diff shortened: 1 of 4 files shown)


KlaasH (approved):
:+1: Works great!

PR #871, opened 2022-03-18
Add State & FIPS cols to admin analysis jobs table (#865)

### LINKED ISSUE #865: Add State & FIPS code columns to admin analysis jobs table
We want to add two columns to the Analysis Jobs table to make it easier to navigate:
- Add a state name column so that it is easier to differentiate between cities with the same name.
- Add a FIPS code column so that cities can be more easily tracked over time and align better with other sources of data

Both should be filterable, like the existing Status, Neighborhood, and Batch ID columns.

jacobscarfmerrell:
## Overview

Adding the state name makes it easier to distinguish between cities with
the same name. Adding FIPS makes it easier to track cities over time and
stay aligned with other data sources. Both fields are filterable.

## Testing Instructions

 * Start the server
 * Go to http://localhost:9301/admin#/admin/analysis-jobs/
 * Should be searchable by FIPS or state abbreviation

## Checklist

Resolves #865

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Upgrade AWS CLI to v2 in analysis container
 - Upgrade base VM from Ubuntu 16.04 to 20.04
 - Update django-q from 1.0.2 to 1.3.6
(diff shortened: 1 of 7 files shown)


KlaasH (changes requested):
This looks good! A few tweaks and cleanup notes below.

KlaasH (approved):
:+1: Looks good!

PR #874, opened 2022-03-30
Revert "Update django-q to fix migration error"

### LINKED ISSUE #808: Missing migrations

lydiascarf:
This reverts commit b187474e87733ea506e26929fed569f2fc97b0e2.

## Overview

The original upgrade was to resolve #808, but what I didn't realize at the time was that upgrading django-q currently breaks the local analysis functionality.


KlaasH (approved):
Yeah, this is a bummer, and we need to address the chain of outdated components that are standing in the way of updating this one, but for now this makes sense. :+1:

PR #878, opened 2022-04-01
Make max trip distance configurable by UI (#866)

### LINKED ISSUE #866: Add ability to change analysis max trip distance
PFB has been getting requests to change the trip distance defaults in the BNA. Rather than having to rely on
running a local analysis to change the NB_MAX_TRIP_DISTANCE and NB_BOUNDARY_BUFFER parameters, they
would like to able able to change this input from the tool’s admin UI, similar to the way the “Run Analysis” page
accepts an OSM file via a URL. To accomplish this, we'll need to add fields to the job model, the form for creating jobs,
and the batch job setup.

If no specific value is given for `NB_BOUNDARY_BUFFER`, it gets set within the analysis scripts to be equal to `NB_MAX_TRIP_DISTANCE`.  I don't think there's any need to adjust it separately, so we can just rely on that behavior to keep the two variables in sync.

`NB_MAX_TRIP_DISTANCE` is defined in meters, so the field should make that clear and should only accept positive integer values. The default is 2680.

jacobscarfmerrell:
## Overview

Client wanted to run interactive and batch analyses using
the angularjs client without having to use the default max
trip distance value of 2680 meters. This was previously only
possible by setting the NB_MAX_TRIP_DISTANCE environment
variable before running `scripts/run-local-analysis` locally.

### Demo

Screenshot from 2022-04-01 15-59-17
Screenshot from 2022-04-01 15-59-25

## Testing Instructions

 * `./scripts/setup`
 * in another tab `./scripts/server`
 * go to http://localhost:9301/admin#/admin/analysis-jobs/create/batch/ 
 * give it a shapefile (feel free to slack me for one) and a trip distance
 * check that the logged commands have the NB_MAX_TRIP_DISTANCE you set
 * check that the analysis jobs in the UI have the correct max trip distance
 * go to http://localhost:9301/admin#/admin/analysis-jobs/create/
 * pick a neighborhood and set a different trip distance from before
 * check that the logged command has the new NB_MAX_TRIP_DISTANCE you set
 *  check that the analysis job in the UI has the new correct max trip distance

## Checklist

Resolves #866


KlaasH (approved):
Looks good!  The individual and batch creation process worked as expected, and I ran jobs for two versions of the same place with different distances to confirm that the calculations were affected by the change.  One note (in two places) below re the field label, but with that tweak this is good to go :+1:

ktohalloran (approved):
Looks great, thanks for inviting me to review!

KlaasH (changes requested):
I forgot that we had tests for `create_from_shapefile` in the Django app. (I also assumed the CI failure was just from squirrelliness rather than looking to see what the error was. Oops. Though to be fair the first build failure was just a transient 500 error from Quay.io.)

Looking at your fixup for the tests made me think it would be nice to have the `max_trip_distance` param be optional for `create_from_shapefile` rather than having to pass `None` when we don't want an override.

Also, `None` gets a bit complicated, since the field is nullable, so if we end up passing `max_trip_distance=None` to `AnalysisJob.objects.create()` it will respect that and leave the field blank rather than using the default.  Which isn't the end of the world, since the analysis itself will fill in the default, but I think it makes sense to save it on the job.

I also noticed, in testing to see how the None case would actually show up on the jobs, that the Run Batch form submission would crash if Max Trip Distance wasn't set.

I ended up writing out the changes to make sure what I was envisioning would work both with and without a `max_trip_distance` value and would allow the tests to run without modification.  Also, having been reminded that we have those tests, I went ahead and added a couple to check this behavior.

Here's the commit: https://github.com/azavea/pfb-network-connectivity/commit/73075704060381951b9c448d4bb01212b7ede461

Note: It's based on the first commit in this branch, so if you want to `git cherry-pick` it in, you'll also want to drop the fixup commit to avoid conflicts.

PR #879, opened 2022-04-08
Destination types with no locations are N/A (#867)

### LINKED ISSUE #867: Remove destination types with no locations
On the detail page, some places show destination types with no locations (e.g., Community Centers, 0). Since this can be a point of confusion for users, PFB would like to remove the destination types without locations for a particular place from the left panel.

The contract description says "by preventing those scores from being written to the results file", but that would mean the analysis jobs that were just run without this change in place would keep showing types with no locations until they were replaced next year.  So if it's feasible to make the distinction on the front-end, it seems preferable to suppress destination types with no value there.  If not, changing the analysis export will do (plus making sure that the front-end is fine with some fields being missing/different analysis jobs having different lists of destination types).

jacobscarfmerrell:
## Overview

In the places detail and compare pages, there has been no way to
differentiate between destination types without instances in the given
place and destination types with instances that are inaccessible by low
stress routes. The former are now marked N/A; the latter are marked 0.

### Demo

detailscreen
comparescreen

## Testing Instructions

 * `./scripts/setup`
 * `vagrant ssh`
 * `./scripts/server`
 * if you don't have any places yet when you go to [all places](http://localhost:9301/admin#/places////)
   * go to http://localhost:9301/admin#/admin/analysis-jobs/create/batch/ 
   * give it a shapefile (feel free to slack me for one)
* go to http://localhost:9301/admin#/places//// and select a place
* for any metric that shows N/A, the associated geojson in your s3 bucket should have 0 features
* for any metric that shows 0, the associated geojson in your s3 bucket should have 1+ features
* the above two bullets should also be true if you go to http://localhost:9301/admin#/places//// then compare two places

## Checklist

Resolves #867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Upgrade base VM from Ubuntu 16.04 to 20.04
 - Add State Abbreviation, City FIPS to Admin Analysis Jobs table
 - Add Max Trip Distance (meters) to Admin Analysis Jobs forms
+- Make Places service show N/A for destination types w/o locations
 
 #### Fixed
 - Fix email sending
--- a/src/angularjs/src/app/components/places.service.js
+++ b/src/angularjs/src/app/components/places.service.js
@@ -9,13 +9,26 @@
     'use strict';
 
     /* @ngInject */
-    function Places($q, $log, Neighborhood, AnalysisJob) {
+    function Places($q, $http, $log, Neighborhood, AnalysisJob) {
 
         var module = {
             getPlace: getPlace
         };
 
         return module;
+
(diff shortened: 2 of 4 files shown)


KlaasH (changes requested):
This looks good, and it's nice that we can do it in the frontend and don't need to change the analysis itself or wait for next year's run to see the resulting change.

In addition to the Menomonee Falls example, which has some good N/As but doesn't actually have any true zeroes, I grabbed the boundary and ran an analysis for [Pelham Manor, NY](https://bna.peopleforbikes.org/#/places/e46046fe-d943-4e6b-974f-8ece05eafc12/), which has a lot of zeroes, some of which are N/As but several of which are true zeroes.

There are a couple things noted below that need to be fixed up, but with those I think it'll be in good shape.

KlaasH (approved):
:+1: Looks good!

PR #883, opened 2022-04-26
Drop unmaintained angular-file-saver

BryanQuigley:
This also broke the build.
I grepped for SaveAs and couldn't find it used anywhere.


BryanQuigley:
Not sure how to test this as I can't find it being used.

KlaasH (approved):
Yeah, I don't know if it was ever used, but it certainly isn't now.  It might have come in with some code that was copied/adapted from elsewhere and the part that used it got dropped.  Anyway, yeah, we don't need it :+1:

PR #887, opened 2022-05-06
Adjust vagrant params and analysis job environment params to fix errors found while testing

KlaasH:
## Overview

Two small fixes for issues I ran into while testing things:
- some of the parameters in the Vagrantfile stopped working when I upgraded to Virtualbox 6
- the analysis job submission wasn't working on staging because the boto client (or the AWS API? comes to the same thing) wants the values in the `environment` parameter to be strings, not numbers.

See the commit messages for a little bit more detail.

## Testing Instructions

I'm not sure how much testing this warrants, but it's worth checking that your VM comes up fine and that you can at least kick off an analysis job.  Testing it again on staging (I didn't test the changed code, but I did test that casting the param to a string made the job submission work in `shell_plus` on the staging instance) probably isn't worth the trouble.

## Checklist

- [ ] ~Add entry to CHANGELOG.md~ (too small)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,7 +8,7 @@ PFB_SHARED_FOLDER_TYPE = ENV.fetch("PFB_SHARED_FOLDER_TYPE", "nfs")
 
 if PFB_SHARED_FOLDER_TYPE == "nfs"
   if Vagrant::Util::Platform.linux? then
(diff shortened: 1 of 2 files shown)


jacobscarfmerrell:
fixed the vpn, everything's working 100% now

PR #888, opened 2022-05-09
Drop rds parameter groups as part of db upgrade

### LINKED ISSUE #859: Upgrade Postgresql in development and staging
AWS is end-of-life-ing Postgresql 9.6, and will forcibly upgrade us to version 12 after ~January 18th~ April 26th.

Further details: https://forums.aws.amazon.com/ann.jspa?annID=8499

BryanQuigley:
The way we specify db parameter group doesn't work well for upgrades.  Removing it is much simpler.

### Notes

Happy to update and remove terraform-rds entirely as I've done for other projects - this may help with eventually upgrading terraform.

## Testing Instructions

 * Confirm stg still works
 * Review parameter group values

Resolves #859

### DIFF
--- a/deployment/terraform/database.tf
+++ b/deployment/terraform/database.tf
@@ -10,48 +10,6 @@ resource "aws_db_subnet_group" "default" {
   }
 }
 
-resource "aws_db_parameter_group" "default" {
-  name        = "${var.rds_database_identifier}"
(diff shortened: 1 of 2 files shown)


KlaasH (approved):
:+1:

Issue #760, opened 2019-09-12
Update docker and terraform versions


KlaasH:
Per @flibbertigibbet on PR #759:
> I noticed some deployment dependencies and roles could use updating, in group_vars/all.example (docker_compose_version and terraform_version, and probably docker_version as well, it's not clear to me what version ECS should be on from [this](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html)) and in roles.yml (terraform and docker).

The `docker_version` should match the version in the AMI we're using, which means I should have changed it when I changed the AMI for [issue #579](https://github.com/azavea/pfb-network-connectivity/issues/579).  I doubt there's a whole lot of risk arising from having them not match, but it makes sense to align the development and deployed environments as closely as we can.

BryanQuigley:
The docker part of this was fixed as part of #856.

PR #896, opened 2022-06-30
Feature/bryan/terraform upgrade

BryanQuigley:
## Overview

Upgrading terraform let's us move to Github Actions, use modern AWS providers/examples, and makes it generally easier to make changes.

### Notes

Was hoping to improve terraform more then just required for upgrading, but it looks like that will have many other dependencies - FARGATE/docker version.

Note, every upgrade did recreate the container instances causing small bit of downtime :(
I had a few theories as to why - not enough options specified in the config or the use of md5 to shorten names.. but none panned out to fix it.

## Testing Instructions

 * Confirm I didn't break staging or at least one other dev environment.

## Checklist

Half of #760

### DIFF
--- a/scripts/infra
+++ b/scripts/infra
@@ -69,7 +69,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                 docker-compose run --rm \
                     terraform plan \
                           -var-file="${PFB_SETTINGS_BUCKET}.tfvars" \
-                          -var="git_commit=\"${GIT_COMMIT}\"" \
+                          -var="git_commit=${GIT_COMMIT}" \
                           -var="batch_analysis_job_definition_name_revision=${BATCH_ANALYSIS_JOB_NAME_REVISION}" \
                           -out="${PFB_SETTINGS_BUCKET}.tfplan"
                 ;;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Replace Feedback link with City Ratings link, point Methodology link to City Ratings methodology page
 - Facilitate non-US analyses with custom population files, jobs files, and osm extracts
 - Make analysis convert kmph speed limits to rounded mph
+- Upgrade Terraform to 1.1.9
 
 #### Fixed
 - Fix email sending
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -12,6 +12,3 @@ pfb_aws_batch_analysis_job_definition_name: "staging-pfb-analysis-run-job"
 docker_version: 5:19.*
 docker_options: "--storage-opt dm.basesize=20G"
 docker_compose_version: 1.23.*
-
-# azavea.terraform
-terraform_version: 0.9.11
--- a/deployment/ansible/pfb.yml
+++ b/deployment/ansible/pfb.yml
@@ -7,7 +7,6 @@
       apt: update_cache=yes
 
   roles:
-    - { role: "azavea.terraform" }
     - { role: "pfb.aws-cli" }
     - { role: "pfb.env" }
     - { role: "pfb.docker" }
--- a/deployment/ansible/roles.yml
+++ b/deployment/ansible/roles.yml
@@ -1,4 +1,2 @@
 - src: azavea.docker
   version: 6.0.0
-- src: azavea.terraform
-  version: 0.4.0
--- a/deployment/terraform/.gitconfig
+++ b/deployment/terraform/.gitconfig
@@ -1,0 +1,5 @@
+[safe]
+        directory = /usr/local/src/.terraform/modules/database
+        directory = /usr/local/src/.terraform/modules/vpc
+
+
--- a/deployment/terraform/alarms.tf
+++ b/deployment/terraform/alarms.tf
@@ -18,11 +18,11 @@ resource "aws_cloudwatch_metric_alarm" "ecs_cpu_util" {
   statistic           = "Average"
   threshold           = "75"
 
-  dimensions {
-    ClusterName = "${aws_ecs_cluster.app_container_instance.name}"
+  dimensions = {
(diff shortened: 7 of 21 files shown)


KlaasH (commented):
I read through the changes and they seem to make sense.  I also spun up a fresh VM and tried to run through the deployment process inside it, and that got pretty far (though for some reason the analysis and nginx containers didn't build on the first try. But they got there after some retries), but `infra plan` is crashing with this:
> │ Error: Failed to download module
│ 
│ Could not download module "vpc" (network.tf:2) source code from
│ "git::https://github.com/azavea/terraform-aws-vpc.git?ref=6.0.1": error downloading
│ 'https://github.com/azavea/terraform-aws-vpc.git?ref=6.0.1': /usr/bin/git exited with 128: fatal: unsafe repository
│ ('/usr/local/src/.terraform/modules/vpc' is owned by someone else)
│ To add an exception for this directory, call:
│ 
│       git config --global --add safe.directory /usr/local/src/.terraform/modules/vpc


Re "Note, every upgrade did recreate the container instances causing small bit of downtime :(", are you running the whole process (build, publish, plan, apply) for each commit, or just plan and apply?  If we're telling the tasks to use different container tags, naturally they'll want to restart for every update, but since the containers don't change at all, it should work to set `GIT_COMMIT` to the final commit hash and keep it that way the whole time.  But maybe that's what you're already doing and the tasks are getting replaced anyway.

BryanQuigley:
Just plan and apply is all I've been doing with the same GIT_COMMIT as is on staging.  It wants to recreate them all anyway - generally trying to set certain AWS defaults back to null, but even when I tried adding them - it still wanted to recreate them.
Example:
`- portMappings     = [] -> null`

You may have to delete any terraform folders or files in deployment/terraform if their permissions don't line up.

BryanQuigley:
Maybe it is just a permissions error - afaict I did not change any local or in-project git config. I did run into that error though.

KlaasH:
> Maybe it is just a permissions error - afaict I did not change any local or in-project git config. I did run into that error though.

How did you get around it?  I didn't run `plan` on intermediate versions, so I don't know if it's only on the 1.1.9 container.  But yeah, this is an unmodified container, so I don't see how it can be a local issue.

BryanQuigley:
Yes, that's it - no changes required now. So for production upgrade I would need to 
1. Comment out update_batch_definitions
2. export production equivalent of BATCH_ANALYSIS_JOB_NAME_REVISION="staging-pfb-analysis-run-job:761"

Then we should be at no (or minimal) downtime IIUC.

__
The last change I pushed to this branch fixes the git permissions/safety issue by having the terraform container run as whoever the local user is.

BryanQuigley:
It was echo locator where I saw the permission/safe directory issue before.  In that project I worked around it via adding:
git config --global --add safe.directory /usr/local/src to the infra script itself.

scripts/infra
```diff
@@ -53,6 +53,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/terraform/terraform.tfvars" "${PFB_SETTINGS_BUCKET}.tfvars"
 
+        # Workaround Git wanting everyone to be the same user
+        CURRENT_UID=$(id -u):$(id -g)
```
KlaasH, line 57: From [this comment](https://github.com/azavea/pfb-network-connectivity/pull/896#issuecomment-1181783109):
> It was echo locator where I saw the permission/safe directory issue before. In that project I worked around it via adding:
git config --global --add safe.directory /usr/local/src to the infra script itself.

Do we want to do that and back out this workaround?  Or are we fine with the CURRENT_USER approach?

BryanQuigley: Backing it out and switching to git config.

scripts/infra
```diff
@@ -53,6 +53,11 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         aws s3 cp "s3://${PFB_SETTINGS_BUCKET}/terraform/terraform.tfvars" "${PFB_SETTINGS_BUCKET}.tfvars"
 
+        # Workaround Git wanting everyone to be the same user
+        docker-compose run --rm \
+                    --entrypoint git \
+                    terraform config --global --add safe.directory /usr/local/src
```
KlaasH, line 59: This is running with `docker-compose run --rm` and not writing to a persistent volume, so I think it sets that git config then immediately throws away the modified container, and the subsequent commands each run in a container of their own that doesn't have the config.

But actually, when I manually ran a Terraform container (`docker-compose run --entrypoint bash terraform`) and ran each of the commands in turn within the same container:
```
git config --global --add safe.directory /usr/local/src
rm -rf .terraform/ terraform.tfstate*
terraform init -backend-config="bucket=${PFB_SETTINGS_BUCKET}" -backend-config="key=terraform/state"
```
it still failed with the same error:
image

BUT If I do that then run the two recommended commands to whitelist the specific directories
```
git config --global --add safe.directory /usr/local/src/.terraform/modules/database
git config --global --add safe.directory /usr/local/src/.terraform/modules/vpc
```
then it works.

So yeah, I don't know how to get that git config to be available to the container that needs it.  It could be done in the Dockerfile, but that's way upstream. Also if the general `/usr/local/src` setting doesn't work and it needs to be specific to the modules, that would not be a suitable addition to the Dockerfile anyway.  I don't remember the nuances of docker-compose "volume" config well enough to know if there's a straightforward way to make that one file persist across containers.

Possibly we should go back to the user workaround...

BryanQuigley: Went to new style mounting a gitconfig file to /root/.gitconfig workaround.

KlaasH (approved):
Ok, I think this is good to go now.  The 3rd workaround (writing a `.gitconfig` that will make Terraform happy and linking it into the container) is working well for me, and I like it better than the workaround that manipulates the user or the option of always running `infra` on host.

And yeah, hopefully with overriding `BATCH_ANALYSIS_JOB_NAME_REVISION` to make it match the deployed version, this can be applied with minimal disruption.

:+1:

Issue #891, opened 2022-06-08
Enable providing custom jobs data or skipping jobs data in the analysis


KlaasH:
Similar to issue #890, the way we get jobs data for US cities doesn't translate to other countries.  We want to add a parameter whereby we can provide the analysis with a URL to download the jobs data from.  As with #890, the assumption is that the file will have been formatted to match what the analysis expects, so it shouldn't require munging or any changes to the analysis logic beyond the retrieval and extraction steps.  We'll want to get sample files from Rebecca for testing.

Unlike #890, we also want to provide an option, for places where jobs data just isn't available, to omit it altogether.

Issue #890, opened 2022-06-08
Enable providing a custom population file to the analysis


KlaasH:
For non-US cities, the population information the analysis gets from the Census isn't available.  Rebecca has been manually creating population files in a compatible format and overriding the analysis (i.e. by modifying the code and running it locally) to use them.  We want to add a parameter for providing them to the analysis.  It should work similarly to the "OSM extract URL" parameter, where we don't upload the file when submitting the job, we just provide an URL where it can be found (like a public URL for a file on S3).

The intention, at least at this point, is for the input files to already be formatted in the way the analysis expects, i.e. this will change where/how the analysis gets them but not how it processes them.  We'll need to get some sample population files from Rebecca for testing.

PR #897, opened 2022-07-05
Facilitate UI-driven analyses for non-US places (#889) (#890) (#891)

### LINKED ISSUE #614: Support running international jobs

### LINKED ISSUE #889: Enable fetching non-US OSM extracts
- Adjust the analysis parameters to support specifying a country (by ISO alpha-3 country code) rather than a US state.
- Adjust the code that downloads, caches, and extracts the OSM extract to get the appropriate country extract. Some countries have state-level extracts, but hopefully we can get away with just using the whole country.

The rest of the analysis won't necessarily work (like it'll crash looking for jobs data until that task is done) but this should be able to get to the point of having the `neighborhood_ways` and other OSM-derived tables imported and clipped to the boundary.

jacobscarfmerrell:
## Overview

The client needs to be able to run analyses using non-US places. To run a successful analysis for a non-US place, a custom population file and jobs file will need to be provided to the analysis script. An OSM extract may be needed as well if the file cannot be found at the geofabrik URL, which we derive the country's alpha3 code. This PR parameterizes the analysis script to fulfill those needs, then extends those parameters to the UI through the "Run Analysis" button in the Analysis Jobs admin page.

### Demo

Screenshot from 2022-07-05 13-46-20

### Notes

We had to increase the allowable geographic width of a shapefile in detect_utm_zone.

## Testing Instructions

* make sure you have a few hours in which you can let this run
* run `scripts/setup`, `vagrant ssh`, `scripts/server`
* navigate to http://localhost:9301/admin#/admin/neighborhoods/create/
* create a neighborhood for Brussels using this [zip](https://drive.google.com/file/d/1Ge9dvSS9kjWKg5C5sNt18fZuN1NYcxBI/view?usp=sharing)
* navigate to http://localhost:9301/admin#/admin/analysis-jobs/create/
* create an analysis using
  * a link to [this population file](https://drive.google.com/file/d/1rwQtUgfTyBKpBKGhVpKI8_fYLVFA-B-X/view?usp=sharing)
    * this will require downloading the file, uploading it to your s3 bucket, then dropping that link in
    * or you can just use [mine](https://jacob-pfb-storage-us-east-1.s3.amazonaws.com/data/tabblock2010_98_pophu.zip)
  * and a link to [this jobs file](https://drive.google.com/file/d/1EkNW-zrViUGTjIvVHMcTJQDu3aB2hlS1/view?usp=sharing)
    * same as above
    * or you can just use [mine](https://jacob-pfb-storage-us-east-1.s3.amazonaws.com/data/be_od_main_JT00_2017.csv.gz)
  * you don't need to set the OSM extract url for belgium because the analysis can find it on geofabrik but you can!
    * use the [osm extract url from geofabrik](https://download.geofabrik.de/europe/belgium-latest.osm.pbf)
* nearly half a work day later, you should see `Updating job status: COMPLETE Finished analysis` and some scores
* for good measure, navigate to http://localhost:9301/admin#/places//// and click on your neighborhood, then verify that the scores came through (i.e. it's not all 0s)

## Checklist

- [x] Tested without OSM file set
- [x] Tested with OSM file set
- [x] Tested with jobs skipped

Resolves #614, #889, #890, #891


src/analysis/import/import_neighborhood.sh
```diff
@@ -76,58 +83,79 @@ then
...
         elif [ "${AWS_STORAGE_BUCKET_NAME}" ] && aws s3 ls "${S3_PATH}"; then
             echo "Using census blocks file from S3"
             BLOCK_DOWNLOAD="${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.zip"
             aws s3 cp "${S3_PATH}" "${BLOCK_DOWNLOAD}"
         else
+            echo "Using census blocks file from official census site"
             BLOCK_DOWNLOAD="${NB_TEMPDIR}/${NB_BLOCK_FILENAME}.zip"
-            wget -nv -O "${BLOCK_DOWNLOAD}" "http://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/${NB_BLOCK_FILENAME}.zip"
+            PFB_POP_URL="http://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/${NB_BLOCK_FILENAME}.zip"
+            wget -nv -O "${BLOCK_DOWNLOAD}" "${PFB_POP_URL}"
```
KlaasH, line 145: For a US town with no population file URL provided, I'm ending up here but it's doing this:
> Using census blocks file from official census site
https://www2.census.gov/geo/tiger/TIGER2010BLKPOPHU/population.zip:
2022-07-22 13:50:50 ERROR 404: Not Found.

So seemingly `NB_STATE_FIPS` isn't set, but I'm not sure why not.

KlaasH (changes requested):
I have more to look at here, but I
- got partway through the code
- tried to run a full Brussels job but it crashed. Abruptly. It might have run out of memory.
- made a tiny boundary for an imaginary town inside the Brussels boundary and ran that. It succeeded :tada: 
- ran an analysis for a small US town and got a crash

KlaasH (commented):
Getting closer to having reviewed all the code, but since I have to pause for a bit, I'm once again submitting what I have so far.

src/analysis/import/import_jobs.sh
```diff
@@ -34,19 +34,53 @@ NB_POSTGRESQL_PASSWORD - Default: gis
...
     chmod -R 775 "${ROOT_TEMPDIR}"
 
-    NB_STATE_ABBREV="${1}"
-    NB_DATA_TYPE="${2:-main}"    # Either 'main' or 'aux'
-
-    NB_JOB_FILENAME="${NB_STATE_ABBREV}_od_${NB_DATA_TYPE}_JT00_2018.csv"
+    PFB_COUNTRY="${1}"
+    PFB_STATE="${2}"
+    if [ -n "${PFB_JOB_URL}" ]; then
+        NB_JOB_FILENAME=$(basename "${PFB_JOB_URL}")
```
KlaasH, line 79: If `PFB_JOB_URL` is already set, we should be downloading the file from that URL.  This is using the URL to get the filename, then looking for the filename in the `/data/` directory of the configured storage bucket.  But we're asking for a whole URL, not just a filename, so it doesn't make sense to reconfigure it like that.  At best it will happen to match, but generally (like if Rebecca is using a file in a bucket she made, or if I'm using a file in your bucket) the file won't be found at the new S3 path.

Also, currently `basename "https://jacob-pfb-storage-us-east-1.s3.amazonaws.com/data/be_od_main_JT00_2017.csv.gz"` returns `be_od_main_JT00_2017.csv.gz`, then we add another `.gz` below when we set `S3_PATH`.  It looks like we could add an argument to get this to also strip the `.gz` extension if it's there:
```suggestion
        NB_JOB_FILENAME=$(basename "${PFB_JOB_URL}" .gz)
```

I guess we do need to parse out the filename, since we need to point the psql `\copy` command at it further down, but I think it probably doesn't make sense to treat it as meaningful the way we do for the US files.

KlaasH (changes requested):
I've successfully run small and medium-sized jobs in Belgium now, and my "a town in the US still works fine" test also succeeded.  Except actually by "successfully ran" I mean they finished and showed results, but for the Brussels ones I noticed the job scores were zero, which turns out to because the jobs import isn't finding the file I tried to point it at.

KlaasH (approved):
:+1: I think this is good!  I'm getting results that look good for Brussels with or without jobs, and the US results seem to still be working fine.

PR #902, opened 2022-08-18
Fix import script

rgreinho:
The image is built from `azavea/pfb-network-connectivity@0037a5c`, and the command I ran was the following:

```bash
docker run --rm \
    -e PFB_SHPFILE="/data/metz-lorraine.shp" \
    -e PFB_OSM_FILE="/data/metz-lorraine.osm" \
    -e PFB_COUNTRY=nonus \
    -e PFB_STATE=zz \
    -e PFB_STATE_FIPS=0 \
    -e NB_OUTPUT_DIR=/data \
    -e RUN_IMPORT_JOBS=0 \
    -e PFB_DEBUG=1 \
    -v "PeopleForBikes/brokenspoke-analyzer/data":/data \
    azavea/pfb-network-connectivity:0.17
```

The goal is to run an analysis for the city of Metz in France, located in the Lorraine region.

In `analysis/import/import_neighborhood.sh` the block attempting to do some cleanup is erroneous and can be removed.

If left, it crashes with the following error:

```bash
+ echo 'Updating job status: IMPORTING' 'Downloading census blocks'
+ '[' -n '' ']'
+ '[' nonus == USA ']'
+ NB_BLOCK_FILENAME=population
+ S3_PATH=s3:///data/population.zip
+ '[' -f /data/population.zip ']'
+ echo 'Using local census blocks file'
+ BLOCK_DOWNLOAD=/data/population.zip
+ unzip /data/population.zip -d /tmp/tmp.jAv86ZgUP8/import_neighborhood
+ '[' nonus '!=' USA ']'
+ cd /tmp/tmp.jAv86ZgUP8/import_neighborhood
+ rm population.zip
rm: cannot remove 'population.zip': No such file or directory
```

This is because the `population.zip` file was never moved to the tmpdir, only unzipped there, therefore there is nothing to remove.

This error could be avoided by switching to `rm -f population.zip`, but then the process fails with the following error:

```bash
+ echo 'Updating job status: IMPORTING' 'Downloading census blocks'
+ '[' -n '' ']'
+ '[' nonus == USA ']'
+ NB_BLOCK_FILENAME=population
+ S3_PATH=s3:///data/population.zip
+ '[' -f /data/population.zip ']'
+ echo 'Using local census blocks file'
+ BLOCK_DOWNLOAD=/data/population.zip
+ unzip /data/population.zip -d /tmp/tmp.MeseJ3QXln/import_neighborhood
+ '[' nonus '!=' USA ']'
+ cd /tmp/tmp.MeseJ3QXln/import_neighborhood
+ rm -f population.zip
+ for x in *
+ mv population.cpg population.cpg
mv: 'population.cpg' and 'population.cpg' are the same file
```

Therfore I believe this block does not add value and can be safely removed. Once these lines where delete, the analysis completed successfully.

After that the analysis proceeds like a charm.

### DIFF
--- a/src/analysis/import/import_neighborhood.sh
+++ b/src/analysis/import/import_neighborhood.sh
@@ -137,7 +137,7 @@ then
             else
                 echo "Using blocks file from PFB_POP_URL: ${PFB_POP_URL}"
             fi
-            wget -nv -O "${BLOCK_DOWNLOAD}" "${PFB_POP_URL}"      
+            wget -nv -O "${BLOCK_DOWNLOAD}" "${PFB_POP_URL}"
 
             if [ "${AWS_STORAGE_BUCKET_NAME}" ]; then
                 echo "Uploading census blocks file to S3 cache"
@@ -145,14 +145,6 @@ then
             fi
         fi
         unzip "${BLOCK_DOWNLOAD}" -d "${NB_TEMPDIR}"
-
-        if [ "${PFB_COUNTRY}" != "USA" ]; then
-            # Rename unzipped files if not from census so they can be found easily
-            cd $NB_TEMPDIR
-            rm "${NB_BLOCK_FILENAME}.zip"
-            for x in *; do mv "$x" "${NB_BLOCK_FILENAME}.${x##*.}"; done
-            cd - 
-        fi
 
         # Import block shapefile
         update_status "IMPORTING" "Loading census blocks"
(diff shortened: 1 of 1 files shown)


KlaasH:
Thanks for writing this up.  The variety of sources that the analysis can pull its input files from (URLs, S3 cache, local files) makes it hard to keep the processing steps unified/coherent and to validate all the cases and combinations.  I think this was a case where we focused on one scenario (loading the population file from a URL) and neglected the one you're trying to use.

The `rm` command does do something in the case where the file is provided via `PFB_POP_URL`, and I think the `mv` commands are necessary in the situation where the zip filename itself doesn't match the names of the files that get extracted from it.  But yeah, we can't have them crashing in the simple `/data/population.zip` case.  I made PR #905 as an alternative fix that I think handles both situations.  Let me know if you think it will serve the purpose.

rgreinho:
@KlaasH Thanks for your answer.

I rebuilt my image with your branch (#905) and still got an issue. I commented on the PR with the output and a potential solution.

rgreinho:
Resolved by #905.

Issue #882, opened 2022-04-25
Port from Jenkins to Github Actions


BryanQuigley:
This includes:
* Basic CI tests
* Add staging deployments (and releases if in Jenkinsfile) to GH Actions
* Removing (or disabling project) in Jenkins

This might end up including adding Docker only instructions at least for Deployments.  (I had issues trying to upgrade the bastion)

BryanQuigley:
Found an ops issue indicating we really need to get to terraform 1.+ before using terraform in github actions.
It also just didn't work due to an S3 issue. 

Leaving branch still up (rough) at test/githubactions

PR #903, opened 2022-08-18
Port from Jenkins to Github Actions

BryanQuigley:
Most items copied over with only minor changes, except:

Commenting out update_batch_definitions as that would need a good number of changes
to support not using an AWS profile in Github and we are looking into improving that
as my potential next task anyway.

Slack notifications for build success or failure are covered when pushing a PR
by the Github integration.

REMINDER: Add infra apply when PR approved

Resolves: #882

### Demo

On PR, Terraform is not run: https://github.com/azavea/pfb-network-connectivity/runs/7904131095?check_suite_focus=true
On Push to to test (and others) it is run; https://github.com/azavea/pfb-network-connectivity/runs/7888427429?check_suite_focus=true

## Testing Instructions

 * Confirm GH Action output looks like Jenkins

## Checklist


BryanQuigley:
I left infra apply off until review is complete - plan to add it when it's done.

BryanQuigley:
The failure was due to GH Actions being down.

jacobscarfmerrell:
where in the [push to test action run](https://github.com/azavea/pfb-network-connectivity/runs/7888427429?check_suite_focus=true) from the test instructions does terraform run? I was looking for that `infra` script call that gets skipped in the PR run

BryanQuigley:
Look at this one: https://github.com/azavea/pfb-network-connectivity/actions/runs/3003482711/jobs/4821824010

The ones that are made from the PR will not run it.

jacobscarfmerrell (approved):
lgtm!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - develop
+      - test/**
```
KlaasH, line 7: This should include `release/` branches as well, and they should be included in the conditions controlling whether to deploy.  We don't always push release branches when we're doing a release for this project, but if we did we would want them to build.

scripts/infra
```diff
@@ -55,7 +55,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
         case "${1}" in
             plan)
-                update_batch_definitions
+                #update_batch_definitions
```
KlaasH, line 58: I ran through the steps of a staging deploy through `infra play` and got "BATCH_ANALYSIS_JOB_NAME_REVISION: unbound variable", but I guess that's expected for now, and the workaround is just to comment this back in when doing a manual deploy (e.g. to production).

.github/workflows/ci.yml
BryanQuigley: adding.

BryanQuigley:
We do get build notifications when the build is from a PR.  but not automagically when pushing manually, etc.  That should be just 1 point to add in the future if we find ourselves really needing it.

PR #899, opened 2022-07-15
 Convert kmph in analysis to mph (#892) 

jacobscarfmerrell:
## Overview

Make analysis convert kmph speed limits to rounded mph

### Demo

```
pfb=# SELECT osm.maxspeed, osm.osm_id FROM neighborhood_osm_full_line osm, neighborhood_ways WHERE neighborhood_ways.osm_id = osm.osm_id                                                         
AND (osm.maxspeed LIKE '% kmph' OR osm.maxspeed ~ '^\d+(\.\d+)?$');
 maxspeed |   osm_id   
----------+------------
 100      |   13224440
(...and a bunch of other rows...)
pfb=# SELECT osm.maxspeed, osm.osm_id FROM neighborhood_osm_full_line osm, neighborhood_ways WHERE neighborhood_ways.osm_id = osm.osm_id
AND (osm.maxspeed LIKE '% kmph' OR osm.maxspeed ~ '^\d+(\.\d+)?$') AND osm.osm_id=13224440
pfb-# ;
 maxspeed |  osm_id  
----------+----------
 100      | 13224440
(1 row)
pfb=# SELECT ROUND(substring(osm.maxspeed from '\d+')::INT / 1.609 / 5)*5
FROM    neighborhood_osm_full_line osm, neighborhood_ways
WHERE   neighborhood_ways.osm_id = osm.osm_id
AND     (osm.maxspeed LIKE '% kmph' OR osm.maxspeed ~ '^\d+(\.\d+)?$') AND osm.osm_id=13224440;
 ?column? 
----------
       60
(1 row)
pfb=# SELECT speed_limit FROM neighborhood_ways WHERE osm_id=13224440;
 speed_limit 
-------------
          60
(1 row)
```

### Notes

Optional. Anillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

 * Before running an analysis, place a bash right before line 266 of `import_osm.sh`
 * when the analysis catches on that bash a while later, run `psql -h $NB_POSTGRESQL_HOST -U ${NB_POSTGRESQL_USER} -d ${NB_POSTGRESQL_DB}` to open a psql shell
 * look above in the demo section and run those commands, changing the osm_id at the end of the 2nd, 3rd, and 4th queries to match an osm_id you see in the 1st command's output
   * the 1st one looks at the existing maxspeeds
   * the 2nd one zooms in on a specific row (you will likely have to change the osm_id to one you see from the 1st command's output). take note of the maxspeed
[...]

PR #904, opened 2022-08-23
Add neighborhood speed limit field and fix km/h conversion in map legend

KlaasH:
## Overview

This adds a couple elements that were missing from the handling of non-US speed limits.  PR #899 handled importing speed limits for segments that are in km/h and converting them to mph for the analysis, but there's also the default residential speed limit, which we get for US cities using state and city FIPS codes, and there's the box on the map that shows the default limit.  This adds an alternative way to provide the speed limit, and adjusts the calculation used by the legend to round km/h values to the nearest 5.

The new way to enter a default residential speed limit for is as a field on the neighborhood.  Here's what the form looks like now for a non-US city:
image

The field only shows up for non-US cities, to avoid confusion between the two ways of setting a value (if they should somehow both get set, the model will allow that and the analysis will prefer the value from the new field rather than the FIPS code table).  It gets provided to the analysis and used in the section of `import_osm.sh` where the speed limit is looked up and saved.  From there, it acts the same as the current setup--gets stored in the database for use within the analysis, then exported to a tiny CSV file, then imported into the AnalysisJob instance for use on the frontend.

Here's how it shows up on the frontend: 
image

### Notes

This PR also includes a couple tweaks to the development tooling, both related to increasing available disk space on the virtual machine.  One expands the size of the filesystem inside the VM, the other changes the `run-local-analysis` script to remove containers automatically, so they don't stick around and fill up the disk after repeated testing.

## Testing Instructions

I pushed these changes in a test branch (https://github.com/azavea/pfb-network-connectivity/tree/test/non-us-speed-limit-badge), so they're currently up on staging.

For testing locally:
- Update and apply migrations
- Make or edit a non-US neighborhood and fill in the default speed limit.  It should load that value and when you view the results on the frontend, it should show up on the map as shown above.
- Other scnenarios (non-US cities with no value for the new field, US cities with or without city FIPS codes) should work as before

## Checklist


jacobscarfmerrell (commented):
looks good! I'll be ready to approve after testing once [this](https://github.com/azavea/pfb-network-connectivity/pull/905#pullrequestreview-1082833424) is in

jacobscarfmerrell (approved):
ran great!

Issue #911, opened 2022-09-26
Improve failure mode when population is zero


KlaasH:
The city of Milton Keynes has been [crashing late in the analysis](https://bna.peopleforbikes.org/#/admin/analysis-jobs/528bb855-5296-435e-b0ba-09bac7f7d068/) and, worse still, creating a record in the database that crashes any AnalysisJob query that touches it.

It turns out the main source of the problem is that there's no population.  The population file used by the failing jobs, https://pfb-public-documents.s3.amazonaws.com/population/tabblock2010_99_pophu.zip, seems to be missing a lot of population values.  Loading it QGIS and looking at the data table, I can see it has 181,408 records, but when I filter for "POP10 = NULL" it shows there are 115,096 records that match.  Some blocks without population might make sense, but almost 2/3rds doesn't.  I didn't manage to get a sense of where the population is and where it's missing, and whether it's all-or-nothing across big areas or if they're interspersed, but when I click around the area of Milton Keynes I didn't find any non-null blocks.

So what's happening is that the population is getting imported, and having the geometries is enough to do all the stress and connectivity calculations, but then at the end when it tries to export the scores they're all zero, including total population.  Zeroes in the other scores are allowed/expected, but for population, it turns out it will insert into the database OK, but then any query that tries to load the scores will fail a the database level, because it's looking for an integer value from the population key within the JSONB field, and getting an empty string instead.

Steps I think we should take to address this:
- Do a validation query after we've loaded the census blocks and trimmed them to the buffered boundary.  If the total population of the remaining census blocks is zero, we should raise an error immediately.  This is fairly early in the analysis, and will both make it easier to see what went wrong and avoid wasting time running an analysis that's doomed to fail.
- In [`load_overall_scores.py`](https://github.com/azavea/pfb-network-connectivity/blob/0.16.0/src/django/pfb_analysis/management/commands/load_overall_scores.py#L21-L23), don't let a non-integer population value get through.  Right now there's an `if ValueError: pass`, which I think is there because some entries in the metrics table it's working with aren't numbers and that's fine (though it might be worth double-checking that).  That shouldn't apply to the population total.  If it's not an integer, the script should raise an exception.

It would be nice to also prevent a bad record that gets into the database from crashing the API when any query tries to interact with it, but I think it would be tricky to identify exactly those cases and do something useful--i.e. don't crash but also don't just swallow potentially serious errors silently.  So I think it makes sense to focus on preventing bad records from getting into the database in the first place.

Issue #910, opened 2022-09-20
Analysis caches non-US population file then uses it instead of provided URL


KlaasH:
The process for getting the population/blocks file for the analysis was already a bit complex--looking for a local file first, then one cached on S3, then downloading from Census and caching on S3 if neither of the other sources worked.  Adding the parameter to provide a file from non-US cities made it even more so, and the interactions aren't quite right.

Rebecca ran into an issue where all non-US analysis jobs are pulling the same population file because after the first time it was provided to an analysis via the new parameter, it was cached as `population.zip` on S3 and subsequent runs always look for that and use it in preference to the provided URL.

I think the way to fix this issue and make the whole thing more correct is to set it up so:
- The PFB_POP_URL parameter is always respected. I.e. if that is provided to the analysis, it gets used every time and neither the local `/data/` directory nor the S3 cache gets checked.
- If that parameter is blank, look for a file in the local `/data/` directory with the standard filename (`population.zip` for non-US cities, `tabblock2010_`-something for US cities), and use it if it's there.
- If neither of the above applies:
   - If the city isn't in the US, throw an error. There's no generic source for non-US population files, so if it's not provided in one of those two ways, we're stuck.
   - If the city is in the US, follow the same process we were using before:
      - Look in the S3 cache for the `tabblock` file and use it if it's there
      - Otherwise download it from Census, then upload it to the S3 cache for next time

Compared to the current setup, this would
1) avoid any cache mismatch by only caching US files downloaded from Census
2) never ignore PFB_POP_URL (currently it always gets ignored for US cities, and gets ignored for non-US cities if there's a local or cached file)

Issue #909, opened 2022-09-13
OSM extract downloader should use "great-britain" in the URLs, not "united-kingdom"


KlaasH:
Geofabric extract URLs are organized under a geographical hierarchy, and we already had to do some fiddling to construct the URLs (specifically, making a country->continent mapping), but we missed the fact that our country list and theirs doesn't agree on what name to use for the island nation in the north Atlantic that has 9 words in its official name.  We'll need to add a special case.

Issue #908, opened 2022-09-13
Neighborhood country search is searching full names rather than abbreviations


KlaasH:
The "Country" column on [the Neighborhoods management page](https://bna.peopleforbikes.org/#/admin/neighborhoods/) isn't searching properly.  The table shows two-letter country codes, but it turns out the search is being done against full country names (e.g. when the column says GB, you can't search for "GB" and get any neighborhoods, but you can get them if you search "united kingdom of great britain and northern ireland", or any part of that.  This is surprising and should be fixed to just search by the thing that's shown in the table.

PR #912, opened 2022-09-29
Fixes for UK OSM URL, non-US population file caching, and country search

KlaasH:
## Overview

This is a grab-bag of three fixes for issues affecting the analysis or the admin views:

Issue #909 
The OSM extract download for places in The United Kingdom of Great Britain and Northern Ireland wasn't working because the package we're using for country names had its short name as "United Kingdom" but the OSM extract URL has it as "great-britain". So this adds a special case.

Issue #910 
The logic for where to get the population file was going wrong because it was using a standard name for non-US population files (`population.zip`) but then both caching it and checking for a cached file under that name.  So the first file to get a `PFB_POP_URL` value downloaded the file and used it as intended, but all subsequent runs were actually ignoring that variable and preferring the `population.zip` file that had been saved to the S3 cache.
This changes the logic to always prefer `PFB_POP_URL` and to only save and use cached population files for US locations.

Issue #908 
The country search in [the Neighborhoods admin view](https://bna.peopleforbikes.org/#/admin/neighborhoods/) started acting weird with the new release--searching on the full name of the country rather than the alpha-2 code, which is what's shown in the table and what it used to search on.  It turned out that was an intentional change in the package we're using to provide our country field, so this just downgrades that package to the last version before the change.

### Demo

The OSM and population files are hard to demonstrate visually, but here's the country search doing the right thing:
image

### Notes

We could also change the lookup for the country search to "iexact" rather than "icontains", which hasn't been overridden by the package to search against the full name.  But right now entering one letter gets you all countries that contain that letter, whereas if we made it exact, entering one letter would clear the list.  The former seems nicer to me, and I don't think there's any real cost to holding the package on an earlier version, so it seemed like the way to go.

## Testing Instructions

- OSM fix: make a neighborhood in the UK (here's the boundary file for Milton Keynes: [milton-keynes.zip](https://github.com/azavea/pfb-network-connectivity/files/9676850/milton-keynes.zip) and start an analysis for it.  It should successfully download and import the OSM extract (you can cancel it after that. see issue #911)
- Population file: as noted above, the logic around this is somewhat involved, and it would be a fair bit of work to try to test every case. It probably makes sense to kick off a non-US job with a `PFB_POP_URL` value set and confirm that it doesn't get uploaded to your storage bucket as `population.zip`, and that even if there is a file named `population.zip` in your storage bucket (i.e. at `s3://pfb-USERNAME-storage-us-east-1/data/population.zip`), when the `PFB_POP_URL` is set for a job it uses that and doesn't try to download the file from S3.
- Country search: just try it out on [the Neighborhoods view](http://localhost:9301/#/admin/neighborhoods/)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Upcoming release]
 
+#### Fixed
+- Use 'great-britain', not 'united-kingdom', in OSM extract URLs
+
 ## [0.16.0] - 2022-09-06
 
 #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 #### Fixed
 - Use 'great-britain', not 'united-kingdom', in OSM extract URLs
 - Stop caching non-US population files
+- Downgrade django-countries package to fix country search in Neighborhoods admin
 
 ## [0.16.0] - 2022-09-06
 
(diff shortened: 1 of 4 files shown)


jacobscarfmerrell (approved):
lgtm! it didn't use the population.zip in my S3 bucket and the country search worked :fire:

PR #916, opened 2022-10-14
Only get state OSM extracts for US states

### LINKED ISSUE #915: OSM extract downloader should ignore state/province for non-US places
When we added non-US support to the [OSM extract download script](https://github.com/azavea/pfb-network-connectivity/blob/develop/src/analysis/scripts/download_osm_extract.py) we missed the fact that it's assuming that the existence of a `state_abbrev` means we're dealing with a US state.

Geofabrik has state-level extracts for some countries (like [France](https://download.geofabrik.de/europe/france.html) and [Germany](https://download.geofabrik.de/europe/germany.html)) but not for others (like [Australia](https://download.geofabrik.de/australia-oceania/australia.html)).  To support using those state extracts, we would need to create a hard-coded list of which countries have them and which don't, and we might need to hard-code the names, too, since there's a good chance the names in the Geofabrik URLs don't exactly match the names we use.

So we should just change the script to pull the country extract for any non-US neighborhood, even if there's a state defined for it.  Which is kind of a bummer for countries with a lot of OSM data where the full extract will be pretty large, but it seems worth it to accept a little slowness in that step of the analysis to avoid more complexity and potential for errors in the downloader.

KlaasH:
## Overview
The OSM downloader was treating any state abbreviation as a US state and trying to download a state extract for it.  But now there are non-US neighborhoods with states that are using this script, so it needed to have that logic protected by a "only in the US" check.

Geofabrik has state-level extracts for several other countries, but knowing which countries it has and which it doesn't, and getting the names to match up exactly right, would be fiddly and error-prone, so we'll just stick with national extracts.

Resolves #915

### Notes
I noticed when I went to update the changelog that I had missed some steps when I updated it for the last release, so I fixed that up here.

## Testing Instructions
I tested it by adding `raise Exception("Time to download OSM extract from " + osm_extract_url)` to the `download_from_geofabrik` function right before it actually tries to download from Geofabrik.  That way I could kick off several different analysis runs (in the US and elsewhere, with and without the state field set) and not have to wait for it to actually try to download the extract or finish the analysis.  I just verified that all the URLs it said it was going to grab were correct.

Note that you have to make sure the file it needs isn't in the `osm-data-cache` on your S3 bucket, or the script will download it from there and not try to get it from Geofabrik.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Upcoming release]
 
-## [0.16.0] - 2022-10-06
+## [0.16.1] - 2022-10-06
 
 #### Fixed
 - Use 'great-britain', not 'united-kingdom', in OSM extract URLs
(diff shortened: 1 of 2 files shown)


jacobscarfmerrell (approved):
makes sense! I didn't think about international countries with states

PR #925, opened 2022-11-30
Add endpoint to serve crash data (#917) and facilitate crash data imports (#918)

### LINKED ISSUE #917: Add Crash endpoint
This will require having a sense of the shape of the incoming crash data

Acceptance Criteria
- new endpoint facilitates requesting crashes
- the endpoint requires a neighborhood to filter the crashes by

### LINKED ISSUE #918: Import crash data
Criteria:
- Crash data is imported in Django after running scripts/update
- Data covers the last five years

Proposed implementation:
- Upload cleaned data to a shared S3 bucket
- Write a management command to import crash data from S3
- Write a data migration that runs the above management command

lydiascarf:
## Overview

I've grouped these two issues together since the work in #918 makes it possible to test #917

### Notes

This endpoint still excludes the boundary buffer.

## Testing Instructions

 * Add an unzipped copy of [crashes.zip](https://github.com/azavea/pfb-network-connectivity/files/10127340/crashes.zip) to the data directory in your S3 bucket
 * Inside vagrant, run `scripts/django-manage migrate`
 * Inside vagrant, run `scripts/django-manage import_crash_data`
 * See if you find any crashes at `http://localhost:9301/api/crashes/?uuid=some_analysis_job_uuid`

## Checklist

Resolves #917 
Resolves #918


src/django/pfb_analysis/management/commands/import_crash_data.py
```diff
@@ -0,0 +1,72 @@
...
+
+logger = logging.getLogger(__name__)
+
+def download_csv(csv_key):
+    tmpdir = tempfile.mkdtemp()
+    s3_client = boto3.client('s3')
+    s3_client.download_file(settings.AWS_STORAGE_BUCKET_NAME,
+                            csv_key,
+                            tmpdir+'/'+'crashes.csv')
+    return tmpdir
```
KlaasH, line 23: A few notes on the downloader:
- Since the file compresses so well (3.2MB for the .zip vs 58MB for the .csv) it would be nice to store and download it in zipped form.  And extracting a zip file is pretty easy in Python.
- Since the filename is hard-coded, I'm not sure there's much value in passing some of these parameters around.  I would be inclined to just hard-code the names where they're needed and only pass values for things that aren't know ahead of time (like the tmpdir) or are used repeatedly.

Here's a version of `download_csv` that makes both those changes:
```python
def download_csv():
    tmpdir = tempfile.mkdtemp()
    zipfile_path = os.path.join(tmpdir, "crashes.zip")
    s3_client = boto3.client('s3')
    s3_client.download_file(settings.AWS_STORAGE_BUCKET_NAME,
                            "data/crashes.zip",
                            zipfile_path)
    with zipfile.ZipFile(zipfile_path, "r") as zip_ref:
        zip_ref.extractall(tmpdir)
    return os.path.join(tmpdir, "crashes.csv")
````

So then in `import_csv` you could just do `csv_path = download_csv()` and hand the resulting path to the `open()` call.  I think that reduces the overall amount of hard-coding and makes it a bit more contained without actually reducing flexibility.

**Review by KlaasH — CHANGES_REQUESTED** [review]
Some notes on the download/load code.  I was able to run it successfully and get results from the API, and I'm still looking at that, but in the meantime I have some thoughts and questions on the loading logic.

File: src/django/pfb_analysis/management/commands/import_crash_data.py [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+        with open(import_tmpdir+'/'+csv_filename, 'r') as csv_file:
+            reader = csv.DictReader(csv_file)
+            for row in reader:
+                crash, created = Crash.objects.get_or_create(
+                    fatality_type=get_fatality_type(row['fatal_typ']),
+                    geom_pt=Point(float(row['LONGITUD']), float(row['LATITUDE'])),
+                    year=row['YEAR'],
+                    )
+                if created != True:
+                    logger.exception('Crash not created', vars(crash))
```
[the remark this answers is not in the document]
lydiascarf: you're right. I don't see a reason to enforce any kind of uniqueness

KlaasH (changes requested):
The import and API code is looking good now.  A few more notes on the latter below, but I think the way it's set up makes sense, and it's successfully returning results.

**Query speed**
With that in place, I dug into the query itself a bit and reached the following conclusions:
- In the current setup, both the boundaries and the crash point geometries are stored using the PostGIS `geometry` type.
- The `ST_DWithin` function has two different modes--geometry or geography--and since both fields are geometry, it's using that one. Which means the distance argument is treated as being in the units of the field SRS, i.e. degrees in this case.
- So the extreme slowness you were seeing when adding a buffer to the current query was not because it's failing to run efficiently (there are indexes on both fields and it's able to use them) but because "within 2650 degrees" means "everywhere", so it's trying to return all 171k crashes.
- Since the crash data spans the country, it wouldn't be ideal to pick a single projected coordinate system to use for these queries (it could get inaccurate around the edges), but it would also be nice not to have to deal with deciding between different projections based on where the neighborhood is (more accurate but complicated and slow).  Using the geographic form of the DWithin query would be ideal.
- Running the query with the fields as they are but casting either or both of them to `::geography` works.  It runs a geographic query and returns the right results--given a boundary and a buffer distance in meters, it returns a reasonable number of records.
- It's faster than the query that's timing out, since it's not trying to serialize the whole table, but it's still kind of slow, because it's not managing to use indexes for searching--the indexes are built on the geometry fields, not the geography values we've cast them to.
- If both fields (the boundaries and the crash points) are stored as geography fields (which is as easy as setting `geography=True` in the model field declaration, and migrating), the DWithin query runs in geography mode and uses the indexes, so it's very fast.

I made a commit, https://github.com/azavea/pfb-network-connectivity/commit/80a58902fc288fac1acdb9739e34bc2f9b16381c, that makes the switch to that. It changes the field on `Crash` to `geography=True` and adds a `geog` field to `Neighborhood`, with a an SQL data migration to copy the geometries into the new field.  I was torn about whether to add a new field or convert the existing `geom` field to `geography=True`, since I think would work just fine for the other purposes we're putting that field to if it were stored as `geography`.  But I'm not sure, so I figured it's better to be cautious.  I don't think adding to the size of the neighborhoods table by storing the extra copy is a problem.

So that's my take on the querying issue.  I was pleasantly surprised by how easy it was to make the switch and how much of a difference it makes (from ~1800ms down to ~50ms for the Philadelphia test query I was doing locally).

**Loading and documentation**
Right now there are instructions for how to load crashes in the testing instructions of this PR, but there's no mention in the README of the crash data, where it comes from, or how to load it.

I think the first step to addressing that should be to make it more automatic, and that a data migration that imports and runs the management command probably makes sense.  That way it'll get loaded automatically when people spin up new dev instances, and deploying the data to staging and production would just be a matter of running migrations.

I also think the `crashes.zip` file should live in the `pfb-public-documents` bucket.  It makes sense to enable loading it from `settings.AWS_STORAGE_BUCKET_NAME` for testing changes or updates (maybe a command-line option like `--use-dev-bucket`?), but I think the default should be a single shared copy, rather than having to direct people to get the file and copy it to their own bucket as part of project setup.

With those two changes, the README wouldn't have to tell people where to find the file or what command to run, it could just need to have a section noting that the crash data exists and gets loaded automatically via migration.

src/django/pfb_analysis/models.py
```diff
@@ -235,6 +236,7 @@ def set_boundary_file(self, geom):
             boundary_file = File(open(zip_filename, 'rb'))
             self.boundary_file = boundary_file
             self.geom = geom
+            self.geog = geom  # Also save as geography for fast lookup in the Crashes query
```
KlaasH, line 239: I missed that there are actually two places where we save the boundary to the Neighborhood model--this one, where we're loading boundaries from a batch file, and in `_set_geom_from_boundary_file()`, where we load from the shapefile provided in the Neigoborhood add/edit form.  So there's should be another copy of this line in that method, down at line 331.

src/django/pfb_analysis/views.py
```diff
@@ -355,3 +360,29 @@ def get(self, request, format=None, *args, **kwargs):
...
+
+    filter_class = None
+    pagination_class = None
+    permission_classes = (RestrictedCreate, IsAuthenticatedOrReadOnly)
+    serializer_class = None
+
+    def get(self, request, format=None, *args, **kwargs):
+        analysis_job_uuid = request.GET.get('uuid')
+        analysis_job = get_object_or_404(AnalysisJob, pk=analysis_job_uuid)
+        neighborhood_geometry = analysis_job.neighborhood.geom
```
KlaasH, line 376:
```suggestion
        neighborhood_geometry = analysis_job.neighborhood.geog
```
I guess we could also rename the variable to e.g. `neighborhood_boundary` to avoid muddling the distinction.

KlaasH (approved):
A couple fixups noted below, related to getting the geography field loaded and using it in the query.  I did a test with those changes applied and was able to load a new Neighborhood, create an analysis job for it, and get the crashes for the analysis job from the API.

So with the two fixups, this will be good to go :+1:

PR #932, opened 2023-01-05
Change crashes copy in layer picker (#930)

### LINKED ISSUE #930: Change crash label text and re-deploy to staging
Change "Crashes" -> "Fatal Crashes (2016 - 2020)" in the layer picker and confirm that it looks alright. We can omit the parens if it helps.

Re-deploy to staging once done.

lydiascarf:
From "Crashes" to "Fatal Crashes (2016 - 2020)". The scrollbar overlaps a little with the text but only when it's highlighted.

Screenshot from 2023-01-04 10-31-56
Screenshot from 2023-01-04 10-32-44

Resolves #930


KlaasH (approved):
:+1: Looks good!

PR #935, opened 2023-01-06
Fix analyses using pre-2018 census data (#934)

### LINKED ISSUE #934: Analysis jobs in states with older census data are crashing

lydiascarf:
## Overview

States like Alaska use census data from as far back as 2016. Our current default is 2019, but the import jobs script should recursively attempt prior years. This work stops recursive calls to fetch_census_data from crashing on 404s from the census data endpoint, which was happening when the most current census data for a state was prior to 2018.

## Testing Instructions

 * Verify that you can run an analysis job in Alaska

## Checklist

Resolves #934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### Fixed
 - Ignore state for non-US neighborhoods when downloading OSM extract
+- Fix analyses using pre-2018 census data
 
 ## [0.16.1] - 2022-10-06
 
--- a/src/analysis/import/import_jobs.sh
+++ b/src/analysis/import/import_jobs.sh
@@ -35,34 +35,32 @@ NB_POSTGRESQL_PASSWORD - Default: gis
 }
 
 function fetch_census_data() {
-    set +e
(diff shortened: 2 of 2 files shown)


KlaasH (changes requested):
I think this produces the right results, but when I was testing it, I found it weird that it ended up downloading the 2016 file for Alaska from the Census, uploading it to S3, then immediately downloading it from S3 again.  In figuring out why that was happening, I came to feel that having some of the logic get run once in `import_job_data` then duplicated within `fetch_census_data` made it more confusing than it needed to be, and that the order of the conditions and the placement of the recursive call could be improved as well.

So I have a counter-proposal, https://github.com/azavea/pfb-network-connectivity/commit/8676d65a081e4bd88b57f493cac3fa669e2fcb18, which consolidates and rearranges things a bit, into something I think does the right thing more directly and is easier to read and understand.  I also made the year into an argument to the function, because even though it works the same, decrementing it in place and using it in the next call with the new value weirded me out.

Have a look, and feel free to pull my commit in or make some third version, as you see fit.

lydiascarf:
@KlaasH I like your organization so I just took it as is! Tested it and it runs well on my machine

KlaasH (approved):
Ok, cool. Good to go, then! :+1:

Issue #839, opened 2021-05-18
Determine resource requirements for Batch jobs


rbreslow:
The Batch jobs have an existing shape, but we'll have an opportunity to choose different instance types that have more flexibility around temporary storage when we move to a managed CE. We should determine what the least vCPU, memory, and temporary storage requirements are.

The general shape of things as they are now:

- Batch jobs [require](https://github.com/azavea/pfb-network-connectivity/blob/develop/deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json) 8 vCPUs and 30 GB of memory. 
- The container instance in production is an `i3.2xlarge`, [which has](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/storage-optimized-instances.html#storage-instances-hardware) 8 vCPUs and 60 GB of memory. 
- According to [instance store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes), `i3.2xlarge` has a single 1,900 GB NVMe volume for ephemeral storage.

Some questions to guide thinking:

- How much temporary storage does the job actually need?
- Can we tune down vCPUs/memory for the job? Right now, we're requesting an instance that wastes 30 GB of memory. If we're able to scale down, we may be able to pack multiple jobs onto a single instance.

KlaasH:
Ok, I did some looking into this. This isn't an answer, exactly, but it narrows things down and hopefully gives us a path to an answer.

### Current resources
#### CPUs
8 VCPUs.  This was increased from 2 in PR #355 because we had to stop running multiple jobs per instance, so it made sense to fully utilize the instances with each job.  (Actually it was first increased from 2 to 7, leaving one VCPU for the jobs that wrote out static tile pyramids. Then when we switched to Tilegarden and didn't need that tiling task anymore, that last CPU was added to the analysis job's allocation.)

#### Memory
Total memory: 30GB. This was increased in PR #355 from 12.3GB. The WORK_MEM, SHARED_BUFFERS, and MAINTENANCE_WORK_MEM variables were also increased (doubled).

[...]

PR #945, opened 2023-05-08
Increase analysis RAM allocation and fix provisioning

### LINKED ISSUE #940: Batch processing sometimes fails
Notes from Klaas:
- You can see jobs and filter by status here (the default filters aren't useful and apparently setting them doesn't affect the URL, so you have to select the queue and change the filters every time).  That has a "total run time" column, which is useful for isolating jobs that ran a long time but ultimately failed.
- When you click into the details for a job, there's a "Status reason" field in the "Job information" box. Some of those say "AnalysisJob terminated via API by rebecca@peopleforbikes.org", so those don't count as failures.
- Ones that say "Essential container in task exited" with status FAILED, like this one, are the interesting ones in terms of trying to identify problems.
- That same "Job information" panel on the detail view has a "Log stream name" field that links to the logs for that task.  If you click through to the logs for that failed one, the spot where things went wrong says:
```
psql:../features/paths.sql:59: server closed the connection unexpectedly
This probably means the server terminated abnormally
```
- We may need to get postgres to log to stdout in order to see a better error message since the logs are in the Postgres container, which doesn't get persisted
  - this may be difficult because the pg container runs as a background process

KlaasH:
## Overview

The functional change here is increasing the memory allocation to the analysis Batch task to use the full RAM available on the i3.2xlarge instance we're using in the Batch compute environment.  I had [previously concluded](https://github.com/azavea/pfb-network-connectivity/issues/839#issuecomment-978232796) that the jobs were only using about half of the 32GB we had allocated, but the biggest jobs have been crashing recently and it's possible memory is to blame.  And there's no reason not to use all the memory on the instance.

I'm opening this as a draft PR because I already deployed the job definition update to production, so what's left is the changes I made in the course of spinning up my dev instance again and doing a limited deploy (I reused the containers, so only the services that have the job definition ID embedded in their parameters needed to change).

Here's what else is going on here:
- There are separate job definition templates for staging and production, and historically staging used smaller instances and lower resource allocations.  That probably made sense early in development, but I think at this point it makes more sense to make staging as similar to production as we can, so I updated all the staging parameters to match the production ones.
- I added an 'update-tfvars' subcommand to `infra`, just for convenience.
- The base image for the `angularjs` and `tilegarden` containers is old enough that its package directory no longer exists in the main Debian repo.  For `tilegarden`, I was able to update the base image from `stretch` to `buster` and it worked fine.  The `angularjs` container uses an older version of Node, and the Grunt build failed with the oldest Node version I could get with `buster`, so I left it on `stretch` and updated the apt repo to use http://archive.debian.org/debian.
- Along the same lines (things being old): the Bower registry is still plugging along, but the `bower install` started throwing "certificate expired" errors.  I think the certificate is actually fine and the problem is that it's based on a root certificate that the container doesn't have because it's too new.  But I didn't go into it too much, just worked around it as recommended [here](https://github.com/bower/bower/issues/2608).

Resolves #940 (hopefully)

## Testing Instructions

When the time comes, this should be tested by:
- Spinning up a dev instance and making sure the front end, API, and analysis (run by hand from the command logged by the API) work as usual.
- Doing a staging deploy to make sure the provisioning parts are working.

## Checklist

### DIFF
--- a/deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
+++ b/deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
@@ -5,7 +5,7 @@
     "containerProperties": {
         "image": "{image}",
         "vcpus": 8,
-        "memory": 30720,
+        "memory": 61000,
         "command": [
             "/pfb/scripts/entrypoint.sh"
         ],
--- a/deployment/aws-batch/job-definitions/staging-pfb-analysis-run-job.json
+++ b/deployment/aws-batch/job-definitions/staging-pfb-analysis-run-job.json
@@ -4,8 +4,8 @@
     "parameters": {},
     "containerProperties": {
         "image": "{image}",
-        "vcpus": 2,
-        "memory": 12288,
+        "vcpus": 8,
+        "memory": 61000,
         "command": [
             "/pfb/scripts/entrypoint.sh"
         ],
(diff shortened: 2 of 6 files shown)


KlaasH:
Closing this, since these changes got folded into PR #947.

Issue #937, opened 2023-02-28
CI task definition update process is broken


lydiascarf:
> Yeah, the part of the build script that updates the batch job definition worked on Jenkins but doesn't work on Github Actions because it relies on a configured AWS profile, which we don't have set up there.  So the options, from simplest to most involved, would be
> - Add a condition to skip that step in CI. That would get the CI builds working except for running analysis jobs, so we'd need to remember to do a manual deploy to staging any time we want to run the analysis there (which isn't that common. Usually we only do that when we're getting ready to do a release.)
> - Configure an AWS profile on Github Actions so it'll work with the job update script as-is.  We didn't do this when we switched to Github Actions, but I think it's possible.  Maybe even not that difficult.  Or it might be a layer of configuration that we don't want to deal with. I'm not sure.
> - Bring the batch job definition update into Terraform.  I believe it was handled separately during initial development because Terraform didn't have full support for Batch actions at that point, but I'm sure it does by now.  This would produce the best long-term outcome, since there wouldn't be a separate script or process at all, the Batch changes would just be applied along with the other Terraform actions.
> -- @KlaasH

PR #947, opened 2023-05-12
Strip backslash from OSM names (and fix provisioning and change analysis parameters)

### LINKED ISSUE #946: Brooklyn analysis results are missing some road segments

KlaasH:
## Overview

The most important part of this PR is the smallest—a pair of lines in `import_osm.sh` that clears any backslash characters from the `converted.osm` file before we feed it to `osm2pgrouting` to import into the database.  It turns out `osm2pgrouting` *really* doesn't like backslash characters.  The error looks like this:
image

That certainly looks like an error, but I had assumed it would only affect the actual segments with bad names. But it turns out where it says "Vertices inserted: 0   Split ways inserted 0" after the errors, that's an indication that it lost a lot more than the affected segments.  Apparently it processes the input in chunks, and whatever other segments happen to be in the same chunk as the bad segments just get dropped on the floor.

I don't believe there's any functional purpose for which we would want backslashes in our OSM input. They're not used for encoding (the file is in UTF-8) and we don't actually use segment names for anything in the analysis anyway.  So I just did a simple `sed 's/\\/backslash/'` to make them go away.

*Other items:*
- The next most important thing this does is increase the memory available to the analysis task from 32GB to 61GB.  It only matters for Los Angeles, I think, but it does seem to matter for Los Angeles. And we're using i3.2xlarge instances anyway, so we might as well allocate all the memory.
- I also adjusted some of the parameters used to set PostgreSQL memory and disk usage limits.  We have lots of memory and lots of disk, so it makes sense to be generous, to get all the speed we can and avoid any failures in the largest jobs.
- Speaking of those parameters, I made the staging ones match production. They had been set lower, presumably to make repeated testing cheaper, but we only very occasionally test on staging these days, so it makes sense to make it as similar to production as possible.
- There were a few errors in the build when I tried to spin my development instance back up.  I tried to find the path of least resistance for getting around them, which meant upgrading the base image in the case of the tilegarden container but just adjusting the Debian package list to be able to keep using the old image for the angularjs container.

### Demo

I found the location of one of the segments named `"\"` and made a tiny boundary centered on it.  Here's are the before-and-after shots showing the affect of losing chunks from the OSM import:
Before: 
image

After: 
image

### Notes

The GitHub Actions deployment to staging isn't working (see #937), but I pushed this branch to staging manually. 

## Testing Instructions

I don't think it seems necessary to spin up a Batch worker on staging and do more runs there, but it probably makes sense to test this locally to the point of running a small analysis.

- Spin up a dev instance per the README instructions
- Make a neighborhood for this boundary (via the ["Neighborhoods" management page](http://localhost:9301/#/admin/neighborhoods/)): [backslash_location.zip](https://github.com/azavea/pfb-network-connectivity/files/11467594/backslash_location.zip) (use FIPS code 3651000)
- Run an analysis for it (click "Run Analysis" on the [Jobs page](http://localhost:9301/#/admin/analysis-jobs/) then copy-paste the command that gets printed into the Docker log into a shell within the VM).
- Confirm that it runs successfully and you can see the results by clicking on it in the [All places list](http://localhost:9301/#/places////) once it has finished.

## Checklist

Resolves #946


src/analysis/scripts/setup_database.sh
```diff
@@ -39,7 +39,7 @@ ALTER SYSTEM SET maintenance_work_mem TO '${PFB_MAINTENANCE_WORK_MEM}';
 ALTER SYSTEM SET max_wal_size TO '${PFB_MAX_WAL_SIZE}';
 ALTER SYSTEM SET shared_buffers TO '${PFB_SHARED_BUFFERS}';
 ALTER SYSTEM SET synchronous_commit TO ${PFB_SYNCHRONOUS_COMMIT};
-ALTER SYSTEM SET temp_file_limit TO ${PFB_TEMP_FILE_LIMIT};
+ALTER SYSTEM SET temp_file_limit TO '${PFB_TEMP_FILE_LIMIT}';
```
KlaasH, line 42: This had been a (large) numerical value, but changing it to "100GB" made it start breaking without quotes.

deployment/aws-batch/requirements.txt
```diff
@@ -1 +1 @@
-boto3==1.4.4
+boto3==1.26.132
```
KlaasH, line 1: This is the requirements file for the little script that updates the batch task definition. Long-term we want to get rid of it and handle the task definition in Terraform, but for now we're still using it. This upgrade was necessary because the very-old version of boto3 didn't know about the `linuxParameters` option.

ddohler (approved):
Everything seems to be working great! I wasn't able to actually view the details for the job due to an apparent CORS error, but the job did run successfully (once I got my VM resuscitated).

deployment/aws-batch/job-definitions/production-pfb-analysis-run-job.json
```diff
@@ -26,19 +26,19 @@
...
-            "value": "7168MB"
+            "value": "24GB"
         }, {
             "name": "PFB_TEMP_FILE_LIMIT",
-            "value": "26214400"
+            "value": "100GB"
         }, {
             "name": "PFB_WORK_MEM",
-            "value": "3072MB"
+            "value": "3GB"
```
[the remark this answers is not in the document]
KlaasH: Yeah, I was going to raise that one,too, but it's the one that gets allocated for every thread in parallel processes so it can actually balloon pretty fast. So I left it, but still changed the format to match the others.

PR #948, opened 2023-08-10
Update osm tags

lalver1:
Updates crossing island tags, adds crossing_ref tag, updates flashing lights tags, and updates traffic island tags.


KlaasH (approved):
Looks good to me :+1:
My test run with the new code worked fine, and made the score for my test area in Manhattan go up slightly, which seems consistent with taking in more information about pedestrian-friendly infrastructure.

PR #949, opened 2023-08-10
Update quantile

lalver1:
Replaced the quantile function from https://github.com/tvondra/quantile.git with the postgresql built-in PERCENTILE_CONT.


KlaasH (changes requested):
This looks good, and it's always nice to remove a dependency.
There are a couple more references that should be removed as well:
- The `CREATE EXTENSION` statement in [`setup_database.sh`](https://github.com/azavea/pfb-network-connectivity/blob/620a5f4dc975891aa3b1266ced3f331fc17de19d/src/analysis/scripts/setup_database.sh#L61) (which throws an error in the analysis, though it's getting ignored, so it doesn't actually get in the way)
- The requirements list item in [README.LOCAL-ANALYSIS.md](https://github.com/azavea/pfb-network-connectivity/blob/620a5f4dc975891aa3b1266ced3f331fc17de19d/README.LOCAL-ANALYSIS.md)
- The environment variable and installation command in [Dockerfile.remote-db](https://github.com/azavea/pfb-network-connectivity/blob/620a5f4dc975891aa3b1266ced3f331fc17de19d/src/analysis/Dockerfile.remote-db).  Actually I'm not sure whether that dockerfile is in use these days.  It was created quite a while ago and if no one is using it or anticipates using it, we could potentially just get rid of it rather than updating it.

Other than those stray references, this looks good.  I ran a test analysis (a smallish chunk of Manhattan) and the updated one looked right.

KlaasH:
I've applied the fixups noted above and tested locally.  Since this is coming from a fork, I can't push to the original branch, and rather than make a new PR from a new branch, I'm going to just merge this then push my commit to develop.

KlaasH (approved):

PR #951, opened 2024-02-17
Update destinations OSM tags

lalver1:
Updated dentists, doctors, hospitals, pharmacies, retail, and schools to incorporate alternate or new OSM tags.


KlaasH (approved):
Looks good :+1: 
I did a test run and compared to the live site, and it seemed like nothing disappeared and a few things appeared (though I didn't go as far as checking whether those were in the new categories or just ones that had been added since the last run on the site).

lalver1:
Thanks!

PR #953, opened 2024-02-26
Update Docker provisioning and docker compose command

### LINKED ISSUE #952: Fix errors in Vagrant and Docker provisioning
Trying to spin up my VM for the first time in a while, I ran into some errors, especially with the Ansible role to install Docker.
I don't think it makes sense to use keep using the role from https://github.com/azavea/ansible-docker since it's not maintained, it only actually performs a few tasks, and we already also have a `pfb.docker` role doing part of the Docker provisioning work.

So I'm going to drop that shared role, move what we need from it into `pfb.docker`, and fix up whatever else needs to be fixed up.

Also, the Ansible installation is giving errors, which can hopefully be fixed by switching to newer versions of Ansible and/or a newer Vagrant base image.

KlaasH:
## Overview

Updates the Docker installation to:
- Not use the azavea.docker role (which is no longer maintained), inlining the relevant steps instead.
- Use a newer version
- Invoke Docker Compose with 'docker compose' (rather than 'docker-compose') since that's how it works in newer Docker versions.

I also updated the Vagrant base box, mainly because we might as well not be out of date if we don't have to be, and renamed the shared 'django' docker-compose config to 'django-common', since apparently docker compose now treats it as a name collision to have services in different files with the same name.

Resolves #952

### Notes

I'm not sure where things stand these days with Vagrant/Virtualbox on Mac machines. Possibly most of these changes won't be relevant.  Though the `docker-compose`->`docker compose` change will likely help if the alternate strategy is to run Docker-on-host.

## Testing Instructions

- Destroy your virtual machine, if you had one (`vagrant destroy`)
- Spin up a new one with `./scripts/setup` (unless you're doing docker-only, in which case just run `./scripts/update`)
- Start the server (`./scripts/server`, whether in or out of the VM)
- Confirm that the local server works (http://localhost:9301/)
- Optional: import some boundaries ([here's a small sample batch file](https://github.com/azavea/pfb-network-connectivity/files/14412117/WI_and_IA.zip)) and run an analysis for one of them.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Upcoming release]
 
+- Update OSM tags for doctors, dentists, hospitals, and pharmacies
+- Update Vagrant and Docker provisioning
+
 ## [0.19.0] - 2023-08-30
 
 - Update OSM tags for crossings and traffic islands (PR #948)
--- a/README.LOCAL-ANALYSIS.md
+++ b/README.LOCAL-ANALYSIS.md
@@ -7,7 +7,7 @@ run `vagrant ssh` from the project directory on your host machine.
 
 The container is built by `scripts/update`, but can also be rebuilt separately by running:
 ```
-docker-compose build analysis
+docker compose build analysis
 ```
 
 ## Example
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Build the docker container for the verification tool within the VM:
 
 ```
 cd src/verifier
-docker-compose build
+docker compose build
 ```
 
 Ensure the exported output from the analysis to check exists in the `data/output` directory. It will be there by default if the `data` directory was used for the neighborhood input shapefile.
@@ -143,7 +143,7 @@ Ensure the exported output from the analysis to check exists in the `data/output
 To compare the analysis output for Boulder, run the verification tool with:
 
 ```
-docker-compose run verifier boulder.csv
+docker compose run verifier boulder.csv
 ```
 
 Any output in the `verified_output` directory may be used for comparison.
@@ -151,7 +151,7 @@ Any output in the `verified_output` directory may be used for comparison.
 To compare to analysis output that has a non-default filename (`analysis_neighborhood_score_inputs.csv`), run the verification tool with the name of the file in `data/output` as the second argument:
 
 ```
-docker-compose run verifier boulder.csv my_output_to_verify.csv
+docker compose run verifier boulder.csv my_output_to_verify.csv
 ```
 
 If there are any differences in the outputs, a summary of the differences will be output to console.
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,7 +25,7 @@ ROOT_VM_DIR = "/vagrant"
 
 Vagrant.configure("2") do |config|
 
-  config.vm.box = "bento/ubuntu-20.04"
+  config.vm.box = "bento/ubuntu-22.04"
   config.vm.hostname = "pfb-network-connectivity"
   config.vm.network :private_network, ip: ENV.fetch("PFB_PRIVATE_IP", "192.168.56.12")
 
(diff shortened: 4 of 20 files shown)


rachelekm (approved):
I'm able to start the server from a docker-only setup 🎉  I'm having issues bringing up the VM to test, but I'm approving since I'm not sure how relevant the Vagrant setup is on Mac now? Either way leaving my notes below! (One note is that I need to run `scripts/setup` with `PFB_SHARED_FOLDER_TYPE=virtualbox` to get past mounting errors so not sure if that plays a part.)

The first error I ran into was a conflict with the upgraded base box and ansible version:
```
==> default: Running provisioner: ansible_local...
    default: Installing Ansible...
    default: Installing pip... (for Ansible installation)
The requested Ansible version (2.10.7) was not found on the guest.
Please check the Ansible installation on your Vagrant guest system (currently: 2.10.17),
or adapt the provisioner `version` option in your Vagrantfile.
See https://docs.vagrantup.com/v2/provisioning/ansible_common.html#version
for more information.
```
I was able to get past the above by setting `ansible_version` to `"latest"` but that seemed to open up other issues with what facts were available in tasks so not quite sure where to take it from here:
```
TASK [pfb.docker : Configure the Docker APT repository] ************************
fatal: [default]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'docker_repository_arch' is undefined. 'docker_repository_arch' is undefined\n\nThe error appears to be in '/vagrant/deployment/ansible/roles/pfb.docker/tasks/main.yml': line 8, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Configure the Docker APT repository\n  ^ here\n"}

PLAY RECAP *********************************************************************
default                    : ok=8    changed=6    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

KlaasH:
Ok, the "TASK [pfb.docker : Configure the Docker APT repository]" error was because I forgot to add a new file to git.  I meant to do a clean run on a separate checkout to make sure everything was in place, but I guess I forgot to go back and do that after all the changes were in.  I pushed a fixup for that error.

The Ansible version error is a little confusing because I did just do `scripts/setup` on a clean repo and it got past that step just fine.  Can you take a look at what version of the base box you have? The third line of my output was:
> Checking if box 'bento/ubuntu-22.04' version '202401.31.0' is up to date...

and I believe the answer was yes, since it didn't print a message about a newer version being available.  I wonder if you might have an older version of the base box on your machine and it has an older version of Ansible.
Possibly 'latest' would make sense as a plan, but I kind of prefer the idea of keeping it specified if we can.

KlaasH:
Ok, my VM build with the fixup for the missed file worked.  Could you try it again?  Given that your first attempt worked partly and exposed a real bug, I think we should be able to get to the point of confirming that it works for more than just me, which would be nice.

rachelekm (approved):
Success! Sort of! I'm still having trouble with that ansible version for some reason. Confirming that it's using the new base box version:
```
==> default: Importing base box 'bento/ubuntu-22.04'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'bento/ubuntu-22.04' version '202401.31.0' is up to date...
```

And annoyingly I can see that it *should* match one of the available distributions:
```
`...from versions:....2.10.0, 2.10.1, 2.10.2, 2.10.3, 2.10.4, 2.10.5, 2.10.6, 2.10.7, 3.0.0b1...`
```

I wonder if this is unique to my Vagrant or virtualbox versions? I'm using `Vagrant 2.3.4` and `VirtualBox 7.0.6`.

Tried a couple other explicitly set ansible versions to no avail, so set it to "latest" again just to get through that error and can successfully provision and bring up the server from the VMs.

KlaasH:
Ok, I'm going to say that's close enough.  It's weird that it says the requested version isn't available then includes it in the list, but if we tried to solve every Vagrant/Virtualbox mystery, we'd never have time for anything else...

PR #966, opened 2025-02-12
Fix CI for staging deployment

### LINKED ISSUE #963: Get CI staging deployment working
When the Jenkins CI server went away, we [moved the CI build to Github Actions](https://github.com/azavea/pfb-network-connectivity/pull/903), but it has never actually fully worked, because the scripts expect to get AWS credentials from a profile named `pfb` but there's no such profile configured for the runner.

I'm not sure if it's possible to configure a profile for a Github Actions runner.  If so, doing that, and providing it with a key pair, would be the simplest solution to this.  If not, we'll have to provide credentials some other way and adjust the `cipublish` and `infra` scripts accordingly.

aaronxsu:
## Overview

This PR:
- updates the `infra` script to read in AWS creds related env vars
- updates `docker-compose` file used to update job def of batch jobs to pass in the env vars

### Demo

See the green checkmark of the following GH action job run:

https://github.com/azavea/pfb-network-connectivity/actions/runs/13292649266

Staging ECS tasks were cycled through earlier today

Screenshot 2025-02-12 at 3 58 07 PM

## Testing Instructions

 * Push up a `test` branch and let the GH actions pick up the job
 * The job should succeed
 * A staging deployment should trigger

## Checklist

Resolves #963

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: CI Build
         run: ./scripts/cibuild
 
-      - uses: aws-actions/configure-aws-credentials@v1
+      - uses: aws-actions/configure-aws-credentials@v4
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/deployment/aws-batch/docker-compose.yml
+++ b/deployment/aws-batch/docker-compose.yml
@@ -1,11 +1,12 @@
 services:
   update-job-defs:
-      image: update-job-defs
-      build:
-        context: .
-        dockerfile: Dockerfile
-      environment:
-        - AWS_PROFILE=pfb
-        - AWS_DEFAULT_REGION
-      volumes:
-        - $HOME/.aws:/root/.aws:ro
+    image: update-job-defs
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      - AWS_DEFAULT_REGION
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
+    volumes:
+      - $HOME/.aws:/root/.aws:ro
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: CI Build
         run: ./scripts/cibuild
 
-      - uses: aws-actions/configure-aws-credentials@v1
+      - uses: aws-actions/configure-aws-credentials@v4
```
aaronxsu, line 35: This is not the culprit, but updating the version to the latest is nice and does not break anything.

deployment/aws-batch/docker-compose.yml
```diff
@@ -1,11 +1,12 @@
...
-      volumes:
-        - $HOME/.aws:/root/.aws:ro
+    image: update-job-defs
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      - AWS_DEFAULT_REGION
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
```
aaronxsu, line 10: Define in `docker-compose` that the above two vars should be read in from env.

scripts/infra
```diff
@@ -27,7 +27,9 @@ function update_batch_definitions() {
     docker compose build
 
     ENVIRONMENT="${ENVIRONMENT:-staging}"
-    BATCH_ANALYSIS_JOB_NAME_REVISION=$(docker compose run --rm \
+    BATCH_ANALYSIS_JOB_NAME_REVISION=$(AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID} \
```
aaronxsu, line 30: There might be a better way to pass them in, but we don't use a dotenv file as the env file in `docker-compose`, so this is the most straightforward way to do it I think. Please suggest if there is a more ideal solution!

KlaasH (approved):
This looks good, and it's nice that it's this simple.  As noted in response to your comment in `scripts/infra`, I think the change there isn't needed.

With that change backed out (and assuming an updated test branch still passes), this will be good to go :+1:

aaronxsu:
Thanks for the review. I removed it and pushed up a `test` branch here https://github.com/azavea/pfb-network-connectivity/actions/runs/13395094103. It went through. I updated this branch and will merge after CI.

Issue #960, opened 2024-12-05
Add high-level documentation of system components and functions


KlaasH:
This project has a fair number of components, and the overall shape of what they are and how they fit together is not currently documented in the main README or the ADRs, and having that written down is going to be important for enabling new people to work on the project.

So the task is to create a new documentation page cataloguing the components of the system and describing, in low-to-moderate detail, how they're set up and how they relate to each other.

KlaasH:
I decided to draft this in Google Docs rather than make a Markdown file, since it's easier to iterate with comments and suggestions in Docs than in a PR.

So [here's the draft](https://docs.google.com/document/d/1CvDxXiUBZZsFmNuDWs-OqU_LAn4rvYHYka0ry1KTyI8/edit?usp=sharing), and my proposed process is that @aaronxsu read through and make suggestions there, and once we've worked through a round of that I'll convert it to a Markdown file in `docs/` and make a PR.

aaronxsu:
@KlaasH thanks for drafting the doc. I had a good read and feel like I learned from it. I added some questions and suggestions in the doc. So please take a look when you have a chance!

PR #968, opened 2025-03-14
Add high-level documentation of system components

KlaasH:
## Overview

Not much to add here. It's a new file in `docs/` with a high-level (high- to mid-level? mostly high-level) overview of the system components, per issue #960.  Hopefully it's helpful.


docs/SystemComponentsOverview.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Django Q
+
+In addition to the Django service, we also run a [Django Q](https://django-q.readthedocs.io/en/latest/) service to handle some of the processing tasks that are triggered by API interactions but won’t fit in a request/response cycle. Namely:
+
+- Processing batch files, by opening them up, saving each geometry as a neighborhood, and submitting a Batch job for each.
+- Loading analysis imports, which includes opening the neighborhood_ways and neighborhood_census_blocks shapefiles and writing their contents to the database.
```
aaronxsu, line 91: Minor, but might be good to add quotation marks around neighborhood_ways and neighborhood_census_blocks - like what you did elsewhere in the doc

aaronxsu (approved):
Reviewed the google doc version and read the md version again just now. Good to go.

docs/SystemComponentsOverview.md
KlaasH: I think I like it better as `neighborhood_ways` in these references to the shapefiles, but still "neighborhood_ways" in the first reference because there I'm drawing attention to it as a name, rather than a file, so quotes make more sense to me. Not wedded to it, though, so quotes everywhere or backticks everywhere would also be fine.

